PR #11, opened 2020-09-15
WIP: [#10] Add non-token entrypoints and BaseDAO configuration

### LINKED ISSUE #10: Specify non-token entrypoints

In #2 we have described token entrypoints. But there will be other entrypoints as well. We should come up with their specification based on https://docs.google.com/document/d/1bTMOntqvxbHmjaERqETmqw0hy2f8JdVsjNyhpWUOqDo/edit

# Acceptance criteria

All entrypoints are described in the spec.

rinn7e:
## Description

Problem: After token related entrypoints are done, we should add other
entrypoints that is specified in the doc specification.

Solution: Add other entrypoints from the doc, add information about
"Creation Phase".

[Rendered](https://github.com/tqtezos/baseDAO/blob/rinn7e/%2310-non-token-entrypoint-spec/docs/specification.md)

## Related issue(s)

Resolves #10 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -27,12 +27,81 @@ These two parts are coupled into one smart contract because interaction between 
 - The storage of the contract must have annotations for all fields
   and must be documented to make its interpretation easy for users.
 
-# State model
+# Configuration
 
-This chapter provides a high-level overview of the contract's state.
-Note that the actual storage type is implementation detail and is not specified here.
+BaseDAO is not a concrete smart contract, but rather a framework to implement various DAOs.
+It can be configured in compile-time for any specific needs.
+"Compile-time" configuration refers to the compilation of the smart contract to Michelson.
+That is, in order to get Michelson version of the contract one has to configure it first.
 
-**TODO**
+The smart contract is implemented in Haskell eDSL Lorentz which allows us to have 2 types of configuration:
+1. Type parameters that should be instantiated to specific types by DAO creator.
+2. Michelson constants, including lambdas, that can be passed as Haskell values.
+
+BaseDAO has 1 type parameter `proposalMetadata` that defines the structure of a proposal.
+It contains fields that are required to submit a proposal.
+It can contain any fields.
+Here is an example of "treasury" style DAO:
+```haskell
+data ProposalMetadata = ProposalMetadata
+  { mutezAmount :: Mutez
+  , recipientId :: Text
+  , agoraPostId :: Natural
+  ...
+  }
+```
+
+Compile-time value parameters are captured by the `Config` type (parameterized by `proposalMetadata`):
+
+```haskell
+data Config proposalMetadata = Config
+  { cDaoName :: Text
+  -- ^ Name of the DAO.
+  , cDaoDescription :: Text
+  -- ^ Description of the DAO.
+  , cProposalCheck :: Lambda (Natural, Proposal proposalMetadata) Bool
+  -- ^ A lambda used to verify whether a proposal can be submitted.
+  -- It checks 2 things: the proposal itself and the amount of tokens frozen upon submission.
+  -- It allows the DAO to reject a proposal by arbitrary logic and captures bond requirements
+  -- (i. e. how much one should "freeze" in order to propose).
+  , cUnfrozenTokenMetadata :: FA2Metadata
+  -- ^ FA2 metadata for unfrozen token.
+  , cFrozenTokenMetadata :: FA2Metadata
+  -- ^ FA2 metadata for frozen token.
+  , cDecisionLambda :: Lambda
+      (Proposal proposalMetadata, Storage proposalMetadata)
+      (List Operation, Storage proposalMetadata)
+  -- ^ The decision lambda is executed based on a successful proposal.
+  , cRejectedProposalReturnValue :: Lambda Natural Natural
+  -- ^ When a proposal is rejected, the value that voters get back can be slashed.
+  -- This lambda specifies how many tokens will be received.
+  -- For example, if Alice freezes 100 tokens to vote and this lambda divides the value by 2,
+  -- only 50 tokens will be unfrozen and other 50 tokens will be burnt.
+  }
+```
+
+The `FA2Metadata` type matches `token_metadata` defined in FA2.
+The `Proposal` type is defined below.
+`Storage` is the full storage type of the contract.
+
+## Runtime configuration
+
+Some configuration values are specified in runtime and can be changed during the contract lifetime.
+They must be provided on origination to construct the contract's initial storage.
+These values are:
+1. `administrator :: Address` is the address that can perform administrative actions.
+2. `votingPeriod :: Natural` specifies how long the voting period lasts.
+3. `quorumThreshold :: Natural` specifies how many total votes are required for a successful proposal.
+
+# Contract logic
+
+This chapter provides a high-level overview of the contract's logic.
+
+- The contract maintains a ledger of address and its balance (frozen and unfrozen tokens)
+- The contract manages a special role called "Administrator".
+- The contract stores a list of proposals with one of the states: "ongoing", "rejected", or "accepted".
+- After a successful migration, the contract `migrated` state will be `True` and all the subsequent
+  operation will fail with `CONTRACT_MIGRATED` and the updated contract address.
 
 ## Roles
 
@@ -41,16 +110,53 @@ whole contract (hence "global"):
 
 * **administrator**
   - Can re-assign this role to a new address.
+  - Can perform administrative operations.
+  - Can trasfer FA2 tokens owned or operated by this contract.
   - There always must be exactly one administrator.
-  - [TODO]: Add DAO functionality
 
 Additionally, the contract inherits the **operator** role from FA2.
 This role is "local" to a particular address.
 Each address can have any number of operators and be an operator of any number of addresses.
 
+## Proposals
+
+Everyone can make a new proposal, however, you have to freeze some tokens for that.
+The proposer specifies how many tokens they want to freeze and this value is checked by
+the contract according to its compile-time configuration.
+If this value is accepted and the proposer has enough unfrozen tokens, these unfrozen tokens
+are frozen and voting starts.
+Apart from that the proposer specifies how many tokens one should lock in order to vote on this proposal.
+
+Proposals are identified by a natural number.
+A counter is used to assign identifiers to proposals.
+So proposal IDs are 0, 1, 2…
+
+## Voting
+
+Once a proposal is submitted, everyone can vote on it as long as they have enough unfrozen tokens.
+The exact value is specified in the proposal.
+The tokens are frozen for the duration of voting.
+Voting period is specified for the whole smart contract and can be updated by the administrator.
+It's possible to vote positively or negatively.
+After the voting ends, the contract is "flushed" by calling a dedicated entrypoint.
+
+## Ledger
+
+Every address that is stored in ledger is associated with its unfrozen token balance and
+frozen token balance.
+When unfrozen tokens are transferred, the balance of the `from_` addresses is decreased
+and the balance of the `to_` addresses is increased according to the transferred values.
+
+<!--
+# Storage
+
+The storage must be a nested tree of `pair`s with field annotations.
+It must contain the following fields:
+-->
+
 # Errors
 
-In error scenarios the stablecoin contract fails with a string or a pair where the first item is a string.
+In error scenarios the baseDAO contract fails with a string or a pair where the first item is a string.
 Here is a summary of all the strings used as error messages.
 We start with standard FA2 errors which are part of the FA2 specification.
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -389,4 +398,104 @@ unit
...
+### **vote**
+
+Types
+```haskell
+type Parameter =
+  | Vote VoteParam
+
+data VoteParam = VoteParam
+  { voProposalId :: Int
+  , .. -- TODO: not sure what other params are needed
```
gromakovsky, line 485: I think there can be a positive vote and a negative vote, so at least `Bool` is needed.

gromakovsky (changes requested):

docs/specification.md
```diff
@@ -389,4 +398,104 @@ unit
 
 **TODO**
 
+### **lock_bond**
+
+Types
+```haskell
```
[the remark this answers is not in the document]
rinn7e: I believe it is from: cameLIGO language as mentioned in the FA2 tzip: https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#interface-specification

I did not have time to understand cameLIGO properly, so I decide to use haskell syntax as a placeholder to convey the desire type at least for the discussion.

But if it is okay to include haskell syntax everywhere, I'm completely good with that and will make the change accordly.

docs/specification.md
```diff
@@ -389,4 +398,104 @@ unit
 
 **TODO**
 
+### **lock_bond**
```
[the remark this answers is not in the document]
rinn7e: That makes sense, I'll change this accordingly, Ivan.

docs/specification.md
```diff
@@ -389,4 +398,104 @@ unit
...
+
+### **proposal_metadata**
+
+Types
+```haskell
+type Parameter =
+  | ProposalMetadata MetadataParam
+
+data MetadataParam = MetadataParam
+  { voProposalId :: Int
```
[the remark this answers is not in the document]
rinn7e: I think we need an identifier for a proposal so that we can vote on it. This is a placeholder for now, I'm not sure to compute this as well. What's your advice on this, Ivan?

rinn7e:
On another note, I was wondering if we need the `TransferList` contracts (blacklist/whitelist) for BaseDAO. This is used used in `stablecoin`, and I think we may need it as well.

Also do we need entrypoint for `pause/unpause`? I believe these entrypoints are useful for updating the contracts and debugging issues.

docs/specification.md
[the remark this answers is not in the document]
gromakovsky: It's not exactly cameLIGO syntax, it's not any real language, just pseudocode, but you are right  that it's similar to cameLIGO.

> But if it is okay to include haskell syntax everywhere, I'm completely good with that and will make the change accordly.

Ok, let's do this way. Note that there are some sentences about this pseudocode like
>  Pseudocode is semi-formally defined as a list of assignments where the last assignment has `entrypoint_name` on the left and its argument type (that maps to `X`) on the right.

They should be updated if we stick to Haskell.

docs/specification.md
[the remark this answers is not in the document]
gromakovsky: I think there are many options, and I am not sure which one is better. Some options I have in mind:
1. Keep it as `Int` and require to pass proposal id when `propose` is called. Then we should check whether a proposal with this ID exists. We can change `Int` to `string` or `bytes` to make proposal IDs more descriptive, the check will remain the same.
2. Keep it as `Int` and automatically assign proposal IDs. When a new proposal is made, we assign it ID of the last proposal plus 1.
3. Use `bytes` computed by hashing a proposal. This way proposal ID computation will be stateless, which is probably good (if you have a proposal, you can compute its ID without any extra data). Note that we still need to check whether such a proposal exists. Two different proposals should have different IDs, but someone may propose two equal proposals I guess (in general case). Or we can hash not only a proposal but also a counter (either a global counter or a counter per address and also hash `SENDER`). This way proposals IDs will always be different, but the computation won't be stateless anymore.

I think we can pick any of these options and keep others in mind in case the one pick appears to be not appropriate.

gromakovsky:
Neither `Transferlist` nor `pause/unpause` are listed in the googledoc, so I propose to not include them for now.

docs/specification.md
```diff
@@ -27,6 +27,15 @@ These two parts are coupled into one smart contract because interaction between
 - The storage of the contract must have annotations for all fields
   and must be documented to make its interpretation easy for users.
 
+# DAO Creation Configuration
+TODO: Is the creation "off-chain" or "on-chain"?
```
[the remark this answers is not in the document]
rinn7e: Thanks for the detail, Ivan.

I've proposed some kind of configuration related to this, hope to get your advice whether or not it fit what you have in mind.

docs/specification.md
```diff
@@ -389,4 +398,104 @@ unit
...
+
+Types
+```haskell
+type Parameter =
+  | Propose Proposal
+
+data Proposal = Proposal
+  { prName :: String
+  , prDescription :: String
+  , prMetadata :: ProposalMetaData
```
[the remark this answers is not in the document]
rinn7e: I have add some kind of `proposalMetadata` as the type parameter here as well. Let me know if you think, Ivan.

docs/specification.md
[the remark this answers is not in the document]
rinn7e: I prefer the option 2 since it is the simplest. Should I mention this in the spec as well or not?

docs/specification.md
rinn7e: I've added `type VoteType = Upvote | Downvote` here to be more descriptive.

docs/specification.md
```diff
@@ -27,6 +27,69 @@ These two parts are coupled into one smart contract because interaction between
...
+```haskell
+data ProposalMetadata = ProposalMetadata
+  { mutezAmount :: Mutez
+  , recipientId :: Text
+  , agoraPostId :: Natural
+  ...
+  }
+```
+
+TODO: What is Associated FA2 Contract:
```
gromakovsky, line 80: The associated FA2 contract is SELF. As stated above, FA2 logic and DAO logic are coupled into one smart contract.

docs/specification.md
gromakovsky: Ok, but I think in Michelson it should still be `bool` because that's more efficient.

docs/specification.md
```diff
@@ -110,9 +96,13 @@ These values are:
...
 This chapter provides a high-level overview of the contract's state.
-Note that the exact storage type is implementation detail and is not specified here.
+Note that the exact storage type is in the implementation detail and is not specified here.
 
-**TODO**
+- The contract maintains a ledger of address and its balance (frozen and unfrozen tokens)
+- The contract manages a special role called "Administrator".
+- The contract stores a list of proposals with one of states: "ongoing", "rejected", or "accepted.
+- After a successful migration, the contract `migrated` state will be `True` and all the subsequent
+  operation will be fail with `CONTRACT_MIGRATED` and the updated contract address.
```
rinn7e, line 105: I've updated all the changes listed in `TODO`, Ivan.

One thing that I'm not clear is the "State Model". I've added this list, as well as the 'Ledger' section below. Let me know if this is the same as what you have in mind.

docs/specification.md
```diff
@@ -357,6 +463,158 @@ unit
...
+  nat
+  (pair
+    (nat %frozen_token_requirement)
+    (<proposal_type> %metadata)
+  )
+)
+```
+
+- The proposal is saved and assigned a unique numeric proposal ID.
+- The `Natural` value (let's say `n`) determines how many sender's tokens will be frozen.
```
rinn7e, line 491: Should this `Natural` (`proposal_token_requirement`) be a part of the contract runtime configuration instead of the proposal?

I think it's something that the DAO owner would want to specify.

docs/specification.md
gromakovsky: The googledoc lists "What is the frozen token ID" among "On-chain Information" that should be exposed. It should return the address of the associated FA2 contract, in our case it's `SELF`. There are several ways to make some data readable from the contract, the most common ways are:
* put it into the storage and specify how it can be found (e. g. by annotation);
* add a `view` entrypoint, that is an entrypoint that takes a callback contract.

The first option is quite complicated in our case because we don't know `SELF` address during origination. So the only way to put it into the storage is by having a dedicated entrypoint and calling it once after origination. For this reason it seems easier to have a view entrypoint instead.

So yes, it should be also an entrypoint, it should take `contract address` I suppose.

docs/specification.md
gromakovsky: My idea is that the DAO owner configures `cProposalCheck` lambda that determines how many tokens are required. In simplest case it may require a specific contracts amount of tokens to be frozen and fail if the amount is different. However, I can imagine that it can be more complex in some cases, e. g. it may depend on the actual proposal. So the person can set arbitrary `frozen_token_requirement`, but this value must pass `cProposalCheck`.

docs/specification.md
```diff
@@ -465,57 +506,115 @@ data Parameter proposalMetadata
...
+  - If proposal got accepted: the return amount are returned in total.
 
-### **flush**
+## Migrating functions
+
+### **migrate**
+
+```haskell
+data Parameter proposalMetadata
+  = Migrate Address
```
gromakovsky, line 607: Actually I intentionally removed Haskell definitions for entrypoints with trivial types like this one. Michelson representation should be treated as source of truth, because all tools can use it. Haskell definitions are provided only for readability. When there is a type with multiple fields or constructors, they are useful, but for atomic types like `address` it looks redundant.

That being said, if you really want to have Haskell definitions for all types (e. g. for consistency), I am kinda ok with that. So I won't remove them anymore, I am leaving this decision to you.

docs/specification.md
gromakovsky: > One thing that I'm not clear is the "State Model". I've added this list, as well as the 'Ledger' section below.

It looks good. I think it should be extended a bit because some logic is not described there. I think I'll do it myself a bit later today.

docs/specification.md
rinn7e: I see, that makes sense. Thanks for the explanation, Ivan.

docs/specification.md
rinn7e: I didn't know you remove it 😅. I'm okay with both, and like you said, Ivan, I only put it for consistency.

docs/specification.md
rinn7e: > So yes, it should be also an entrypoint, it should take `contract address` I suppose.

We have `token_metadata_registry` like this, and I was wondering if it is the same thing as you describe or we need a new entrypoint similar to this, Ivan?
```
type TokenMetadataRegistryParam = ContractRef Address

data Parameter proposalMetadata
  = Token_metadata_registry TokenMetadataRegistryParam

Parameter (in Michelson)
(contract %token_metadata_registry address)

- Return contract address that holds token metadata.
- Token metadata MUST include the DAO metadata.
```

gromakovsky: `token_metadata_registry` is a slightly different thing that uses similar mechanics. In our case, `tokenAddress` getter returns `SELF`, but `token_metadata_registry` may return something different. In `stablecoin`, we store metadata in a separate contract to keep the main contract a bit smaller. Here we can do the same :shrug:

docs/specification.md
```diff
@@ -357,6 +463,158 @@ unit
...
+Parameter (in Michelson):
+
+```
+(pair %vote
+  (nat %proposal_id)
+  (bool %vote_type)
+)
+```
+
+- The `Natural` value (let's say `n`) denotes the amount of tokens required by the proposal to vote.
```
gromakovsky, line 575: The only `Natural` value here is `ProposalId` and it's not the amount of tokens, probably this statement was wrongly copied here.

docs/specification.md
rinn7e: I see, so in other word, do they serve the same purpose? If they do, maybe we want to replace `token_metadata_registry` with `token_address` and if not, I'll add a separate `token_address` entrypoint.

docs/specification.md
rinn7e: I was meant to introduce a variable `n` as the amount require by the proposal to vote. Would this wording a bit better, Ivan?

`- Let `n` denotes the amount of tokens required by the associated proposal to vote.`

docs/specification.md
gromakovsky: @rinn7e not exactly the same, they are semantically different. We can't remove `token_metadata_registry` because it's required by FA2. So I think adding `token_address` is the only way to go.

docs/specification.md
gromakovsky: Yes, looks good (if you `s/denotes/denote`).

docs/specification.md
rinn7e: I've added this, let me know if it's good enough, Ivan.

docs/specification.md
rinn7e: Fixed

docs/specification.md
```diff
@@ -442,6 +443,56 @@ Parameter (in Michelson):
...
+
+data TransferParam = TransferParam
+  { from_ :: Address
+  , txs :: [TransferDestination]
+  }
+
+type TransferParams = [TransferParam]
+
+data Parameter proposalMetadata
+  = Transfer Address TransferParams
```
rinn7e, line 465:
```suggestion
  = TransferContractTokens Address TransferParams
```
Just renaming suggestion for consistent.


From what I understand, when this entrypoint is called, it delegates this call to another FA2 contract `transfer` entrypoint. Is this correct, Ivan?

docs/specification.md
```diff
@@ -97,9 +216,7 @@ Parameter (in Michelson): X
 * The previous bullet point implies that each `X` must have a field annotations with the corresponding entrypoint name.
 In the definitions below it may be omitted, but it is still implied.
 
-Pseudocode is semi-formally defined as a list of assignments where the last assignment has `entrypoint_name` on the left and its argument type (that maps to `X`) on the right.
-
-Note: pseudocode is provided only for readability.
+Note: Haskell definitions are provided only for readability.
 If Michelson type contradics what's written in pseudocode, the Michelson defition takes precedence.
```
rinn7e, line 220: If Michelson type contradicts what's written in Haskell definition, the Michelson definition takes precedence.

docs/specification.md
gromakovsky: > Just renaming suggestion for consistent.

Oh, I forgot to rename it.

> From what I understand, when this entrypoint is called, it delegates this call to another FA2 contract transfer entrypoint. Is this correct, Ivan?

Yes, exactly. The list of `operation`s it returns should contain one `TRANSFER_TOKENS` operation that calls `transfer` entrypoint.

gromakovsky: Done, renamed to `Transfer_contract_tokens`.

PruStephan (approved):

gromakovsky (approved):

Issue #6, opened 2020-08-31
Implement FA2 functionality


gromakovsky:
# Motivation and clarification

We already know that the contract must follow FA2 specification. Details are described in https://github.com/tqtezos/baseDAO/pull/2 (which is not finished yet). We should implement all that logic. But first let's wait for https://gitlab.com/morley-framework/morley-ledgers/-/issues/5

# Acceptance criteria

1. Our smart contract has all entrypoints required by FA2.
2. They reuse as much as possible from `morley-ledgers`.

gromakovsky:
Not all entrypoints have been implemented in #13.

PR #12, opened 2020-09-22
[#5] Add tests for FA2 functionality

### LINKED ISSUE #5: Add tests for FA2 functionality

The contract we are developing should follow the FA2 specification. We will need tests for FA2 functionality.

# Acceptance criteria

There are tests based on our [`cleveland` package](https://gitlab.com/morley-framework/morley/-/tree/master/code/cleveland). They should test FA2 functionality.
Note that `cleveland` has two interfaces: Nettest and Integrational. The former should be preferred, but may lack some features of the latter. If a test can be written using Nettest interface, you should prefer Nettest interface.
We can have two test-suites as we do in other contracts: one test-suite for nettest tests and one test for integrational tests. Note that all nettest tests should run using both implementations (pure and client).
Alternatively, we can have just one test-suite with root `tasty` runner. Ideally, it'd be nice to have a flag specifying whether to run tests using Client implementation (because it requires some setup).

rinn7e:
## Description
Problem: Since BaseDAO is based on FA2 functionality, we should add test cases that ensure all FA2 related parts of BaseDAO work properly.

Solution: Add nettests that tests FA2 functionality of BaseDAO.

## Related issue(s)

Resolves #5 

Depends on #6 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e:
@gromakovsky Continue from the previous comments, since the `morley-ledger`'s sample FA2 contract is merged now, I think it is the preferred option, and we should somehow re-use the Nettest: https://gitlab.com/morley-framework/morley-ledgers/-/blob/master/code/morley-ledgers-test/src/Indigo/Contracts/FA2/Nettest.hs

Do you think it is alright to re-use these tests, Ivan? What modification should I add to fit with our use case?

gromakovsky:
@rinn7e usually we use `tasty` functions to group tests and give them names. However, nettest interface is currently poorly integrated with `tasty`. For example, in https://gitlab.com/morley-framework/morley-ledgers/-/blob/bcc9ae56c07dbf7982e92199bf59069cdd56b651/code/morley-ledgers-test/src/Indigo/Contracts/Nettest/FA2.hs there is one big test that stops on the first failure. Ideally we want each small test to be a separate individual test, so that `tasty` would consider them as separate tests and report their results for each test.

With nettest interface you can do something like
```
test_genericMultisig :: TestTree
test_genericMultisig = testGroup "tests to check genericMultisig contract functionality"
  [ testCase "Call with enough amount of correct signatures \
             \with correct counter and 0 amount transfer" $
    nettestTestExpectation $ uncapsNettest $ do
      let
        counter = 0
        threshold = 2
      msig <- originateMultisig counter threshold masterPKList
      let bs = bsToSign msig counter simpleLambda
      call msig (Call @"Main") $
        mkMultisigParam counter simpleLambda $
        Nothing : Prelude.map (\sk -> Just $ SignatureEd25519 $ Ed22519.sign sk bs)
        [bobSK, carlosSK]
  , testCase "Call with insufficient signatures" $
    nettestTestExpectation $ uncapsNettest $ do
      let
        counter = 0
        threshold = 3
      msig <- originateMultisig counter threshold masterPKList
      let bs = bsToSign msig counter simpleLambda
      expectCustomError  #insufficientSignatures () $ call msig (Call @"Main")
        (mkMultisigParam counter simpleLambda $
        Nothing : Prelude.map (\sk -> Just $ SignatureEd25519 $ Ed22519.sign sk bs)
        [bobSK, carlosSK])
```

to achieve it. This way we don't run these tests in real network, but in https://gitlab.com/morley-framework/morley/-/issues/365 I suggested a way to do that.

So let's use nettest interface and put each test into its own test case using `nettestTestExpectation`. Also let's not add SMT tests in this PR, let's take care of them later.

> Do you think it is alright to re-use these tests, Ivan? What modification should I add to fit with our use case?

Since our contract is different from FA2Sample, there will be some differences in tests as well. Eventually we would like to extract common parts of tests and import them, but for now it seems simpler to have two independent versions. So instead of importing `Indigo.Contracts.FA2.Nettest` you can copy-paste relevant parts into this repo. One things become more stable, we will import tests instead.

Does it answer your questions? I didn't look at code because it may change after my answers. Let me know if/when it's ready for initial review.

rinn7e:
> So instead of importing `Indigo.Contracts.FA2.Nettest` you can copy-paste relevant parts into this repo. One things become more stable, we will import tests instead.
> Does it answer your questions? I didn't look at code because it may change after my answers. Let me know if/when it's ready for initial review.

Got this, Ivan. I'll make the change accordingly and yes, there may be significant changes, I'll let you know when this is ready for review.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,166 @@
...
+test_BaseDAO = testGroup "tests to check BaseDAO contract functionality"
+  [ testCase "Allows zero transfer from non-existent accounts" $
+      nettestTestExpectation zeroTransferScenario
+  , testCase "Is denied if only owner transfer is allowed" $
+      nettestTestExpectation ownerTransferAllowScenario
+  , testCase "aborts if there is a failure (due to low balance)" $
+      nettestTestExpectation lowBalanceScenario
+  -- TODO: testcase: "Allows zero transfer from non-existent accounts"
+  -- TODO: testcase: "Is denied if only owner transfer is allowed"
+  -- TODO: testcase: "aborts if there is a failure (due to low balance)"
```
rinn7e, line 90: @gromakovsky I've refactored the tests as above.

It is not finished yet, my plan is to implement all the TODO test cases. 

Currently, I use `FA2Interface` and `SampleFA2` from morley-ledger with slight modification to help running the tests.

I would like your feedback whether or not this tests structure is good, and whether or not the TODO test cases cover all our needs or we want like to add more (or remove some).

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,166 @@
...
+      , odContract = baseDAOContract p
+      }
+  dao <- originate originateData
+  pure ((owner1, operator1), (owner2, operator2), dao)
+
+test_BaseDAO :: TestTree
+test_BaseDAO = testGroup "tests to check BaseDAO contract functionality"
+  [ testCase "Allows zero transfer from non-existent accounts" $
+      nettestTestExpectation zeroTransferScenario
+  , testCase "Is denied if only owner transfer is allowed" $
```
gromakovsky, line 84: We've decided to use the default permissions descriptor that has `Owner_or_operator_transfer`, so operator transfer is allowed and this test looks redundant.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,166 @@
...
+  , testCase "aborts if there is a failure (due to low balance)" $
+      nettestTestExpectation lowBalanceScenario
+  -- TODO: testcase: "Allows zero transfer from non-existent accounts"
+  -- TODO: testcase: "Is denied if only owner transfer is allowed"
+  -- TODO: testcase: "aborts if there is a failure (due to low balance)"
+  -- TODO: testcase: "aborts if there is a failure (due to non existent source account)"
+  -- TODO: testcase: "aborts if there is a failure (due to bad operator)"
+  -- TODO: testcase: "aborts if there is a failure (due to bad operator for zero transfer)"
+  -- TODO: testcase: "accepts an empty list of transfers"
+  -- TODO: testcase: "is denied if operator transfer is denied in permissions descriptior"
```
gromakovsky, line 95: We've decided to use the default permissions descriptor that has `Owner_or_operator_transfer`, so operator transfer is allowed and this test looks redundant.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,166 @@
...
+  -- Self Transfer
+  -- TODO: testcase: "Cannot transfer foreign money"
+  -- TODO: testcase: "is denied if self transfer is forbidden in permissions descriptior"
+  -- TODO: testcase: "validates token id"
+  -- TODO: testcase: "aborts if there is a failure (due to low balance)"
+  -- TODO: testcase: "aborts if there is a failure (due to non existent source account)"
+  -- TODO: testcase: "aborts if there is a failure (due to bad token id)"
+  -- TODO: testcase: "balanceOf request validates token id"
+  -- TODO: testcase: "balanceOf request accepts empty list"
+  -- TODO: testcase: "Permissions_descriptor entrypoint is available"
```
gromakovsky, line 108: We don't need this entrypoint since we are using the default permission descriptor. I am not going to continue, it looks like at least one more test (the next one) is redundant. Also passing the descriptor to `baseDAOContract` is redundant.

test/Test/BaseDAO.hs
gromakovsky: The structure looks good. I have some comments about these tests. Note that in this contract we have two token_ids, so in `balance_of` tests we should test both of them. Also there is this special behavior that should be tested:
> The administrator can transfer tokens from any address to any address.

rinn7e: I've removed all the non-default permission descriptor test cases, add the admin transfer test case, as well as includes both tokens for `balance_of` tests. 

Hope to hear your feedback, Ivan.

Related to the CI, it fails due to this error: https://buildkite.com/serokell/basedao/builds/18#ccfe2eb7-9516-4e32-b535-bab5d59726d9, I'm not sure why.

gromakovsky: `indigo-0.2.2` is not present in `stackage.nix`/`hackage.nix` snapshot that we are using. I have updated them, the error should be fixed.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,344 @@
...
+      { tdFrom = AddressResolved op1
+      , tdTo = addressResolved dao
+      , tdAmount = toMutez 0
+      , tdEntrypoint = DefEpName
+      , tdParameter = FA2.Transfer [
+          ( #from_ .! owner1
+          , #txs .! [ (#to_ .! owner2, (#token_id .! 0, #amount .! 200)) ]
+          ) ]
+      }
+  transfer transferData `expectFailure`
```
gromakovsky, line 147: Oh, I wanted to write some comments yesterday, but forgot, and now there are more tests to fix, sorry :pensive: 

1. There are special cases of `transfer` adapted to smart contract calls where amount is 0: `call` and `callFrom`. I think it's better to use these functions, their use is smaller (no need to construct a large `TransferData` record).
2. You are calling the default entrypoint and construct a value of the whole parameter type. It should work but has two drawbacks. First, it forces us to update all these tests when parameter type changes. Currently it has only FA2 methods, but later it will have more and probably `FA2.Parameter` will be wrapped into another constructor. So we'll have to add something like `CallFA2 $` here. If you specify the entrypoint name explicitly, the parameter will always be just `TransferParams`. Second, it makes logs a bit less clear because in this case the parameter is a large value from which it's not clear which entrypoint is called. So it's better to specify specific entrypoint name in tests.

rinn7e: That makes sense, thanks for explanation, Ivan.

I've updated to use `callFrom` instead, you can check it.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,271 @@
...
+          { token_id = 1
+          , symbol = [mt|frozen_token|]
+          , name = [mt|Frozen Token|]
+          , decimals = 8
+          , extras = Map.empty
+          }
+      ]
+
+-- TODO: Replace SampleFA2 with BaseDAO
+baseDAOContract :: FA2.PermissionsDescriptor -> Contract FA2.Parameter SampleFA2.Storage
```
gromakovsky, line 47: Why do you still pass `PermissionsDescriptor` here if it's known in advance? :thinking:

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,271 @@
...
+lowBalanceScenario = uncapsNettest $ do
+  ((owner1, op1), (owner2, _), dao, _) <- originateBaseDAO SampleFA2.defaultPermissionDescriptor
+  let params =
+        [ ( #from_ .! owner1
+          , #txs .! [ (#to_ .! owner2, (#token_id .! 0, #amount .! 200)) ]
+          )
+        ]
+  callFrom (AddressResolved op1) dao (Call @"Transfer") params
+    `expectFailure` (NettestFailedWithError
+      (CustomError #fA2_INSUFFICIENT_BALANCE (#required .! 200, #present .! 100)))
```
gromakovsky, line 137: In latest `cleveland` there are `expectCustomError` and `expectCustomError_` functions, using them should be a bit neater (in ≈all tests).

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,271 @@
...
+  , testGroup "Owner:"
+      [ testCase "aborts if there is a failure (due to low balance)" $
+          nettestTestExpectation lowBalanceOwnerScenario
+      , testCase "validates token id" $
+          nettestTestExpectation validateTokenOwnerScenario
+      , testCase "cannot transfer foreign money" $
+          nettestTestExpectation noForeignMoneyOwnerScenario
+      , testCase "balanceOf request validates token id" $
+          nettestTestExpectation balanceOfOwnerScenario
+      ]
```
gromakovsky, line 109: It looks like the only *positive* test is zero transfer, all other tests check incorrect scenarios when something is expected to fail. Can you add tests for most common cases: when there is valid transfer and we expect it to update balances properly.

gromakovsky (changes requested):

test/Test/BaseDAO.hs
rinn7e: I forgot this, updated and remove the passed `PermissionsDescriptor`.

test/Test/BaseDAO.hs
rinn7e: That's definitely neater, I've updated to use these as well.

test/Test/BaseDAO.hs
rinn7e: I've added the `validTransfer` cases, and `updateOperator` cases. Let me know if there are other cases that I should add.

stack.yaml
```diff
@@ -21,7 +21,7 @@ extra-deps:
   - git:
       https://gitlab.com/morley-framework/morley.git # CI can't use SSH
     commit:
-      2d0506578493fec3851d711a5cc26c9dc5885001 # morley-1.7.0
+      619915ceccb66ba36d0eca06cedd76f605a15bd3 # Unreleased: rvem/#334-expect-custom-error-in-nettest
```
gromakovsky, line 24: It's already merged to master so this comment looks confusing. This branch doesn't even exist, just write `master` here.

gromakovsky (commented):
Okay, looks good now. We need to wait for https://github.com/tqtezos/baseDAO/pull/13 which should be merged soon, then adapt this PR accordingly (hopefully it should be easy). There's one problem: currently all FA2 types contain a lot of tuples and it's a bit inconvenient to work with them. We are going to update them in https://gitlab.com/morley-framework/morley-ledgers/-/issues/8, but it requires https://gitlab.com/morley-framework/morley/-/issues/357. Once we do it, we will have to update a lot of code in this repo. We can minimize this effort by doing those issues quickly, but I am not sure we have time for that. While we are waiting for #13, can you please take a look at #357? If we can do it relatively quickly, let's do it.

src/Lorentz/Contracts/Spec/FA2Interface.hs
```diff
@@ -0,0 +1,308 @@
+-- SPDX-FileCopyrightText: 2020 Tocqueville Group
+--
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | FA2 interface specification.
+-- https://gitlab.com/tzip/tzip/-/blob/667f925471728bb8e81fbd30b93a171e401b6dc1/proposals/tzip-12/tzip-12.md
+--
+module Lorentz.Contracts.Spec.FA2Interface
```
gromakovsky, line 9: Btw, why are you not using this module from `morley-ledgers`? I think we can use it here as is, because it matches FA2 directly. And it's added as a dependency in #13 anyway.

rinn7e:
> While we are waiting for #13, can you please take a look at #357? If we can do it relatively quickly, let's do it.

Sure, Ivan. I work on these issues tomorrow.

test/Test/Basic.hs
```diff
@@ -2,33 +2,16 @@
 -- SPDX-License-Identifier: LicenseRef-MIT-TQ
 
 module Test.Basic
```
gromakovsky, line 4: @rinn7e now that #13 is merged, you can rebase again and cleanup some stuff. For example, this module is not needed anymore.

rinn7e: I've removed this as well as using `BaseDAO` contract with the tests, but most of the tests fail pretty much.

I still keep `SampleFA2.hs` just in case we want to make the tests pass. Let me know if you want to change something else, Ivan.

gromakovsky: > but most of the tests fail pretty much.

They are expected to fail because `transfer` is not implemented yet. But it should be implemented soon, and then hopefully tests will pass for the actual contract.

rinn7e: FYI, currently I'm working on `debitFrom` and `creditTo` which is needed for the `propose` entrypoint (freezing the token) and since they are used to implement the `transfer` entrypoint as well, I think there probably duplicate work between me and @PruStephan. 

So if Stepan has not work on that entrypoint yet, either I could work on it or me and Stepan could share code so that it would be easier to merge later on.

What do you think, Ivan?

app/Main.hs
```diff
@@ -44,7 +44,7 @@ usageDoc = Just $ mconcat
 contracts :: ContractRegistry
 contracts = ContractRegistry $ Map.fromList
   [ "BaseDAO" ?:: ContractInfo
-    { ciContract = DAO.baseDaoContract
+    { ciContract = DAO.baseDAOContract
```
gromakovsky, line 47: We have a convention to avoid 4 capital letters in a row because it's usually hard to read. Even though `baseDAOContract` is more correct, `baseDaoContract` looks a bit more readable (it's easier to see the boundary between "DAO" and "Contract"). I wouldn't object if it was called `baseDAOContract` from the beginning, but since it's already called `baseDaoContract`, I think it's better to keep this name.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -25,18 +30,58 @@ data Storage = Storage
...
+  -> Operators
+  -> Storage
+mkStorage admin balances operators = Storage
+  { sLedger = BigMap balances
+  , sOperators = operators
+  , sTokenAddress = genesisAddress
+  , sAdmin = admin
+  }
+
+-- TODO: Needed for testing purpose
```
gromakovsky, line 64: If it's only for testing, let's move it to the test suite.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,297 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+module Test.BaseDAO
```
gromakovsky, line 4: Putting all tests into a single `Test.BaseDAO` module doesn't work in long term, soon we will have too many tests for one module. I propose to rename this module to `Test.BaseDAO.FA2` (since it targets FA2 functionality) and rename `test_BaseDAO` to `test_BaseDAO_FA2` or something like that.

test/Test/BaseDAO.hs
```diff
@@ -0,0 +1,297 @@
...
+      , odContract = DAO.baseDAOContract
+      }
+  dao <- originate originateData
+  pure ((owner1, operator1), (owner2, operator2), dao, admin)
+
+test_BaseDAO :: TestTree
+test_BaseDAO = testGroup "tests to check BaseDAO contract functionality"
+  [ testGroup "Operator:"
+      [ testCase "allows zero transfer from non-existent operator" $
+          nettestTestExpectation zeroTransferScenario
```
gromakovsky, line 57: I think we can temporarily define 
```
ignore :: Expectation -> Expectation
ignore _ = pass
```

and use it for all failing tests, we will remove it in #6 as soon as everything is properly is implemented. This way we can merge this PR (after making other requested changes) and not wait for #6.

gromakovsky (changes requested):

test/Test/Basic.hs
gromakovsky: Well, my idea was to reuse `debitFrom` and `creditTo` from `morley-ledgers`, but for that we need https://gitlab.com/morley-framework/morley-ledgers/-/issues/10 and that's what Stepan is currently working on.

If everything else required for #14 is already implemented or blocked, you can implement `debitFrom` and `creditTo` in Lorentz (because eventually we'll need them implemented in Lorentz anyway). But if there is something else to do in #14, I think it's better to do that first.

rinn7e: I see, that makes sense, Ivan. I am working on other aspect of `propose` and other entrypoints for now.

stack.yaml
```diff
@@ -1,15 +1,15 @@
 # SPDX-FileCopyrightText: 2020 TQ Tezos
 # SPDX-License-Identifier: LicenseRef-MIT-TQ
 
-resolver: lts-16.11
+resolver: lts-16.5
```
rinn7e, line 4: @gromakovsky I've changed this due to `morley` and `morley-ledgers` both have `resolver: lts-16.5`.

test/Test/BaseDAO.hs
rinn7e: I've added this, as well as other suggestion you've made, Ivan. Let me know if this MR is good enough to be merged.

test/Test/BaseDAO.hs
rinn7e: I've renamed this.

app/Main.hs
rinn7e: I see, that makes sense, Ivan. I've updated this back.

stack.yaml
gromakovsky: Yep, I assumed the same. I don't mind.

test/Test/BaseDAO/FA2.hs
```diff
@@ -36,57 +39,58 @@ originateBaseDAO = do
...
   pure ((owner1, operator1), (owner2, operator2), dao, admin)
 
-test_BaseDAO :: TestTree
-test_BaseDAO = testGroup "tests to check BaseDAO contract functionality"
+test_BaseDAO_FA2 :: TestTree
+test_BaseDAO_FA2 = testGroup "tests to check BaseDAO contract functionality"
   [ testGroup "Operator:"
       [ testCase "allows zero transfer from non-existent operator" $
-          nettestTestExpectation zeroTransferScenario
+          ignore $ nettestTestExpectation zeroTransferScenario
```
gromakovsky, line 61: AFAIU, this specific test passes already, so there is no need to ignore it.

gromakovsky (approved):
One small comment, and I think we can merge it.

PR #13, opened 2020-09-24
[#6] Add FA2 functionality

### LINKED ISSUE #6: Implement FA2 functionality

PruStephan:
## Description

We were asked to implement FA2 functionality in Lorentz. This PR contains all required functionality to be merged into project

## Related issue(s)

Resolves #6

- Tests
        silently reappearing again.

- Documentation
    - Haddock


PruStephan:
Note: I had changed the import list in order to use `FA2Interface.hs` from `morley-ledgers`, as it required a newer version of Lorentz, Indigo and Morley. I will change it back when finish, if necessary

PruStephan:
Is this one legal -  `dipN @2 assertNeq [mt|NOT_OWNER|]`? I want to check if the two top elements are equal and throw exception if they are not. Also I would like to save elements in order they are

rinn7e:
> Note: I had changed the import list in order to use `FA2Interface.hs` from `morley-ledgers`, as it required a newer version of Lorentz, Indigo and Morley. I will change it back when finish, if necessary

I think this is to be expected, since currently we use the old version, so no need to change it back.

gromakovsky:
> I had changed the import list in order to use FA2Interface.hs from morley-ledgers, as it required a newer version of Lorentz, Indigo and Morley.

Which import list? Do you mean `extra-deps` list? It's fine to update `lorentz` and `morley` versions to the latest ones used by `morley-ledgers`. That is, 0.6.1 and 1.7.0. I think you don't need to take `morley` and `lorentz` from git.

gromakovsky:
> Is this one legal - `dipN @2 assertNeq [mt|NOT_OWNER|]`?

This particular line is certainly not legal because it lacks `$` after `dipN @2`.

src/FA2.hs
```diff
@@ -0,0 +1,85 @@
...
+  transferTokens # nil # swap # cons # pair
+
+updateOperator :: Entrypoint UpdateOperatorsParam Storage
+updateOperator = do
+  iter $
+    caseT @UpdateOperator
+      ( #cAdd_operator /-> addOperator
+      , #cRemove_operator /-> removeOperator
+      )
+  nil; pair
```
gromakovsky, line 48: Well, the idea was to not implement `updateOperators` logic here but instead take it from `morley-ledgers`. Specifically, there is
```
updateOperators
  :: HasStorage Storage
  => OwnerOrOperatorTransfer
  -> Var UpdateOperatorsParam
  -> IndigoProcedure
```

`OwnerOrOperatorTransfer` in our case is `OwnerOrOperatorTransfer ("owner_or_operator_transfer" :! ())`. So it's `HasStorage => Var UpdateOperatorsParam -> IndigoProcedure`. We need to turn it into something like `'[ UpdateOperatorsParam, Storage ]  :-> '[ UpdateOperatorsParam, Storage ]`. In order to do that, we should first make `Var Storage` argument explicit, there should be some way.

`IndigoProcedure` is just a synonym for `IndigoM ()` and if you have a `Var a -> Var b -> IndigoProcedure` and you want to obtain a Lorentz `'[b, a] :-> '[b, a]` you can do so with `compileIndigo @2 funcName` and it should work.

So the only tricky thing is to turn implicit storage argument (`HasStorage`) into an explicit one. It was previously discussed in this thread: https://serokell.slack.com/archives/C017ZDQP2N8/p1599566372004000

PruStephan: Am I right than, that I should not write all this functions, instead just set all necessary params and then `compileIndigo` with the function from morley-ledgers?

PruStephan: > '[ UpdateOperatorsParam, Storage ] :-> '[ UpdateOperatorsParam, Storage ]
Do we need Params on the top of the stack in the end? I don't see any problems with removing it, cause it is not necessary for the other functions, except this one

PruStephan: Also a note. All required functions, that are implemented in `morley-ledgers` `FA2Sample.hs` are not exported from there, so I can't use the lattest version of `morley-ledgers`

gromakovsky: > Am I right than, that I should not write all this functions, instead just set all necessary params and then `compileIndigo` with the function from morley-ledgers?

Since we are told to implement it in Lorentz, it seems better not to use Indigo code in it, so long term your approach is better. But my idea was to start with reusing code from `morley-ledgers` because I expect it to be faster. Since `addOperator` is already implemented in Lorentz, let's keep it. For other methods you can do as you wish:
1. Implement them in Lorentz ← that's what we eventually need, but writing some logic in Lorentz is hard and time consuming.
2. Convert existing Indigo functions to Lorentz and use them ← it seems easier, but will have to be rewritten at some point. Currently there is no Indigo→Lorentz function that does exactly what we want, which adds extra complexity. I think it's better to create an issue about it and ask Indigo team to do it, will do.

> Do we need Params on the top of the stack in the end? 

Maybe we don't, but that's what `compileIndigo` generates. If you don't need something, you can always `drop` it.

> Also a note. All required functions, that are implemented in morley-ledgers FA2Sample.hs are not exported from there, so I can't use the lattest version of morley-ledgers

Right, I wanted to write about it yesterday, but forgot. First of all, you are welcome to make any changes you need in `morley-ledgers`. Apart from exporting some functions, you probably need to generalize them because your storage is different from the one in `FA2Sample`.

Ideally we should patch `morley-ledgers` and import functions from there, but since we need to hurry up I propose not to bother doing that for now and feel free to copy-paste some code. We will decopypaste it later.

src/FA2.hs
```diff
@@ -0,0 +1,85 @@
...
+  deriving stock (Generic, Show)
+  deriving anyclass (IsoValue, HasAnnotation)
+
+instance TypeHasDoc Storage where
+  typeDocMdDescription =
+    "Contract's storage holding a big_map with all balances and the operators."
+
+tokenMetadataRegistry :: forall store. Entrypoint TokenMetadataRegistryParam store
+tokenMetadataRegistry = do
+  sender
```
gromakovsky, line 36: Why does it return `sender`? Let's store metadata registry contract in the storage like we are doing in `stablecoin`.

src/FA2.hs
```diff
@@ -0,0 +1,85 @@
...
+import Text.Show (Show (..))
+
+import Lorentz
+import Lorentz.Contracts.Spec.FA2Interface hiding (name, sender) 
+
+import Util.Markdown
+import Util.Named
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+
+type Operators = BigMap (Address, Address) ()
```
gromakovsky, line 15: Recently there was [a change to FA2](https://gitlab.com/tzip/tzip/-/merge_requests/85) after which operators are assigned per token_id. You can change the key to `(Address, Address, TokenId)` or have two `Operators` maps (since there are 2 token_ids).

src/FA2.hs
```diff
@@ -0,0 +1,85 @@
...
+
+type Ledger = BigMap (Address, TokenId) Natural
```
gromakovsky, line 17: Another option is to have 2 `BigMap Address Natural` since there are only two token_ids. I am not sure what's better.

src/FA2.hs
```diff
@@ -0,0 +1,85 @@
+module FA2
```
gromakovsky, line 1: I know it's WIP and you are experimenting, but note that eventually it should be incorporated into the contract that we have. Currently the module is called `Basic`, we need something like `Lorentz.Contracts.BaseDAO` instead.

gromakovsky (changes requested):
@PruStephan as an intermediate milestone let's have a contract that implements only `token_metadata_registry` and `update_operators`. All other entrypoints can do nothing (not change storage, emit 0 operations). Can you please estimate how much you need to finish that?

src/FA2.hs
PruStephan: > Can you please estimate how much you need to finish that?

I think It won't take long, since token_metadata_registry seems finished (with one mistake, that will be corrected now) and updateOperators also will be finished when addOperator is finished. 

> For other methods you can do as you wish:

I think it will be better if this two stay as pure Lorentz functions, because they are near-complete. All other stuff will be transported as indigo functions firstly, and then decopypasted

src/FA2.hs
PruStephan: > Can you please estimate how much you need to finish that?

I think It won't take long, since token_metadata_registry seems finished (with one mistake, that will be corrected now) and updateOperators also will be finished when addOperator is finished. 

> For other methods you can do as you wish:

I think it will be better if this two stay as pure Lorentz functions, because they are near-complete. All other stuff will be transported as indigo functions firstly, and then decopypasted

src/FA2.hs
PruStephan: As I see in specs the variant `type Operators = BigMap (Address, (Address, TokenId)) ()` suits better, since I can update   map with operatorsParam, that I received as input

src/FA2.hs
PruStephan: Same as above for ledgers

src/FA2.hs
PruStephan: I don't understand you. Is it Fine to store it like `Lorentz/Contracts/BaseDAO/FA2.hs`?

src/FA2.hs
PruStephan: Can you explain further please

gromakovsky: This entrypoint returns address of the contract that holds token metadata. It's NOT the sender. We should store this address in the storage and return it here (by passing it to the callback contract).

src/FA2.hs
gromakovsky: It's fine to store FA2 logic in `Lorentz/Contracts/BaseDAO/FA2.hs`. But there should be the root module that exports the whole contract. This module should be called `Lorentz.Contracts.BaseDAO` and should export `Contract Parameter Storage`. This contract, in turn, should replace the `measureBoaConstrictor` contract in `Main.hs`.

stack.yaml
```diff
@@ -20,9 +18,16 @@ extra-deps:
   - git:
       https://gitlab.com/morley-framework/morley.git # CI can't use SSH
     commit:
-      3bc23ad17a0719ee96d83a94b0194ca5bfe3b8c7 # morley-1.6.0
+      2d0506578493fec3851d711a5cc26c9dc5885001 # past morley-1.7.0
```
gromakovsky, line 21: We put either branch or tag name here. AFAIU this commit is literally morley-1.7.0 tag, so you can remove "past". If you change it to a newer commit that doesn't have a tag, you can write "master" (as long as it's in master).

stack.yaml
```diff
@@ -7,8 +7,6 @@ packages:
   - .
 
 extra-deps:
-  - lorentz-0.6.0@sha256:546c75525aa41a3e5782aaa4da17bd13b72a57ecee9836c553582b675bb5f4f1,4095
-  - morley-1.6.0@sha256:62f2e31fa68c4e67fb4f4626664b5a0f02185c3d48f106b5dd10fb7efe569922,7419
```
gromakovsky, line 11: I am pretty sure you can still take this packages from Hackage rather than git repos. Just update them to 0.6.1 and 1.7.0

src/FA2.hs
```diff
@@ -0,0 +1,122 @@
...
+  dig @2
+  getField #sOperators
+  dig @2
+  dig @3
+  pair
+  dupTop2
+  mem
+  if_ ifKeyExists ifKeyDoesntExist
+  where
+    dupTop2 :: a & b & f :-> a & b & a & b & f
```
gromakovsky, line 105: Please, don't copy-paste it. In fact, I think it makes sense to add this function to Lorentz, will make an issue probably.

src/FA2.hs
PruStephan: Btw, I don't see any updates of ledger_storage from the above link. Is it updated to store the same `(Address, (Address, TokenId))` key?

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -0,0 +1,179 @@
...
+    ( #cTransfer /-> transfer
+    , #cBalance_of /-> balanceOf
+    , #cToken_metadata_registry /-> tokenMetadataRegistry
+    , #cPermissions_descriptor /-> permissionsDescriptor
+    , #cUpdate_operators /-> updateOperators
+    )
+
+-- | TODO: Implement missing methods
+transfer :: Entrypoint TransferParams Storage
+transfer = undefined
```
gromakovsky, line 59: `undefined` is bad, it causes a warning and makes it impossible to print the contract. You need something like `drop; nil; pair` (not sure it typechecks, but I hope you got the idea).

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -0,0 +1,185 @@
...
+
+-- | TODO: Implement missing methods
+transfer :: Entrypoint TransferParams Storage
+transfer = doNothing
+
+balanceOf :: Entrypoint BalanceRequestParams Storage
+balanceOf = doNothing
+
+permissionsDescriptor :: Entrypoint PermissionsDescriptorParam Storage
+permissionsDescriptor = doNothing
```
gromakovsky, line 66: Actually this entrypoint is not required by FA2, but I understand why you added it. I have created https://gitlab.com/morley-framework/morley-ledgers/-/issues/9 to help. We can keep it for now until that issue is done.

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -0,0 +1,185 @@
...
+      push Nothing
+      swap
+      update
+      setField #sOperators
+
+
+    ifKeyDoesntExist :: (Address, (Address, TokenId)) & Operators & Storage & s :-> Storage & s
+    ifKeyDoesntExist = dropN @2
+
+-- | TODO: Move this one to Lorentz
```
gromakovsky, line 167: Btw, you are welcome to review https://gitlab.com/morley-framework/morley/-/merge_requests/591. Let's keep it here for now, we will include it into the next `lorentz` release and update this place.

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -0,0 +1,185 @@
...
+dupTop2 = do
+  duupX @2
+  duupX @2
+
+-- | TODO: Probably this one can be moved to Lorentz as well
+assertNeq0or1 :: Natural & f :-> f
+assertNeq0or1 = do
+  dup
+  push @Natural 0
+  if IsEq
```
gromakovsky, line 178: Btw, there is `IsZero`.

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -0,0 +1,185 @@
...
+  then drop @Natural
+  else do
+    dup
+    push @Natural 1
+    if IsEq
+    then failUsing [mt|FA2_OPERATION_PERMITED|]
```
gromakovsky, line 184: Why "PERMITED"? Shouldn't it be "NOT_PERMITED" or "PROHIBITED"?

gromakovsky:
I have added a few more commits and comments, please take a look. Overall it looks good to me now. It's hard to review Lorentz code (whether it's correct), but in a neighbor PR we have tests, so soon we'll know whether it passes tests at least. One last thing I want you to do is some restructuring. Let's only export 5 methods from `FA2.hs` (`transfer`, `balanceOf`, etc.) and move the actual contract definition to `Lorentz.Contracts.BaseDAO`. Also let's add `Lorentz.Contracts.BaseDAO.Types` and move types there. They will be used by other modules as well.

Also you can squash most of commits now (I would keep at least "Update {hackage,stackage}.nix" and "Add Makefile" separated) and rebase on master (that's required for merge).

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: Oops, my fail. I will rewrite it

rinn7e (approved):

gromakovsky (approved):

Issue #15, opened 2020-10-06
Implement non-FA2 token functions


gromakovsky:
# Clarification and motivation

After #6 we should have all FA2 entrypoints in place. However, there are other token functions that we have to implement:
* `mint`
* `burn`
* `transfer_contract_tokens`
* `token_address`

# Acceptance criteria

`mint`, `burn`, `transfer_contract_tokens` and `token_address` are implemented according to https://github.com/tqtezos/baseDAO/blob/master/docs/specification.md and covered with tests.

PR #19, opened 2020-10-07
[#14] Implement proposals and voting logic with tests

### LINKED ISSUE #14: Implement proposals and voting logic with tests

After #5 and #6 we will have FA2 entrypoints, but there are more in the specification. We should implement them as well.
In this issue let's focus on proposals and voting.

# Acceptance criteria

The following entrypoints are implemented and tested:
* `propose`
* `proposal_metadata`
* `set_voting_period`
* `set_quorum_threshold`
* `vote`
* `flush`

Note that there are two complications:
1. #5 is not merged yet, so there is no contract to extend (I don't count `Basic.hs` because it's a stub). But #5 is close to finish, I hope it will be merged today or tomorrow.
2. The specification is not approved yet, maybe there will be some changes in it.

I hope we can still start working on this issue now, but please keep these facts in mind.

rinn7e:
Problem: After #5 and #6 we will have FA2 entrypoints, but there
are more in the specification. We are going to focus implementing
proposals and voting in this issue.

Solution: Add all entrypoints related to proposals and voting. Add
tests for these entrypoints.

## Related issue(s)

Resolves #14 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/Lorentz/Contracts/BaseDAO.hs
+++ b/src/Lorentz/Contracts/BaseDAO.hs
@@ -2,7 +2,10 @@
 -- SPDX-License-Identifier: LicenseRef-MIT-TQ
 
 module Lorentz.Contracts.BaseDAO
-  ( Ledger
+  ( Config(..)
+  , defaultConfig
+
+  , Ledger
   , Operators
   , Parameter (..)
   , Storage (..)
@@ -15,30 +18,36 @@ module Lorentz.Contracts.BaseDAO
 import Lorentz
 import Lorentz.Contracts.BaseDAO.FA2
 import Lorentz.Contracts.BaseDAO.Management
+import Lorentz.Contracts.BaseDAO.Proposal
 import Lorentz.Contracts.BaseDAO.Types as BaseDAO
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
-import Util.Markdown
 
-contractDoc :: Markdown
-contractDoc = [md|
-  This documentation describes a smart contract which implements FA2 spec
-  (https://gitlab.com/tzip/tzip/-/blob/667f925471728bb8e81fbd30b93a171e401b6dc1/proposals/tzip-12/tzip-12.md)
-  |]
-
-baseDaoContract :: Contract BaseDAO.Parameter Storage
-baseDaoContract = defaultContract $ contractName "FA2 smart contract" $ do
-  doc $ DDescription contractDoc
+baseDaoContract
+  :: forall pm.
+      ( NiceParameter pm, TypeHasDoc pm
+      , HasAnnotation pm, TypeHasDoc (Proposal pm)
+      , NicePackedValue pm
+      )
+  => Config pm -> Contract (BaseDAO.Parameter pm) (BaseDAO.Storage pm)
+baseDaoContract config@Config{..} = defaultContract $ contractName cDaoName $ do
+  doc $ DDescription cDaoDescription
   ensureZeroTransfer
   unpair
-  entryCase @BaseDAO.Parameter (Proxy @PlainEntrypointsKind)
+  entryCase @(BaseDAO.Parameter pm) (Proxy @PlainEntrypointsKind)
     ( #cCall_FA2 /-> fa2Handler
     , #cTransfer_ownership /-> transferOwnership
     , #cAccept_ownership /-> acceptOwnership
     , #cMigrate /-> migrate
     , #cConfirm_migration /-> confirmMigration
+    , #cPropose /-> propose config
+    , #cProposal_metadata /-> proposalMetadata
+    , #cVote /-> vote config
+    , #cSet_voting_period /-> setVotingPeriod config
+    , #cSet_quorum_threshold /-> setQuorumThreshold config
+    , #cFlush /-> flush config
     )
 
-fa2Handler :: Entrypoint FA2.Parameter Storage
+fa2Handler :: (IsoValue pm) => Entrypoint FA2.Parameter (BaseDAO.Storage pm)
 fa2Handler =
   entryCase @FA2.Parameter (Proxy @PlainEntrypointsKind)
     ( #cTransfer /-> transfer
--- a/src/Lorentz/Contracts/BaseDAO/FA2.hs
+++ b/src/Lorentz/Contracts/BaseDAO/FA2.hs
@@ -7,6 +7,8 @@ module Lorentz.Contracts.BaseDAO.FA2
    , tokenMetadataRegistry
    , updateOperators
    , defaultPermissionsDescriptor
+   , debitFrom
+   , creditTo
    ) where
 
 import Lorentz
@@ -22,13 +24,13 @@ defaultPermissionsDescriptor = PermissionsDescriptor
   , pdCustom = Nothing
   }
 
-transfer :: Entrypoint TransferParams Storage
+transfer :: forall pm. IsoValue pm =>  Entrypoint TransferParams (Storage pm)
 transfer = do
   dip ensureNotMigrated
   iter transferItem
   nil; pair
   where
-    transferItem :: '[TransferItem, Storage] :-> '[Storage]
+    transferItem :: '[TransferItem, Storage pm] :-> '[Storage pm]
     transferItem = do
       swap
       getField #sAdmin
@@ -44,7 +46,7 @@ transfer = do
       iter transferOne
       dropN @2
 
-    transferOne :: '[TransferDestination, Bool, Address, Storage] :-> '[Bool, Address, Storage]
+    transferOne :: '[TransferDestination, Bool, Address, Storage pm] :-> '[Bool, Address, Storage pm]
     transferOne = do
       swap
       dup
@@ -90,8 +92,8 @@ transfer = do
       swap
 
     checkSender
-      :: forall f. Address & Storage & f
-      :-> Address & Storage & f
+      :: forall f. Address : Storage pm : f
+      :-> Address : Storage pm : f
     checkSender = do
       dup
       Lorentz.sender
@@ -110,22 +112,24 @@ transfer = do
         mem
         if_ nop (failCustom_ #fA2_NOT_OPERATOR)
 
-debitFrom :: forall s. Storage & Address & (TokenId, Natural) & s :-> Storage & s
+debitFrom
+  :: forall s store pm. StorageC store pm
+  => store & Address & (TokenId, Natural) & s :-> store & s
 debitFrom = do
   dig @2
   unpair
   dig @3
   pair
   dig @2
-  getField #sLedger
+  stGetField #sLedger
   dig @2
   dupTop2
   get
   ifSome ifKeyExists ifKeyDoesntExist
   where
     ifKeyExists
-      :: Natural & (Address, TokenId) & Ledger & Storage & Natural & s
-      :-> Storage & s
+      :: Natural : (Address, TokenId) : Ledger : store : Natural : s
+      :-> store : s
     ifKeyExists = do
       dig @4
       dupTop2
@@ -140,7 +144,7 @@ debitFrom = do
           drop @Natural
           swap
           update
-          setField #sLedger)
+          stSetField #sLedger)
         (do
           toNamed #required
           swap
@@ -150,8 +154,8 @@ debitFrom = do
           failCustom #fA2_INSUFFICIENT_BALANCE)
 
     ifKeyDoesntExist
-      :: (Address, TokenId) & Ledger & Storage & Natural & s
-      :-> Storage & s
+      :: (Address, TokenId) : Ledger : store : Natural : s
+      :-> store : s
     ifKeyDoesntExist = do
       dropN @2
       swap
@@ -166,42 +170,44 @@ debitFrom = do
         failCustom #fA2_INSUFFICIENT_BALANCE
       else (dropN @2)
 
-creditTo :: forall s. Storage & Address & (TokenId, Natural) & s :-> Storage & s
+creditTo
+  :: forall s pm store. (StorageC store pm)
+  => store : Address : (TokenId, Natural) : s :-> store : s
 creditTo = do
   dig @2
   unpair
   dig @3
   pair
   dig @2
-  getField #sLedger
+  stGetField #sLedger
   dig @2
   dupTop2
   get
   ifSome ifKeyExists ifKeyDoesntExist
   where
     ifKeyExists
-      :: Natural & (Address, TokenId) & Ledger & Storage & Natural & s
-      :-> Storage & s
+      :: Natural : (Address, TokenId) : Ledger : store : Natural : s
+      :-> store : s
     ifKeyExists = do
       dig @4
       add
       some
       swap
       update
-      setField #sLedger
+      stSetField #sLedger
 
     ifKeyDoesntExist
-      :: (Address, TokenId) & Ledger & Storage & Natural & s
-      :-> Storage & s
+      :: (Address, TokenId) : Ledger : store : Natural : s
+      :-> store : s
     ifKeyDoesntExist = do
       dig @3
       some
       swap
       update
-      setField #sLedger
+      stSetField #sLedger
 
 
-balanceOf :: Entrypoint BalanceRequestParams Storage
+balanceOf :: forall pm. IsoValue pm => Entrypoint BalanceRequestParams (Storage pm)
 balanceOf = do
   dip ensureNotMigrated
   checkedCoerce_
@@ -209,15 +215,15 @@ balanceOf = do
     @(View [BalanceRequestItem] [BalanceResponseItem])
   view_ checkAll
   where
-    checkAll :: [BalanceRequestItem] & Storage & s :-> [BalanceResponseItem] & s
+    checkAll :: [BalanceRequestItem] : Storage pm : s :-> [BalanceResponseItem] : s
     checkAll = do
       nil @BalanceResponseItem
       swap
       iter checkOne
       swap
-      drop @Storage
+      drop @(Storage pm)
 
-    checkOne :: forall s. BalanceRequestItem & [BalanceResponseItem] & Storage & s :-> [BalanceResponseItem] & Storage & s
+    checkOne :: forall s. BalanceRequestItem : [BalanceResponseItem] : Storage pm : s :-> [BalanceResponseItem] : Storage pm : s
     checkOne = do
       getField #briOwner
       swap
@@ -245,23 +251,23 @@ balanceOf = do
       ifSome ifKeyExists ifKeyDoesntExist
 
     ifKeyExists
-      :: Natural & BalanceRequestItem & [BalanceResponseItem] & Storage & s
-      :-> [BalanceResponseItem] & Storage & s
+      :: Natural : BalanceRequestItem : [BalanceResponseItem] : Storage pm : s
+      :-> [BalanceResponseItem] : Storage pm : s
     ifKeyExists = do
       swap
       constructStack @BalanceResponseItem @[BalanceRequestItem, Natural]
       cons
 
     ifKeyDoesntExist
-      :: BalanceRequestItem & [BalanceResponseItem] & Storage & s
-      :-> [BalanceResponseItem] & Storage & s
+      :: BalanceRequestItem : [BalanceResponseItem] : Storage pm : s
+      :-> [BalanceResponseItem] : Storage pm : s
     ifKeyDoesntExist = do
       push @Natural 0
       swap
       constructStack @BalanceResponseItem @[BalanceRequestItem, Natural]
       cons
 
-tokenMetadataRegistry :: Entrypoint TokenMetadataRegistryParam Storage
+tokenMetadataRegistry :: IsoValue pm => Entrypoint TokenMetadataRegistryParam (Storage pm)
 tokenMetadataRegistry = do
   dip ensureNotMigrated
   swap
@@ -272,7 +278,7 @@ tokenMetadataRegistry = do
   swap
   transferTokens # nil # swap # cons # pair
 
-updateOperators :: Entrypoint UpdateOperatorsParam Storage
+updateOperators :: IsoValue pm => Entrypoint UpdateOperatorsParam (Storage pm)
 updateOperators = do
   dip ensureNotMigrated
   iter $
@@ -282,13 +288,13 @@ updateOperators = do
       )
   nil; pair
 
-addOperator :: '[OperatorParam, Storage] :-> '[Storage]
+addOperator :: forall pm. IsoValue pm => '[OperatorParam, Storage pm] :-> '[Storage pm]
 addOperator = do
   getField #opTokenId
   dip do
     getField #opOperator
     dip $ toField #opOwner
-  stackType @(TokenId & Address & Address & Storage & _)
+  stackType @(TokenId : Address : Address : (Storage pm) : _)
   assertEq0or1
   swap
   dup
@@ -305,10 +311,10 @@ addOperator = do
   mem
   if_ ifKeyExists ifKeyDoesntExist
   where
-    ifKeyExists :: '[(Address, Address), Operators, Storage] :-> '[Storage]
+    ifKeyExists :: '[(Address, Address), Operators, Storage pm] :-> '[Storage pm]
     ifKeyExists = dropN @2
 
-    ifKeyDoesntExist ::  '[(Address, Address), Operators, Storage] :-> '[Storage]
+    ifKeyDoesntExist ::  '[(Address, Address), Operators, Storage pm] :-> '[Storage pm]
     ifKeyDoesntExist = do
       unit
       some
@@ -316,13 +322,13 @@ addOperator = do
       update
       setField #sOperators
 
-removeOperator :: '[OperatorParam, Storage] :-> '[Storage]
+removeOperator :: forall pm. IsoValue pm => '[OperatorParam, Storage pm] :-> '[Storage pm]
 removeOperator = do
   getField #opTokenId
   dip do
     getField #opOperator
     dip $ toField #opOwner
-  stackType @(TokenId & Address & Address & Storage & _)
+  stackType @(TokenId : Address : Address : (Storage pm) : _)
   assertEq0or1
   swap
   dup
@@ -339,7 +345,7 @@ removeOperator = do
   mem
   if_ ifKeyExists ifKeyDoesntExist
   where
-    ifKeyExists ::  '[(Address, Address), Operators, Storage] :-> '[Storage]
+    ifKeyExists ::  '[(Address, Address), Operators, Storage pm] :-> '[Storage pm]
     ifKeyExists = do
       push Nothing
       swap
@@ -347,11 +353,11 @@ removeOperator = do
       setField #sOperators
 
 
-    ifKeyDoesntExist ::  '[(Address, Address), Operators, Storage] :-> '[Storage]
+    ifKeyDoesntExist ::  '[(Address, Address), Operators, Storage pm] :-> '[Storage pm]
     ifKeyDoesntExist = dropN @2
 
 -- | TODO: Probably this one can be moved to Lorentz as well
-assertEq0or1 :: Natural & f :-> f
+assertEq0or1 :: Natural : f :-> f
 assertEq0or1 = do
   dup
   push @Natural 0
--- a/src/Lorentz/Contracts/BaseDAO/Proposal.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Proposal.hs
@@ -1,0 +1,675 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Proposal related entrypoints
+module Lorentz.Contracts.BaseDAO.Proposal
+  ( propose
+  , proposalMetadata
+  , vote
+  , setVotingPeriod
+  , setQuorumThreshold
+  , flush
+  ) where
+
+import Lorentz
+import Lorentz.Contracts.BaseDAO.Management (authorizeAdmin)
+import Lorentz.Contracts.BaseDAO.FA2 (debitFrom, creditTo)
+import Lorentz.Contracts.BaseDAO.Types
+import Lorentz.Contracts.Spec.FA2Interface (TokenId)
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+------------------------------------------------------------------------
+-- Helper
+------------------------------------------------------------------------
+
+checkIfProposalExist
+  :: forall store pm s. (NiceParameter pm, StorageC store pm)
+  => (ProposalKey : store : s) :-> (Proposal pm : s)
+checkIfProposalExist = do
+  stGet #sProposals
+  ifSome nop (failCustom_ #pROPOSAL_NOT_EXIST)
+
+ensureVotingPeriodIsNotOver
+  :: forall store pm s. (NiceParameter pm, StorageC store pm)
+  => (Proposal pm : store : s) :-> s
+ensureVotingPeriodIsNotOver = do
+  toField #pStartDate
+  swap
+  stToField #sVotingPeriod; int
+  add -- add to the start_date
+  toNamed #expectedEndDate
+  now; toNamed #currentDate
+  if #currentDate >=. #expectedEndDate then
+    failCustom_ #vOTING_PERIOD_OVER
+  else
+    nop
+
+ensureProposalIsUnique
+  :: forall store pm s. (NicePackedValue pm, StorageC store pm)
+  => ProposeParams pm : store : s :-> ProposeParams pm : store : s
+ensureProposalIsUnique = do
+  dupTop2
+  sender; swap; pair; toProposalKey
+  swap; stToField #sProposalKeyListSortByDate
+  stackType @([ByteString] : ByteString : ProposeParams pm : store : s)
+  iter $ do
+    dip dup
+    if IsEq then
+      failCustom_ #pROPOSAL_NOT_UNIQUE
+    else
+      nop
+  drop
+
+toProposalKey
+  :: forall pm s. NicePackedValue pm
+  => (ProposeParams pm, Address) : s :-> ByteString : s
+toProposalKey = do
+  pack @(ProposeParams pm, Address)
+  blake2B
+
+------------------------------------------------------------------------
+-- Propose
+------------------------------------------------------------------------
+
+checkIsProposalValid
+  :: forall store pm. (StorageC store pm)
+  => Config pm -> '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+checkIsProposalValid Config{..} = do
+  dup
+  cProposalCheck
+  stackType @[Bool, ProposeParams pm, store]
+  if Holds then
+    nop
+  else
+    failCustom_ #fAIL_PROPOSAL_CHECK
+
+checkProposerUnfrozenToken
+  :: forall store pm. (IsoValue pm, StorageC store pm)
+  => '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+checkProposerUnfrozenToken = do
+  duupX @2
+  push unfrozenTokenId
+  sender
+  pair
+
+  stGet #sLedger; ifSome nop $ do
+    constructT
+      ( fieldCtor $ getField #ppFrozenToken >> toNamed #required
+      , fieldCtor $ push (0 :: Natural) >> toNamed #present
+      )
+    failCustom #fA2_INSUFFICIENT_BALANCE
+
+  toNamed #current_balance
+  stackType @["current_balance" :! Natural, ProposeParams pm, store]
+  duupX @2
+  toFieldNamed #ppFrozenToken
+  if #ppFrozenToken >. #current_balance then
+    failCustom_ #pROPOSAL_INSUFFICIENT_BALANCE
+  else do
+    nop
+
+checkProposalLimitReached
+  :: forall store pm. (StorageC store pm)
+  => Config pm -> '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+checkProposalLimitReached Config{..} = do
+  duupX @2
+  stToField #sProposalKeyListSortByDate
+  size
+  stackType @[Natural, ProposeParams pm, store]
+  toNamed #currentAmount
+  push (cMaxProposals); toNamed #maxAmount
+  if #maxAmount <=. #currentAmount then
+    failCustom_ #mAX_PROPOSALS_REACHED
+  else nop
+
+-- | Freeze the account's unfrozen token associated with the address
+freeze
+  :: forall store pm s. (StorageC store pm)
+  => Natural : Address : store : s :-> store : s
+freeze = do
+  dig @2
+  duupX @3
+  duupX @3
+  stackType @(Natural : Address : store : Natural : Address : s)
+
+  push unfrozenTokenId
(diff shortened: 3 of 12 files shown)


src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -5,72 +5,208 @@
...
   , sTokenAddress :: Address
 
   , sAdmin        :: Address
+  , sProposalIdCounter :: Natural
+
+  , sProposals :: BigMap ProposalId (Proposal proposalMetadata)
+  , sProposalsIdSortByDate :: [ProposalId] -- Needed for `flush`
+
+  -- Configuration (TODO: make it a separate record)
+  , sRejectedProposalReturnValue :: Natural
```
rinn7e, line 79: Took me a while to figure out how to make the types that contain `proposalMetadata` type params work (Storage, Proposal etc ...)

This is still pretty much WIP, only `propose` entrypoint is implemented which is still incomplete.

There seems to be some changes from the spec feedback that will change most of the things here as well. 

So the MR is not yet ready for review, submitted for showing current progress only. I probably focus on adding tests for these entrypoints first instead until we got the first initial feedback for the spec.

rinn7e: @gromakovsky I've updated this MR. Even though it is still WIP, I would like some of your feedback whether or not I'm on the right track, Ivan.

I have implemented:
- `propose` entrypoint: It can now check if a proposal is valid, and add to the storage. 
  - It is still in completed because it relies on `creditTo` and `debitFrom`
- `propose` tests: Current it can check for some invalid proposals.

Incompleted:
- `cProposalCheck` is current just a `Natural` which stand for the minimum amount of token required to create a proposal. I will wait for the feedback before making this a lambda.
- `vote` and other entrypoints/tests are not yet implemented.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,106 @@
...
+  if #ppProposalPriceInToken <. #sRejectedProposalReturnValue then
+    failCustom_ #fAIL_PROPOSAL_CHECK -- TODO: placeholder error
+  else nop
+
+checkSenderUnfrozenToken
+  :: forall store pm. (IsoValue pm, StorageC store pm)
+  => '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+checkSenderUnfrozenToken = do
+  duupX @2
+  push (0 :: Natural) -- unfrozen token_id
```
gromakovsky, line 36: It'd be nice to have `unfrozenTokenId :: TokenId` defined somewhere. Same for `frozenTokenId`.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -14,78 +19,203 @@ module Lorentz.Contracts.BaseDAO.Types
...
+  -- | SetQuorumThreshold ()
+  -- | Vote ()
+  -- | Flush ()
   deriving stock (Generic, Show)
-  deriving anyclass (IsoValue)
 
-instance ParameterHasEntrypoints Parameter where
-  type ParameterEntrypointsDerivation Parameter = EpdDelegate
+instance (HasAnnotation pm, IsoValue pm) => ParameterHasEntrypoints (Parameter pm) where
+  type ParameterEntrypointsDerivation (Parameter pm) = EpdRecursive
```
gromakovsky, line 134: Why not keep `EpdDelegate`?

rinn7e: Ater adding the type params `proposalMetadata`, EpdDelegate no longer works:
```
    • Could not deduce: (Michelson.Typed.Scope.FailOnOperationFound
                           (Michelson.Typed.Scope.ContainsOp (ToT pm)),
                         Michelson.Typed.Scope.FailOnNestedBigMapsFound
                           (Michelson.Typed.Scope.ContainsNestedBigMaps (ToT pm)))
        arising from the superclasses of an instance declaration
      from the context: (HasAnnotation pm, IsoValue pm)
        bound by the instance declaration
        at src/Lorentz/Contracts/BaseDAO/Types.hs:184:10-82
    • In the instance declaration for
        ‘ParameterHasEntrypoints (Parameter pm)’
    |       
184 | instance (HasAnnotation pm, IsoValue pm) => ParameterHasEntrypoints (Parameter pm) where
```

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: I have added this in `BaseDAO.Types`.

test/Test/BaseDAO/Nettest/FA2.hs
```diff
@@ -1,61 +1,26 @@
 -- SPDX-FileCopyrightText: 2020 TQ Tezos
 -- SPDX-License-Identifier: LicenseRef-MIT-TQ
 
-module Test.BaseDAO.FA2
+module Test.BaseDAO.Nettest.FA2
```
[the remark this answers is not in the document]
rinn7e: I remove the `Nettest` and reorder some of the test files structure.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: I just figure this out. I have update this by adding the `Config`. Let me know if this is good enough, Ivan.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -10,30 +10,23 @@ import Lorentz
...
-  stackType @["ppProposalPriceInToken" :! Natural, "sRejectedProposalReturnValue" :! Natural, ProposeParams pm, store]
-
-  -- TODO: `sRejectedProposalReturnValue` should be lambda instead of static nat
-  if #ppProposalPriceInToken <. #sRejectedProposalReturnValue then
-    failCustom_ #fAIL_PROPOSAL_CHECK -- TODO: placeholder error
-  else nop
+  :: forall store pm. (StorageC store pm)
+  => Config pm -> '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+checkIsProposalValid Config{..} = do
+  duupX @1
```
gromakovsky, line 16: Isn't it just `dup`?

src/Lorentz/Contracts/BaseDAO/Types.hs
gromakovsky: Yes, looks good now.

test/Test/BaseDAO/FA2.hs
```diff
@@ -126,7 +127,7 @@ validTransferOwnerScenario = uncapsNettest $ do
 
 updatingOperatorScenario :: (Monad m) => NettestImpl m -> m ()
 updatingOperatorScenario = uncapsNettest $ do
-  ((owner1, _), (owner2, _), dao, _) <- originateBaseDAO
+  ((owner1, _), (owner2, _), dao, _) <- originateBaseDAO DAO.defaultConfig
```
gromakovsky, line 130: `originateBaseDAO DAO.defaultConfig` seems to be used in many places, I would define a helper for that. Or hardcode `defaultConfig` in `originateBaseDAO` if we don't use anything else in tests.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: Thanks, Ivan. I'll continue to implement the rest of entrypoints and adding the tests. I'll let you know when this is ready for review again.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: I've updated this, and all other comments as well, Ivan.

rinn7e:
@gromakovsky I've added `proposalMetadata`, `setVotingPeriod`, `setQuorumThreshold` and `vote`.

A few assumptions, I've made which are:
- `setVotingPeriod`, `setQuorumThreshold`: will affect all ongoing proposals since all proposals use the same `votingPeriod` and `quorumThreshold` stored in `Storage`. I'm not sure if this is what we prefer. We can avoid affecting ongoing proposals' by storing their specific votingPeriod and quorum inside the `Proposal` itself, so that is what we prefer. Hope to get your advice on this.

- Currently, the votingPeriod is the day counts in: `Natural`, and I assume that it starts right away when a proposal passed the `cProposalCheck`.

- Since `vote` accepts a list of `VoteParam`, if one of the `VoteParam` failed, the ones that are succeeded are still recorded. I'm not exactly sure how to do this atomically (if one failed, the all should fail). Hope to get your advice on this as well.

I will work on `flush` next, and implements the tests for these entrypoints.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -4,10 +4,50 @@
...
+  stGet #sProposals
+  ifSome nop (failCustom_ #pROPOSAL_NOT_EXIST)
+
+ensureVotingPeriodIsNotOver
+  :: forall store pm s. (NiceParameter pm, StorageC store pm)
+  => (Proposal pm & store & s) :-> s
+ensureVotingPeriodIsNotOver = do
+  toField #pStartDate
+  swap
+  stToField #sVotingPeriod; int;
```
gromakovsky, line 37: There are some redundant `;`s in many places. Trailing `;` is redundant.

src/Lorentz/Contracts/BaseDAO.hs
```diff
@@ -33,7 +33,11 @@ baseDaoContract config@Config{..} = defaultContract $ contractName cDaoName $ do
     , #cTransfer_ownership /-> transferOwnership
     , #cAccept_ownership /-> acceptOwnership
     , #cPropose /-> propose config
-    -- , #cVote /-> vote --TODO
+    , #cProposalMetadata /-> proposalMetadata
+    , #cVote /-> vote
+    , #cSetVotingPeriod /-> setVotingPeriod
```
gromakovsky, line 38: AFAIU this entrypoint's name will be `setVotingPeriod`, however, our specification says it should be `set_voting_period`. Same applies to some other entrypoints.

gromakovsky:
> setVotingPeriod, setQuorumThreshold: will affect all ongoing proposals since all proposals use the same votingPeriod and quorumThreshold stored in Storage.

I think it's fine, but we should document such details in the specification, so that they are clear. We can revisit it later.

> Currently, the votingPeriod is the day counts in: `Natural`, and I assume that it starts right away when a proposal passed the cProposalCheck.

I think a day is too much for granularity here. Even though I agree that usually voting will usually last several days, I can imagine that someone may want to conduct quick voting. Since we are developing something very general, I think we shouldn't limit voting periods to be multiples of a day. I would specify it in seconds. This way we'll be able to have tests for this behavior and run them in a real network. If it's at least a day or an hour, running such tests will be unrealistic.

gromakovsky:
>  I'm not exactly sure how to do this atomically (if one failed, the all should fail).

AFAIU it already works this way. If one vote fails, we do `FAILWITH` which stops contract execution and doesn't mutate any state.

rinn7e:
> I think it's fine, but we should document such details in the specification, so that they are clear. We can revisit it later.

Should I update the spec in this MR, or the currently opened "Update Spec" MR?

gromakovsky:
> Should I update the spec in this MR, or the currently opened "Update Spec" MR?

Hm, I think let's do it in "Update Spec", because it's close to be finished and probably we'll merge it soon, while this one still lacks some things which will likely take more time.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: I have removed the unnecessary `;`.

src/Lorentz/Contracts/BaseDAO.hs
rinn7e: Got this, Ivan. I've fixed this accordingly.

rinn7e:
@gromakovsky I have added `flush` entrypoint, and fixed a few issues with previous entrypoints. I have added the tests for those entrypoints as well but they are incomplete.

One note is that, I need to add `pVoters :: [(Address, Natural)]` in `Proposal` since we need to keep track of voter address and the amount of token he has spent so that we can distribute it back when the proposal is `flush`. Not sure if storing in a List is a good idea, probably `Set` is a better choice in this case.

Next week, I will try various way to improve above issues, handle the TODOs that are left, and add more tests.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -3,90 +3,332 @@
...
   , sTokenAddress :: Address
   , sAdmin        :: Address
   , sPendingOwner :: Maybe Address
+
+  , sProposalIdCounter :: Natural
+  , sVotingPeriod :: VotingPeriod
+  , sQuorumThreshold :: QuorumThreshold
+
+  , sProposals :: BigMap ProposalId (Proposal proposalMetadata)
+  , sProposalsIdSortByDate :: [ProposalId] -- Newest first (probably change to set)
```
gromakovsky, line 126: Having unbounded lists in the storage is dangerous. For example, if there are too many proposals, the contract may become unusable because storage deserialization will become too expensive. So we should explicitly limits lengths of all strict data types. For example, we can require that there is no more than 500 active proposals and it's impossible for one proposal to have more than 1000 votes. I think it can be part of compile-time configuration (I mean these numbers: 500 and 1000). Then we should mention it in the specification.

If there are other unbounded types, we should limit them as well. For example, `QuorumThreshold` is `Natural` and `nat` type is unbounded, so it can be arbitrarily large as well. Even though `sQuorumThreshold` is available only to the admin who we trust, I think we should limit it as a safety measure.

rinn7e: I have added the bounded value in `Config` as below: 

```haskell
data Config = Config pm 
  { -- Bounded
  , cMaxProposals :: Natural
  , cMaxVotes :: Natural -- including both upvotes and downvotes
  , cMaxQuorumThreshold :: Natural -- Must be smaller than or equal to MaxVotes or else the threshold could be never met.
  , cMinQuorumThreshold :: Natural
  , cMaxVotingPeriod :: Natural -- Maximum of seconds allow to be set
  , cMinVotingPeriod :: Natural
  }

defaultConfig :: Config pm
defaultConfig =
  { 
  , cMaxVotingPeriod = 60 * 60 * 24 * 30 -- 1 month
  , cMinVotingPeriod = 60 * 60 -- 1 hour

  , cMaxQuorumThreshold = 1000
  , cMinQuorumThreshold = 1

  , cMaxVotes = 1000

  , cMaxProposals = 500
  }
```

I also added a check using the abouve bounded-value in `propose`, `vote`, `setQuorumThreshold` and `setVotingPeriod` entrypoints as well. Let me know if this is good enough or not, Ivan.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -86,6 +86,22 @@ checkProposerUnfrozenToken = do
...
+  push (1 :: Natural)
+  add
+  stackType @[Natural, ProposeParams pm, store]
+  toNamed #newAmount
+  push (cMaxProposals); toNamed #maxAmount
+  if #maxAmount <. #newAmount then
+    failCustom_ #mAX_PROPOSALS_REACHED
+  else nop
+
 -- TODO: implement
```
gromakovsky, line 105: The PR where it's implemented got merged recently, so it can be implemented/adapted here now.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -86,6 +86,22 @@ checkProposerUnfrozenToken = do
...
 
+checkProposalLimitReached
+  :: forall store pm. (StorageC store pm)
+  => Config pm -> '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+checkProposalLimitReached Config{..} = do
+  duupX @2
+  stToField #sProposalsIdSortByDate
+  size
+  push (1 :: Natural)
+  add
```
gromakovsky, line 97: Can we avoid adding 1 by replacing `<.` check with `<=.` below?

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -265,21 +280,39 @@ submitVote = do
...
+  :: forall store pm. (IsoValue pm, StorageC store pm)
+  => Config pm -> '[Proposal pm, VoteParam, store] :-> '[Proposal pm, VoteParam, store]
+checkVoteLimitReached Config{..} = do
+  duupX @2; duupX @2
+
+  getField #pUpvotes; dip (toField #pDownvotes); add
+  dip $ toField #vVoteAmount
+  add; toNamed #newAmount
+
+  push (cMaxVotes); toNamed #maxAmount
```
gromakovsky, line 293: Brackets look redundant.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -265,21 +280,39 @@ submitVote = do
...
+  if #maxAmount <. #newAmount then
+    failCustom_ #mAX_VOTES_REACHED
+  else nop
+
 -- | Vote
 -- TODO: If one of the vote param fails, does it cancel all the votes?
 -- Current solution, will fail if one of the vote params causes error.
```
gromakovsky, line 300: I think it's ok.

src/Lorentz/Contracts/BaseDAO/Types.hs
gromakovsky: There new checks look ok. But there is still some work that needs to be done in this PR. Please ping me when it's ready.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -3,89 +3,351 @@
...
+  , cRejectedProposalReturnValue :: forall s
+      . (Proposal proposalMetadata) & s
+    :-> ("slash_amount" :! Natural) & s
+  -- ^ When a proposal is rejected, the value that voters get back can be slashed.
+  -- This lambda specifies how many tokens will be received.
+  -- For example, if Alice freezes 100 tokens to vote and this lambda divides the value by 2,
+  -- only 50 tokens will be unfrozen and other 50 tokens will be burnt.
+
+  , cDecisionLambda :: forall s
+      . (Proposal proposalMetadata) & (Storage proposalMetadata) & s
```
gromakovsky, line 68: `:` should be better than `&` for the reason described in https://issues.serokell.io/issue/TM-307. Moreover, according to that issue `&` may end up being deprecated. So I think it's better to use `:` in new code.

rinn7e: @gromakovsky I've updated this, add flush tests, as well as handle all the left TODOs.

2 issues that I found which I marked as `[issue 1]` and `[issue 2]` in the source code.

- [issue 1]: Since it is currently not possible to do a `delay 3 seconds` in Nettest, we cannot test if a proposal voting period is over. I have a workaround for the tests by having a `cForceVotingPeriod` in the `Config` which is just to ensured that the tests work.
  - This issue is linked with issue in morley: https://gitlab.com/morley-framework/morley/-/issues/351
  - We will remove the `cForceVotingPeriod` after the above issue is resolved. I propose we handle this in a separate issue.

- [issue 2]: It is currently not possible to check just field value in `Storage` via Nettest. This is related to `Storage` containing `BigMap`. This is the related issue in morley: https://gitlab.com/morley-framework/morley/-/issues/415#note_435327096
  - I propose we handle this in a separate issue as well.

Some other problems that I have is that, some implementations are depended on `burn` entrypoints, so I add a `TODO: [#15]` which should be handled in #15 itself. (Example: slash value need to be burn`.

The last thing is testing `cDecisionLambda`, I'm not really sure how to test this.

For the rests, everything is ready for review.

Hope to get your advice on this, Ivan

gromakovsky:
There are many conflicts now:
```
docs/specification.md
src/Lorentz/Contracts/BaseDAO.hs
src/Lorentz/Contracts/BaseDAO/FA2.hs
src/Lorentz/Contracts/BaseDAO/Types.hs
test/Test/BaseDAO/FA2.hs
test/Test/Management.hs 
```

src/Lorentz/Contracts/BaseDAO/Types.hs
gromakovsky: > [issue 1]: 

I wrote [a comment](https://serokell.slack.com/archives/CF2UVU8RX/p1603736318499400?thread_ts=1603706879.484000&cid=CF2UVU8RX) about that. I think it can be implemented quite easily for Emulator, so I would wait for that.

> [issue 2]:

AFAIu a more relevant upstream issue is https://gitlab.com/morley-framework/morley/-/issues/346. Anyway, I don't mind handling it in a separate issue since the upstream issue is not started yet.

> slash value need to be `burn`

Hm, AFAIU, `burn` is basically `debitFrom`, can't you use `debitFrom`?

> The last thing is testing cDecisionLambda, I'm not really sure how to test this.

Well, you can define a decision lambda that does _something_, for example it can transfer 100 mutez to a predefined address when proposal passes. And then check that 100 mutez is actually transferred.

test/Test/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,381 @@
...
+      , testCase "can flush finished proposals" $
+          nettestTestExpectation flushAffectFinishedProposals
+      , testCase "flush should not affecting ongoing proposals" $
+          nettestTestExpectation flushNotAffectOngoingProposals
+      -- TODO [#15]: burn proposer token and test "flush"
+      -- TODO: test 'cDecisionLambda'
+      ]
+  , testGroup "Other:"
+      [ testCase "can call proposalMetadata" $
+          ignore $ nettestTestExpectation proposalMetadata
```
gromakovsky, line 61: Why is it ignored now?

test/Test/Common.hs
```diff
@@ -2,15 +2,30 @@
...
 import Lorentz.Contracts.BaseDAO (baseDaoContract)
 import Lorentz.Contracts.BaseDAO.Types
 import Morley.Nettest
+import Test.Hspec (Expectation)
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+
+import qualified Lorentz.Contracts.BaseDAO as DAO
+
+
+ignore :: Expectation -> Expectation
```
gromakovsky, line 27: Is it the only reason why `hspec` dependency was added? You can change `Expectation` to `Assertion` to avoid adding an additional dependency.

rinn7e: I've removed this for now, since we do not use it.

test/Test/BaseDAO/Proposal.hs
rinn7e: I've fixed this.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: I've created 2 separate issues to track the above problem.

I also add `burnFrozenToken` to make the slash function works properly now. 
I added a test for `cDecisionLambda` as well.

Related to using `ByteString` as key, I've refactored that as well. You can check it, Ivan.

test/Test/BaseDAO/ProposalConfig.hs
```diff
@@ -0,0 +1,69 @@
...
+  }
+  where
+    decisionLambda :: forall s store. (DAO.StorageC store pm)
+      => (DAO.Proposal pm) : store : s
+      :-> (List Operation, store) : s
+    decisionLambda = do
+      toField #pProposer
+      dip $ do
+        push (10 :: Natural)
+        push DAO.unfrozenTokenId; pair
```
gromakovsky, line 65: Isn't it simpler to `push (10 :: Natural, DAO.unfrozenTokenId)` here?

test/Test/BaseDAO/ProposalConfig.hs
```diff
@@ -0,0 +1,69 @@
...
+        push False
+  , DAO.cRejectedProposalReturnValue = do
+      -- ^ divide frozen value by half
+      toField #pProposerFrozenToken
+      push (2 :: Natural)
+      swap
+      ediv
+      ifSome
+        (do
+          unpair; dip drop
```
gromakovsky, line 39: AFAIU `unpair; dip drop` is just `car`.

gromakovsky:
The specification should be updated with regards to proposal IDs. Currently there is this
> Proposals are identified by a natural number. A counter is used to assign identifiers to proposals. So proposal IDs are 0, 1, 2…

which is not true anymore.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -40,73 +153,201 @@ instance TypeHasDoc MigrationStatus where
...
+  , sProposalKeyListSortByDate = []
+  }
+
+------------------------------------------------------------------------
+-- Proposal
+------------------------------------------------------------------------
+
+type ProposalKey = ByteString
+
+data Proposal proposalMetadata = Proposal
```
gromakovsky, line 280: It'd be nice to comment more types added in this module.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -40,73 +153,201 @@ instance TypeHasDoc MigrationStatus where
...
 
 type TransferOwnershipParam = ("newOwner" :! Address)
 type MigrateParam = ("newAddress" :! Address)
 
-emptyStorage :: Storage
+-- | Voting period in seconds
+type VotingPeriod = Natural
+
+-- | QuorumThreshold that a proposal need to meet
+-- quorum_threshold = upvote - downvote
```
gromakovsky, line 237: I see at least 2 possible interpretations of quorum threshold. It can be `positive + negative` or `positive - negative`. I guess it can also be just `positive`. I am not sure which one is correct, I've asked [here]((https://tocquevillegroup.slack.com/archives/C01BTEPMY5U/p1603814934010800).

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -172,3 +413,72 @@ instance CustomErrorHasDoc "fORBIDDEN_XTZ" where
   customErrClass = ErrClassActionException
   customErrDocMdCause =
     "Received some XTZ as part of a contract call, which is forbidden"
+type instance ErrorArg "fAIL_PROPOSAL_CHECK" = ()
```
gromakovsky, line 416:
```suggestion

type instance ErrorArg "fAIL_PROPOSAL_CHECK" = ()
```

src/Lorentz/Contracts/BaseDAO/Types.hs
gromakovsky: UPD: so it's actually the total number of votes.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,682 @@
...
+ensureVotingPeriodIsNotOver
+  :: forall store pm s. (NiceParameter pm, StorageC store pm)
+  => (Proposal pm : store : s) :-> s
+ensureVotingPeriodIsNotOver = do
+  toField #pStartDate
+  swap
+  stToField #sVotingPeriod; int
+  -- push (60 * 60 * 24 :: Natural); -- one day
+  -- mul; -- multiply with the voting period
+  add; -- add to the start_date
```
gromakovsky, line 42: AFAIU commented out code above can be removed now. Also `;` looks redundant here.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,682 @@
...
+    failCustom_ #vOTING_PERIOD_OVER
+  else
+    nop
+
+ensureProposalIsUnique
+  :: forall store pm s. (NicePackedValue pm, StorageC store pm)
+  => ProposeParams pm : store : s :-> ProposeParams pm : store : s
+ensureProposalIsUnique = do
+  duupX @2; duupX @2
+  sender; swap; pair; pack @(ProposeParams pm, Address)
```
gromakovsky, line 55: 1. I propose to define a type alias for data that is packed, otherwise it's easy to pass something wrong to `pack`. I mean, if you pass `(Address, ProposeParams)` instead of `(ProposeParams, Address)`, it will still compile. I would even go further and define something like `toProposalId` instruction and not use `pack` directly (outside of `toProposalId`).
2. I previously suggested that we also `blake2b` packed data, so that proposal IDs look more uniform and don't depend on proposal size. For large a proposal its ID can be too long.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,682 @@
...
+  stackType @[[ProposalKey], store, ProposeParams pm]
+  stSetField #sProposalKeyListSortByDate
+
+  swap
+  -- incrementProposalId
+
+-- incrementProposalId
+--   :: forall store pm. (StorageC store pm)
+--   => '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+-- incrementProposalId = do
```
gromakovsky, line 206: Probably this thing can be removed now as well.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,682 @@
...
+            dip $ dip swap
+          stackType @["slash_amount" :! Natural, Natural, Address, store, Proposal pm, ProposalKey, [Operation]]
+          burnSlashAmount
+
+          -- Calculate unfreeze amount
+          swap; dip (fromNamed #slash_amount)
+          sub; isNat
+          ifSome
+            (do
+              nop
```
gromakovsky, line 589: I think `nop` is redundant and then you don't need `do`.

rinn7e: I've refactored this.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: Removed

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: I've created a helper function `toProposalKey`  as well as add `blake2b` after `pack`.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: Fixed this.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: I have refactored this according to our discussion from slack. Let me know if this is good enough, Ivan.

test/Test/BaseDAO/ProposalConfig.hs
rinn7e: Updated this.

test/Test/BaseDAO/ProposalConfig.hs
rinn7e: Fixed this.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: I added a comment to describe this type as well as update the spec.

test/Test/BaseDAO/ProposalConfig.hs
```diff
@@ -34,12 +34,8 @@ config = DAO.defaultConfig
...
       swap
       ediv
-      ifSome
-        (do
-          unpair; dip drop
-        ) (do
-          push (0 :: Natural)
-        )
+      ifSome car $ do
+        push (0 :: Natural)
```
gromakovsky, line 38: `do` here looks redundant, I think it can be written as one line.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,674 @@
...
+    , fieldCtor $ getField #ppProposalMetadata
+    , fieldCtor $ sender
+    , fieldCtor $ getField #ppFrozenToken
+    , fieldCtor $ push []
+    )
+  dip swap
+  stackType @[Proposal pm, store, ProposeParams pm]
+  some
+  duupX @3; sender; swap; pair; toProposalKey
+  -- stToField #sProposalIdCounter
```
gromakovsky, line 194: Should be removed.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,674 @@
...
+  else nop
+
+-- | Vote
+vote
+  :: forall store pm. (StorageC store pm, NiceParameter pm)
+  => Config pm -> Entrypoint [VoteParam] store
+vote config = do
+  iter $ do
+    duupX @2
+    duupX @2
```
gromakovsky, line 349: Minor: you could use `dupTop2` here, it should be available in the `lorentz` version we use here (if it's not, ignore my comment).

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,674 @@
...
+  else do
+    -- | Add the proposalId back to #sProposalKeyListSortByDate
+    -- since we initially set this to []
+    dip drop
+
+    dip $ stGetField #sProposalKeyListSortByDate
+    cons
+    stSetField #sProposalKeyListSortByDate
+
+-- | Update quroum threshold of all ongoing and new proposals.
```
gromakovsky, line 651: Is this description correct? Quorum threshold is one for the whole contract and is updated by another entrypoint. I think this function can be better described.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -40,73 +155,209 @@ instance TypeHasDoc MigrationStatus where
...
+  { pUpvotes :: Natural
+  , pDownvotes :: Natural
+  , pStartDate :: Timestamp
+
+  , pMetadata :: proposalMetadata
+
+  , pProposer :: Address
+  , pProposerFrozenToken :: Natural
+
+  , pVoters :: [(Address, Natural)] -- This can be redundant with "Upvote/Downvote"
```
gromakovsky, line 300: > -- This can be redundant with "Upvote/Downvote"

What do you mean? AFAIU it's not redundant because we need to unfreeze tokens.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,674 @@
...
+    cond = case (cForceVotingPeriod config) of
+      True -> do drop; drop; push True
+      False -> do
+          if #currentDate >=. #expectedEndDate then
+            push True
+          else
+            push False
+
+
+-- Used in "flush". When unfreezing the tokoens of voters/proposers
```
gromakovsky, line 469:
```suggestion
-- Used in "flush". When unfreezing the tokens of voters/proposers
```

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,674 @@
...
+-- to unfreeze.
+checkBalanceLessThanFrozenValue :: forall store pm. (StorageC store pm)
+  => [Natural, Address, store, Proposal pm, ProposalKey, [Operation]]
+  :-> [Natural, Address, store, Proposal pm, ProposalKey, [Operation]]
+checkBalanceLessThanFrozenValue = do
+  toNamed #unfreeze_value
+  swap; dip swap
+  stackType @[Address, store, "unfreeze_value" :! Natural, Proposal pm, ProposalKey, [Operation]]
+  duupX @2; duupX @2
+  push (frozenTokenId); swap; pair
```
gromakovsky, line 482:
```suggestion
  push frozenTokenId; swap; pair
```

test/Test/BaseDAO/ProposalConfig.hs
rinn7e: Fixed this.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: I've removed this.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: `dupTop2` is indeed available. I've replaced all `duupX @2; duupX @2` with `dupTop2` instead.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: This is wrong comment for `flush`. I removed this, and add proper `flush` description instead.

I also made some changes to `setQuorumTheshold` comment as well.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: That comment was outdated, I removed this.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: Fixed.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
rinn7e: Fixed.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -40,73 +155,209 @@ instance TypeHasDoc MigrationStatus where
...
+  typeDocMichelsonRep = concreteTypeDocMichelsonRep @(ProposeParams ())
+
+------------------------------------------------------------------------
+-- Propose
+------------------------------------------------------------------------
+type VoteType = Bool
+
+data VoteParam = VoteParam
+  { vProposalKey :: ProposalKey
+  , vVoteType :: Bool
```
PruStephan, line 339: Why don't you use `VoteType` here?

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -90,8 +92,8 @@ transfer = do
       swap
 
     checkSender
-      :: forall f. Address & Storage & f
-      :-> Address & Storage & f
+      :: forall f. Address : Storage pm : f
+      :-> Address : Storage pm : f
```
PruStephan, line 96: What is the difference between `:` and `&`, since nothing else had changed

PruStephan (commented):
LGTM, Except for two minor questions, that I just want to clear to myself

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: Yeah, I forgot. Fixed this.

src/Lorentz/Contracts/BaseDAO/FA2.hs
rinn7e: I believe `&` will be deprecated. It is recommended to use `:`, from what I see, they serve the same purpose.

PruStephan (approved):

gromakovsky (approved):
Lorentz code is hard to review due to its low-levelness, but overall it looks fine. I think we can proceed with merge and refine later if something appears to be wrong.

PR #20, opened 2020-10-08
[#16] Implement transfer ownership entrypoints

### LINKED ISSUE #16: Implement admin transfership entrypoints

`transfer_ownership` and `accept_ownership` are not present in our contract yet, but are required by [the specification](https://github.com/tqtezos/baseDAO/blob/master/docs/specification.md).

# Acceptance criteria

`transfer_ownership` and `accept_ownership` are implemented according to https://github.com/tqtezos/baseDAO/blob/master/docs/specification.md and covered with tests.

sras:
## Description

Implements code and tests for `transferOwnership`/`acceptOwnership` entrypoints.

## Related issue(s)

Resolves #16 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/Lorentz/Contracts/BaseDAO/Management.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Management.hs
@@ -1,0 +1,83 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+module Lorentz.Contracts.BaseDAO.Management
+  ( acceptOwnership
+  , transferOwnership
+  ) where
+
+import Lorentz
+import Lorentz.Contracts.BaseDAO.Types
+import Lorentz.Contracts.ManagedLedger.Doc (DRequireRole(..))
+
+-- | Set the value of pending owner in storage
+-- using the address in parameter.
+transferOwnership
+  :: Entrypoint TransferOwnershipParam Storage
+transferOwnership = do
+  dip authorizeAdmin
+
+  -- Check if new owner is current admin. If so, we just store a `none`
+  -- value to the pending owner field in the storage, and leave a False
+  -- value on stack so that the rest of operation can be skipped. If not
+  -- we just leave a True value to indicate that setting of new pending
+  -- owner should continue.
+  fromNamed #newOwner
+  dup
+  dip $ do
+    dip $ stGetField #sAdmin
+
+    if IsEq
+      then do
+        none
+        stSetField #sPendingOwner
+        push False
+      else push True
+  swap
+
+  -- If the top value is True, we set the new pending owner. Or else
+  -- we just drop the argument and be done with it.
+  if_ (do
+    some
+    stSetField #sPendingOwner
+    ) drop
+  nil; pair
+
+-- | Checks if pending owner is set and set the value of new administrator
+acceptOwnership
+  :: Entrypoint () Storage
+acceptOwnership = do
+  drop @()
+  -- check if pending owner is set and put the unwrapped value of
+  -- pending owner at the top. Throws error if pending owner is not set
+  -- or sender address is different from the value of pending owner.
+  authorizePendingOwner
+
+  stSetField #sAdmin
+
+  -- unset pending owner field in storage
+  none
+  stSetField #sPendingOwner
+  nil; pair
+
+-- Authorise sender and move pending owner address to stack top.
+authorizePendingOwner ::
+  StorageC store => store : s :-> Address : store : s
+authorizePendingOwner = do
+  doc $ DRequireRole "pending owner"
+  stGetField #sPendingOwner
+  if IsNone
+    then failCustom_ #nO_PENDING_ADMINISTRATOR_SET
+    else do
+      dup
+      dip $ do
+        sender
+        if IsEq then nop else failCustom_ #nOT_PENDING_ADMINISTRATOR
+
+-- Authorise administrator.
+authorizeAdmin ::
+  StorageC store => store : s :-> store : s
+authorizeAdmin = do
+  doc $ DRequireRole "administrator"
+  stGetField #sAdmin; sender;
+  if IsEq then nop else failCustom_ #nOT_ADMINISTRATOR
--- a/src/Lorentz/Contracts/BaseDAO/Types.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Types.hs
@@ -1,35 +1,37 @@
 -- SPDX-FileCopyrightText: 2020 TQ Tezos
 -- SPDX-License-Identifier: LicenseRef-MIT-TQ
-
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Lorentz.Contracts.BaseDAO.Types
   ( Operators
   , Ledger
+  , Parameter (..)
   , Storage (..)
+  , StorageC
+  , TransferOwnershipParam
 
   , emptyStorage
   , mkStorage
   ) where
 
 import qualified Data.Map.Internal as Map
-import Text.Show (Show (..))
+import Text.Show (Show(..))
 
 import Lorentz
-import Lorentz.Contracts.Spec.FA2Interface
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Michelson.Runtime.GState (genesisAddress)
 
-type Operators = BigMap (Address, (Address, TokenId)) ()
+type Operators = BigMap (Address, (Address, FA2.TokenId)) ()
 
-type Ledger = BigMap (Address, TokenId) Natural
+type Ledger = BigMap (Address, FA2.TokenId) Natural
 
 -- | Storage of the FA2 contract
 data Storage = Storage
   { sLedger       :: Ledger
   , sOperators    :: Operators
   , sTokenAddress :: Address
-
   , sAdmin        :: Address
+  , sPendingOwner :: Maybe Address
   }
   deriving stock (Generic, Show)
   deriving anyclass (IsoValue, HasAnnotation)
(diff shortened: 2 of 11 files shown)


src/Lorentz/Contracts/BaseDAO/Management.hs
```diff
@@ -0,0 +1,79 @@
...
+    some
+    stSetField #sPendingOwner
+    ) drop
+  nil; pair
+
+-- | Checks if pending owner is set and set the value of new administrator
+acceptOwnership
+  :: Entrypoint () Storage
+acceptOwnership = do
+  drop
```
gromakovsky, line 46:
```suggestion
  drop @()
```

makes it clearer why we drop.

src/Lorentz/Contracts/BaseDAO/Management.hs
```diff
@@ -0,0 +1,79 @@
...
+  -- value on stack so that the rest of operation can be skipped. If not
+  -- we just leave a True value to indicate that setting of new pending
+  -- owner should continue.
+  fromNamed #newOwner
+  dup
+  dip $ do
+    dip $ stGetField #sAdmin
+
+    eq
+    if_ (do none; stSetField #sPendingOwner; push False) (push True)
```
gromakovsky, line 31: IMO, it would look better if you used `if IsEq then …` here, but that's up to you.

src/Lorentz/Contracts/BaseDAO/Management.hs
```diff
@@ -0,0 +1,79 @@
...
+  stSetField #sPendingOwner
+  nil; pair
+
+-- Authorise sender and move pending owner address to stack top.
+authorizePendingOwner ::
+  StorageC store => store : s :-> Address : store : s
+authorizePendingOwner = do
+  doc $ DRequireRole "pending owner"
+  stGetField #sPendingOwner
+  ifNone
```
gromakovsky, line 65: Another place where using `if then else` seems to look prettier.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -1,35 +1,39 @@
 -- SPDX-FileCopyrightText: 2020 TQ Tezos
 -- SPDX-License-Identifier: LicenseRef-MIT-TQ
+{-# OPTIONS_GHC -Wno-orphans #-}
```
gromakovsky, line 3: Looks duplicated.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -38,24 +42,51 @@ instance TypeHasDoc Storage where
...
+  | Transfer_ownership TransferOwnershipParam
+  | Accept_ownership AcceptOwnershipParam
+  deriving stock (Generic, Show)
+  deriving anyclass (IsoValue)
+
+instance ParameterHasEntrypoints Parameter where
+  type ParameterEntrypointsDerivation Parameter = EpdDelegate
+
+type TransferOwnershipParam = ("newOwner" :! Address)
+type AcceptOwnershipParam = ()
```
gromakovsky, line 68: `TransferOwnershipParam` is more or less ok, but `AcceptOwnershipParam` only complicates code, IMO, using `()` directly would be simpler.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -72,3 +103,29 @@ instance CustomErrorHasDoc "fROZEN_TOKEN_NOT_TRANSFERABLE" where
...
+  customErrArgumentSemantics = Nothing
+
+type instance ErrorArg "nOT_ADMINISTRATOR" = ()
+
+instance CustomErrorHasDoc "nOT_ADMINISTRATOR" where
+  customErrClass = ErrClassActionException
+  customErrDocMdCause =
+    "Received an operation that require administrative privileges\
+    \ from an address that is not the current administrator"
+  customErrArgumentSemantics = Nothing
```
gromakovsky, line 131: Well, it's `Nothing` by default, I am not sure there is benefit in specifying `Nothing` explicitly, but I don't mind.

test/Test/Management.hs
```diff
@@ -0,0 +1,111 @@
...
+testOwner = genesisAddress1
+wallet1 = genesisAddress2
+wallet2 = genesisAddress3
+
+-- | We test non-token entrypoints of the BaseDAO contract here
+test_baseDAO :: [TestTree]
+test_baseDAO =
+  [ testGroup "Ownership transfer"
+    [ testCase "Transfer ownership entrypoint authenticates sender" $
+        integrationalTestExpectation $ withOriginated initialStorage $ \baseDAO -> do
```
gromakovsky, line 30: Integrational interface is being replaced with nettest, and in new code we try to stick to nettest. See how tests in `Test.BaseDAO.FA2` are written. It should be relatively easy to transform these tests to use nettests.

After we do https://gitlab.com/morley-framework/morley/-/issues/365, we'll be able to run all these tests in a real network.

test/Test/Management.hs
```diff
@@ -0,0 +1,111 @@
...
+    initialStorage = mkStorage testOwner mempty mempty
+
+-- Helpers
+checkAdministrator :: Address -> Storage -> Either TestError ()
+checkAdministrator cadmin (sAdmin -> admin)
+  | admin == cadmin = pure ()
+  | otherwise =
+      Left $ CustomTestError $
+        "Administrator status was unexpected. Expected: " <>
+        (show cadmin) <> ", Found: " <> (show admin)
```
gromakovsky, line 94: 1. Redundant brackets.
2. Why not `pretty` instead of `show`? `show @Address` doesn't look nice.

gromakovsky (changes requested):

test/Test/Management.hs
sras: This function no longer exists.

test/Test/Management.hs
sras: Yes, I thought about it, but figured Integrational tests could work for now. Have changed it.

src/Lorentz/Contracts/BaseDAO/Types.hs
sras: Changed

src/Lorentz/Contracts/BaseDAO/Types.hs
sras: Changed

src/Lorentz/Contracts/BaseDAO/Types.hs
sras: Rebase artifact. Fixed.

src/Lorentz/Contracts/BaseDAO/Management.hs
sras: Changed.

sras:
@gromakovsky I have addressed your comments.

src/Lorentz/Contracts/BaseDAO/Management.hs
```diff
@@ -27,8 +27,7 @@ transferOwnership = do
   dip $ do
     dip $ stGetField #sAdmin
 
-    eq
-    if_ (do none; stSetField #sPendingOwner; push False) (push True)
+    if IsEq then (do none; stSetField #sPendingOwner; push False) else (push True)
```
gromakovsky, line 30: One of the reasons why I think using `IsEq` is better is because you can put this `do` statement into multiple lines which looks more natural to me. Though since this whole line is not too long, probably it's fine. Another reason is that you can remove brackets now. Brackets around `do` don't look nice, IMO. The same applies to other places below.

test/Test/Management.hs
```diff
@@ -1,111 +1,133 @@
...
-          lExpectStorage baseDAO (checkPendingOwner (Just wallet1))
-          withSender testOwner $ lCallEP baseDAO (Call @"Transfer_ownership") (#newOwner .! wallet2)
-          lExpectStorage baseDAO (checkPendingOwner (Just wallet2))
+        nettestTestExpectation $ uncapsNettest $ do
+          withOriginated 2 (\(owner:_) -> initialStorage owner) $
+            \[owner, wallet1] baseDao -> do
+              callFrom (AddressResolved owner) baseDao (Call @"Transfer_ownership")
+                (#newOwner .! wallet1)
+              -- Make the call once again to make sure the admin still retains admin
+              -- privileges
```
gromakovsky, line 48: Previously you were checking the storage using `lExpectStorage`. I think this approach sort of works as well, but did you consider using `getStorage` or `checkStorage` available in `Morley.Nettest.Caps`?

gromakovsky (changes requested):

test/Test/Management.hs
```diff
@@ -1,111 +1,133 @@
...
 test_baseDAO :: [TestTree]
 test_baseDAO =
   [ testGroup "Ownership transfer"
-    [ testCase "Transfer ownership entrypoint authenticates sender" $
-        integrationalTestExpectation $ withOriginated initialStorage $ \baseDAO -> do
-          err <- expectError $ withSender wallet1 $ lCallEP baseDAO (Call @"Transfer_ownership") (#newOwner .! wallet1)
-          expectNotAdmin err
+    [ testCase "transfer ownership entrypoint authenticates sender" $
+        nettestTestExpectation $ uncapsNettest $ do
+          withOriginated 2 (\(owner:_) -> initialStorage owner) $ \[_, wallet1] baseDao ->
```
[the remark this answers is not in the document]
sras: The `withOriginated` function provides the required number of address to both origination callback and to the test. So same number to both. So here, for origination, only one address is needed, and in tests the second address is only needed.

test/Test/Common.hs
```diff
@@ -5,15 +5,31 @@ module Test.Common
...
-  tests baseDAO
+  :: MonadNettest caps base m
+  => Integer
+  -> ([Address] -> Storage)
+  -> ([Address] -> TAddress Parameter -> m a)
+  -> m a
+withOriginated addrCount storageFn tests = do
+  addresses <- mapM (\x -> newAddress $ "address" <> (show x)) [1 ..addrCount]
+  baseDao <- originate $ OriginateData
+    { odFrom = AddressResolved genesisAddress
```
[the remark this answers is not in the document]
sras: Ah, yes. But I guess then should it be fixed in FA2 tests as well ? 

https://github.com/tqtezos/baseDAO/blob/master/test/Test/BaseDAO/FA2.hs#L48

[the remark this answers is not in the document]
gromakovsky: Looks like that, probably I've missed that when I was reviewing FA2 tests.

test/Test/Management.hs
sras: I didn't use them because we never seem to have used in any of our nettests. 

Some days back, I had asked Diogo about it and he said that we don't usually do storage checkes because storage checks only works for storages that don't have a big map in them. Not sure if that has changed though..

test/Test/Common.hs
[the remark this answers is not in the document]
sras: Changed in both places.

test/Test/Management.hs
gromakovsky: Hm, right, it only works for storage without big_maps and can't be used in our case. Probably we'll need this functionality at some point, so we should prioritize https://gitlab.com/morley-framework/morley/-/issues/346. Until it's done, if we ever need to inspect storage, we'll have to stick to Integrational interface in such case.

gromakovsky (approved):

PR #21, opened 2020-10-08
Update specification according to the feedback

rinn7e:
## Description

New changes in the specification includes:

- Remove `Natural` param from `propose` entrypoint which determines how
many tokens needed for a vote.
  - One vote now always requires one unfrozen token.

- `vote` entrypoint now accepts a list of votes which allows users to
vote on multiple proposals. Users now are allow to specify amount of votes
they want on a proposal as long as they have enough unfrozen tokens.

- Add extra fields in the configuration to set the maximum amount of
proposals, votes, and max/min value of quorum threshold and
voting_period.

- Add more custom errors needed for BaseDAO.

- Fix various small changes and improve readability of the
specification.

Thread for detail: 
- https://tocquevillegroup.slack.com/archives/C01BTEPMY5U/p1602081573007500
- https://serokell.slack.com/archives/C019PCGKTK2/p1602086626016500

## Related issue(s)

Resolves None

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -5,14 +5,8 @@ SPDX-License-Identifier: LicenseRef-MIT-TQ
 
 # Overview
 
-This specification is based on the following documents:
-
-- [BaseDAO](https://docs.google.com/document/d/1bTMOntqvxbHmjaERqETmqw0hy2f8JdVsjNyhpWUOqDo/edit#)
-
-- Tezos Token Standard: [FA2][FA2]
-
 The contract described here consists of two parts:
-- Token functionality (FA2-based).
+- Token functionality ([FA2-based][FA2]).
 - DAO functionality with proposals and voting.
 
 These two parts are coupled into one smart contract because interaction between smart contracts in Tezos is expensive and hard to get right.
@@ -55,28 +49,40 @@ Compile-time value parameters are captured by the `Config` type (parameterized b
 
 ```haskell
 data Config proposalMetadata = Config
-  { cDaoName :: Text
+  { daoName :: Text
   -- ^ Name of the DAO.
-  , cDaoDescription :: Text
+  , daoDescription :: Text
   -- ^ Description of the DAO.
-  , cProposalCheck :: Lambda (Natural, Proposal proposalMetadata) Bool
+  , unfrozenTokenMetadata :: FA2.TokenMetadata
+  -- ^ FA2 metadata for unfrozen token.
+  , frozenTokenMetadata :: FA2.TokenMetadata
+  -- ^ FA2 metadata for frozen token.
+  , proposalCheck :: Lambda (ProposeParams proposalMetadata) Bool
   -- ^ A lambda used to verify whether a proposal can be submitted.
   -- It checks 2 things: the proposal itself and the amount of tokens frozen upon submission.
   -- It allows the DAO to reject a proposal by arbitrary logic and captures bond requirements
-  -- (i. e. how much one should "freeze" in order to propose).
-  , cUnfrozenTokenMetadata :: FA2Metadata
-  -- ^ FA2 metadata for unfrozen token.
-  , cFrozenTokenMetadata :: FA2Metadata
-  -- ^ FA2 metadata for frozen token.
-  , cDecisionLambda :: Lambda
-      (Proposal proposalMetadata, Storage proposalMetadata)
-      (List Operation, Storage proposalMetadata)
-  -- ^ The decision lambda is executed based on a successful proposal.
-  , cRejectedProposalReturnValue :: Lambda Natural Natural
+  , rejectedProposalReturnValue :: Lambda (Proposal proposalMetadata) Natural
   -- ^ When a proposal is rejected, the value that voters get back can be slashed.
   -- This lambda specifies how many tokens will be received.
   -- For example, if Alice freezes 100 tokens to vote and this lambda divides the value by 2,
   -- only 50 tokens will be unfrozen and other 50 tokens will be burnt.
+  , decisionLambda :: Lambda
+      (Proposal proposalMetadata, Storage proposalMetadata)
+      (List Operation, Storage proposalMetadata)
+  -- ^ The decision lambda is executed based on a successful proposal.
+  , maxProposals :: Natural
+  -- ^ Determine the maximum number of proposals that are allowed in the contract.
+  , maxVotes :: Natural
+  -- ^ Determine the maximum number of votes associated with a proposal including positive votes
+  -- and negative votes.
+  , maxQuorumThreshold :: Natural
+  -- ^ Determine the maximum value of quorum threshold that is allowed to be set.
+  , minQuorumThreshold :: Natural
+  -- ^ Determine the minimum value of quorum threshold that is allowed to be set.
+  , maxVotingPeriod :: Natural
+  -- ^ Determine the maximum value of voting period that is allowed to be set.
+  , minVotingPeriod :: Natural
+  -- ^ Determine the minimum value of voting period that is allowed to be set.
   }
 ```
 
@@ -84,6 +90,21 @@ The `FA2Metadata` type matches `token_metadata` defined in FA2.
 The `Proposal` type is defined below.
 `Storage` is the full storage type of the contract.
 
+```haskell
+type ProposalId = Natural
+
+data Proposal proposalMetadata = Proposal
+  { pUpvotes :: Natural
+  , pDownvotes :: Natural
+  , pStartDate :: Timestamp
+  , pMetadata :: proposalMetadata
+  , pProposer :: Address
+  , pProposerFrozenToken :: Natural
+  , pVoters :: [(Address, Natural)]
+  -- ^ List of voter addresses associated with the vote amount
+  -- Needed for `flush` entrypoint.
+  }
+```
 ## Runtime configuration
 
 Some configuration values are specified in runtime and can be changed during the contract lifetime.
@@ -125,7 +146,6 @@ The proposer specifies how many tokens they want to freeze and this value is che
 the contract according to its compile-time configuration.
 If this value is accepted and the proposer has enough unfrozen tokens, these unfrozen tokens
 are frozen and voting starts.
-Apart from that the proposer specifies how many tokens one should lock in order to vote on this proposal.
 
 Proposals are identified by a natural number.
 A counter is used to assign identifiers to proposals.
@@ -134,7 +154,7 @@ So proposal IDs are 0, 1, 2…
 ## Voting
 
 Once a proposal is submitted, everyone can vote on it as long as they have enough unfrozen tokens.
-The exact value is specified in the proposal.
+One unfrozen token is required for one vote.
 The tokens are frozen for the duration of voting.
 Voting period is specified for the whole smart contract and can be updated by the administrator.
 It's possible to vote positively or negatively.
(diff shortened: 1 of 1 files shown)


gromakovsky:
One more thing we should do is to remove the private googledoc link https://docs.google.com/document/d/1bTMOntqvxbHmjaERqETmqw0hy2f8JdVsjNyhpWUOqDo/edit. Probably it won't be public, so it's quite useless in a public repo. I think we can remove the whole "This specification is based on the following documents:" section because after we remove this link it will only contain FA2 link. And we can change "FA2" below to be a link (in "Token functionality (FA2-based)").

docs/specification.md
```diff
@@ -634,27 +629,33 @@ type VoteType = Upvote | Downvote
 data VoteParam = VoteParam
   { proposal_id :: ProposalId
   , vote_type :: VoteType
+  , vote_amount :: Natural
```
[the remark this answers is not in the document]
rinn7e: This can be dangerous since it is probably not obvious for the user. The best choice would be to have a separate `vote_all` entrypoint (which accept only 1 `VoteParams`, `vote` currently accepts list of `VoteParam`). But this is just my opinion, if you are good with that, I'm good as well, Ivan.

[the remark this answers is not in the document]
gromakovsky: Okay, let's ignore this suggestion for now, it's just a suggestion, not a requirement. Things are simpler without it.

docs/specification.md
```diff
@@ -59,10 +59,15 @@ data Config proposalMetadata = Config
...
-  , cProposalCheck :: Lambda (Natural, Proposal proposalMetadata) Bool
+  , cProposalCheck :: Lambda (Natural, Proposal proposalMetadata) (Maybe Natural)
   -- ^ A lambda used to verify whether a proposal can be submitted.
   -- It checks 2 things: the proposal itself and the amount of tokens frozen upon submission.
   -- It allows the DAO to reject a proposal by arbitrary logic and captures bond requirements
+  -- It return `None` if the proposal is accepted
+  -- It return `Just Natural` if the proposal is rejected. `Natural` refer to
+  -- the amount of token that will be slashed.
+  -- For example, if Alice freezes 100 tokens to vote and this lambda divides the value by 2,
+  -- only 50 tokens will be unfrozen and other 50 tokens will be burnt.
```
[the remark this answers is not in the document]
rinn7e: I've updated this back to the way we originally thought about it, and simply add extra `Proposal` to `cRejectedProposalReturnValue` since we need access to `proposerFrozenToken` of this particular proposal to be able to apply this slash function.

rinn7e:
> One more thing we should do is to remove the private googledoc link [docs.google.com/document/d/1bTMOntqvxbHmjaERqETmqw0hy2f8JdVsjNyhpWUOqDo/edit](https://docs.google.com/document/d/1bTMOntqvxbHmjaERqETmqw0hy2f8JdVsjNyhpWUOqDo/edit). Probably it won't be public, so it's quite useless in a public repo. I think we can remove the whole "This specification is based on the following documents:" section because after we remove this link it will only contain FA2 link. And we can change "FA2" below to be a link (in "Token functionality (FA2-based)").

I've updated this as well, Ivan.

docs/specification.md
```diff
@@ -544,9 +537,9 @@ unit
 
 ```haskell
 data Proposal proposalMetadata = Proposal
-    { frozen_token_requirement :: Natural
-    -- ^ How many tokens should be frozen in order to vote on this proposal.
-    , metadata :: proposalMetadata
+    { metadata :: proposalMetadata
+    , proposer :: Address
```
gromakovsky, line 541: Can `proposer` be different from `SENDER`?

docs/specification.md
```diff
@@ -634,27 +630,33 @@ type VoteType = Upvote | Downvote
...
+      (bool %vote_type)
+      (nat %vote_amount)
+    )
+  )
 )
 ```
 
-- Let `n` denote the amount of tokens required by the associated proposal to vote.
-- Sender MUST have enough unfrozen tokens (i. e. `≥ n`).
+- Sender MUST have unfrozen tokens equal to `vote_amount` (1 unfrozen token is needed for 1 vote).
```
gromakovsky, line 654: Sender must have `vote_amount` or more. It may sound confusing because one may think that the balance must be equal to `vote_amount`, but in fact it can be greater.

docs/specification.md
rinn7e: `proposer` is always `SENDER` of `Propose` entrypoint, unless we allow `Operators` to do it on the behalf. But currently we do not take operators into account.

I have to add this field to keep track of who the proposer so that we can unfreeze his tokens when `flush`.

docs/specification.md
rinn7e: I have updated this, and add other detail related to `set_voting_period` and `set_quorum_threshold` as well.

docs/specification.md
gromakovsky: It's ok to store `proposer` address in storage, but it shouldn't be passed to `propose` entrypoint because it's redundant and error-prone. We have to make an extra check (that `proposer` equals to `SENDER`) which costs some gas and can be accidentally forgotten. So `Proposal` that we pass to `propose` shouldn't have `proposer :: Address` in it. In storage we can store something like `(Proposal, Address)`, for example.

docs/specification.md
```diff
@@ -544,9 +537,9 @@ unit
 
 ```haskell
 data Proposal proposalMetadata = Proposal
-    { frozen_token_requirement :: Natural
-    -- ^ How many tokens should be frozen in order to vote on this proposal.
-    , metadata :: proposalMetadata
+    { metadata :: proposalMetadata
+    , proposer :: Address
+    , proposerFrozenToken :: Natural
```
gromakovsky, line 542: Now it looks confusing. There is also `Natural` inside `Propose` constructor. How are these 2 different? AFAIU only one should be present.

gromakovsky (changes requested):
> Apart from that the proposer specifies how many tokens one should lock in order to vote on this proposal.

There is the following statement which doesn't look right anymore.

docs/specification.md
gromakovsky: Also I think it makes sense to return `proposer` address from `proposal_metadata`, so maybe we need two types: one without `proposer` and one with it.

gromakovsky: Oh, I see you already have `ProposeParams`. That's what I had in mind as well, please update the specification accordingly.

docs/specification.md
rinn7e: > Apart from that the proposer specifies how many tokens one should lock in order to vote on this proposal.

I've remove this.

> Now it looks confusing. There is also Natural inside Propose constructor. How are these 2 different? AFAIU only one should be present.

I've updated this to use `ProposeParam` instead.

docs/specification.md
rinn7e: Sorry for the confusion, Ivan. I've updated `Config`, and added `Proposal`, `Storage` types according to the current MR of propose-related impl of baseDAO.

I've added a few errors that are needed, use `ProposeParam` for `propose` entrypoint, and add some new description addressing the new errors and bounded value.

docs/specification.md
```diff
@@ -57,33 +51,83 @@ Compile-time value parameters are captured by the `Config` type (parameterized b
...
-  , cFrozenTokenMetadata :: FA2Metadata
-  -- ^ FA2 metadata for frozen token.
-  , cDecisionLambda :: Lambda
-      (Proposal proposalMetadata, Storage proposalMetadata)
-      (List Operation, Storage proposalMetadata)
-  -- ^ The decision lambda is executed based on a successful proposal.
-  , cRejectedProposalReturnValue :: Lambda Natural Natural
+  , cRejectedProposalReturnValue :: forall s
+      . (Proposal proposalMetadata) & s
+    :-> ("slash_amount" :! Natural) & s
```
gromakovsky, line 68: Oh, it's getting more and more technical, and now only developers with Haskell and Lorentz experience can understand what's written there. This specification is supposed to be read not only by developers. Some technical knowledge is of course required, but it should be understood by a person who has some programming experience (in any language) and is familiar with Michelson. Things like `forall s`, `&`, `"slash_amount" :!` and maybe something else are very domain specific and should be avoided. `Lambda Natural Natural` is quite understandable by everyone familiar with Michelson, but the current definition is complicated. `lambda a b` can be converted to `forall s. a : s :-> b : s`, but the latter is less readable and the difference is not essential, so we omit such details.

`Markdown` type is also too technical. What is it? How is it mapped to Michelson? And why markdown, where does it come from? Why don't we permit something else? A plain string is the most generic type there.

docs/specification.md
```diff
@@ -57,33 +51,83 @@ Compile-time value parameters are captured by the `Config` type (parameterized b
...
+  , cDecisionLambda :: forall s
+      . (Proposal proposalMetadata) & (Storage proposalMetadata) & s
+    :-> (List Operation, Storage proposalMetadata) & s
+  -- ^ The decision lambda is executed based on a successful proposal.
+  , cMaxProposals :: Natural
+  , cMaxVotes :: Natural
+  , cMaxQuorumThreshold :: Natural
+  , cMinQuorumThreshold :: Natural
+  , cMaxVotingPeriod :: Natural
+  , cMinVotingPeriod :: Natural
```
gromakovsky, line 82: Can you please comment these fields too?

docs/specification.md
```diff
@@ -57,33 +51,83 @@ Compile-time value parameters are captured by the `Config` type (parameterized b
...
+
+```haskell
+-- | Voting period in seconds
+type VotingPeriod = Natural
+
+-- | Quorum threshold of a proposal that needs to be met to be accepted
+-- quorum_threshold = upvotes - downvotes
+type QuorumThreshold = Natural
+
+data Storage = Storage
```
gromakovsky, line 98: I'd rather not specify the exact storage type here because:
1. If we specify it here, we will have to maintain it up to date with code and most likely we'll forget to do it at some point. Since not everything is implemented yet, the storage type will likely change many times.
2. This specification is supposed to explain what the contract does and how one can use it. "Usage" basically boils down to calling entrypoints (that's why we specify their exact types) and constructing the initial storage. A large part of the initial storage is predefined. For example, `sProposalIdCounter` is always 0 initially. There are some values that should be provided explicitly, we already list them in "Runtime specification". One may also inspect storage fields directly, that's why we say that "The storage of the contract must have annotations for all fields and must be documented to make its interpretation easy for users".
3. As I wrote in another comment, this specification shouldn't be deeply technical. The fact that we, for example, store proposal IDs sorted by date is really just an implementation detail. Keep in mind that we also have auto-generated documentation that contains the whole storage type which is updated automatically. Those who want to interact with the contract at lower level (e. g. implement their own decision lambda) can refer to autogenerated documentation instead.

docs/specification.md
rinn7e: That makes sense, Ivan. I was not thinking it through, I have updated the syntax to be more general now. I've removed the `Markdown`, use the `Lambda` back, and rename the record field as well. Let me know if this is good enough, Ivan.

docs/specification.md
rinn7e: I've added comment to these fields.

docs/specification.md
rinn7e: I agree that it can be a pain to maintain. Initially, I thought this may be useful as an example of the implementation of `Storage`. But it can be confusing and too technical.

I've removed this for now, but I still keep the type `Proposal` since it is used in lambda functions specified in `Config` and would be useful to understand how those functions work.

gromakovsky: `Proposal` is ok, it's less likely to change while `Storage` gathers data from all components of the contract, so it is expected to change sometimes.

docs/specification.md
gromakovsky: Looks good, removal of prefixes also makes sense.

gromakovsky (approved):
Ok, looks good, I approve, but please describe the changes in the commit message. You can squash everything and write one message with all changes or put each logical change into own commit.

rinn7e:
> Ok, looks good, I approve, but please describe the changes in the commit message. You can squash everything and write one message with all changes or put each logical change into own commit.

I've squashed and add more detail description of the new changes. You can check this, Ivan.

gromakovsky:
I think it's good enough now and can be merged.

PR #22, opened 2020-10-09
Pru stephan/#6 implement fa2 functionality

### LINKED ISSUE #6: Implement FA2 functionality

PruStephan:
## Description

This PR adds two FA2 Entrypoints - `balance_of` and `transfer`
## Related issue(s)

Resolves #6

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/Lorentz/Contracts/BaseDAO.hs
+++ b/src/Lorentz/Contracts/BaseDAO.hs
@@ -38,7 +38,7 @@ baseDaoContract = defaultContract $ contractName "FA2 smart contract" $ do
 fa2Handler :: Entrypoint FA2.Parameter Storage
 fa2Handler =
   entryCase @FA2.Parameter (Proxy @PlainEntrypointsKind)
-    ( #cTransfer /-> transfer
+    ( #cTransfer /-> transfer defaultPermissionsDescriptor
     , #cBalance_of /-> balanceOf
     , #cToken_metadata_registry /-> tokenMetadataRegistry
     , #cUpdate_operators /-> updateOperators
--- a/src/Lorentz/Contracts/BaseDAO.hs
+++ b/src/Lorentz/Contracts/BaseDAO.hs
@@ -38,7 +38,7 @@ baseDaoContract = defaultContract $ contractName "FA2 smart contract" $ do
 fa2Handler :: Entrypoint FA2.Parameter Storage
 fa2Handler =
   entryCase @FA2.Parameter (Proxy @PlainEntrypointsKind)
-    ( #cTransfer /-> transfer defaultPermissionsDescriptor
+    ( #cTransfer /-> transfer
     , #cBalance_of /-> balanceOf
     , #cToken_metadata_registry /-> tokenMetadataRegistry
     , #cUpdate_operators /-> updateOperators
--- a/src/Lorentz/Contracts/BaseDAO/FA2.hs
+++ b/src/Lorentz/Contracts/BaseDAO/FA2.hs
@@ -5,7 +5,6 @@ module Lorentz.Contracts.BaseDAO.FA2
    ( transfer
    , balanceOf
    , tokenMetadataRegistry
-   , permissionsDescriptor
    , updateOperators
    , defaultPermissionsDescriptor
    ) where
@@ -22,31 +21,40 @@ defaultPermissionsDescriptor = PermissionsDescriptor
   , pdCustom = Nothing
   }
 
--- | TODO: Implement missing methods
-transfer :: PermissionsDescriptor -> Entrypoint TransferParams Storage
-transfer pdsec = do
+transfer :: Entrypoint TransferParams Storage
+transfer = do
   iter transferItem
   nil; pair
   where
+    transferItem :: forall s. TransferItem & Storage & s :-> Storage & s
     transferItem = do
-      getField #tiFrom
       swap
+      getField #sAdmin
+      Lorentz.sender
+      if IsEq
+      then push True
+      else push False
+      dig @2
       getField #tiTxs
       swap
+      getField #tiFrom
+      swap
       drop @TransferItem
+      dug @2
       iter transferOne
-      drop @Address
+      dropN @2
 
-    transferOne :: forall s. TransferDestination & Address & Storage & s :-> Address & Storage & s
+    transferOne :: forall s. TransferDestination & Bool & Address & Storage & s :-> Bool & Address & Storage & s
     transferOne = do
-      getField #tdTokenId
       swap
+      dup
+      if_ nop $ do
+        dug @3
+        dug @3
+        checkSender
+        dig @3
+        dig @3
       dug @3
-      swap
-      checkAdmin (pdOperator pdsec)
-      swap
-      drop @TokenId
-      dig @2
       getField #tdTo
       swap
       getField #tdAmount
@@ -56,9 +64,14 @@ transfer pdsec = do
       drop @TransferDestination
       dup
       push @Natural 0
-      assertEq [mt|#FA2_TOKEN_UNDEFINED|]
-      swap
-      swap
+      if IsEq
+      then nop
+      else do
+        dup
+        push @Natural 1
+        if IsEq
+        then failCustom_ #fROZEN_TOKEN_NOT_TRANSFERABLE
+        else failCustom_ #fA2_TOKEN_UNDEFINED
       pair
       dup
       dig @3
@@ -69,35 +82,29 @@ transfer pdsec = do
       dig @4
       debitFrom;
       creditTo;
+      dug @2
       swap
 
-    checkAdmin
-      :: forall f. OwnerOrOperatorTransfer
-      ->  Address & TokenId & Storage & f
-      :-> Address & TokenId & Storage & f
-    checkAdmin NoTransfer = failUsing [mt|#fA2_TX_DENIED|]
-    checkAdmin OwnerTransfer = do
-      dup
-      Lorentz.sender
-      if IsEq
-      then nop
-      else failUsing [mt|#fA2_NOT_OWNER|]
-    checkAdmin OwnerOrOperatorTransfer = do
+    checkSender
+      :: forall f. Address & Storage & f
+      :-> Address &  Storage & f
+    checkSender = do
       dup
       Lorentz.sender
       if IsEq
       then nop
       else do
-        dupTop2
-        pair
-        Lorentz.sender
-        pair
-        dig @3
-        getField #sOperators
+        dup
         dig @2
+        getField #sOperators
+        swap
+        dug @3
+        swap
+        Lorentz.sender
+        swap
+        pair
         mem
-        if_ (dug @2) (failUsing [mt|#fA2_NOT_OPERATOR|])
-
+        if_ nop (failCustom_ #fA2_NOT_OPERATOR)
 
 
 debitFrom :: forall s. Storage & Address & (TokenId, Natural) & s :-> Storage & s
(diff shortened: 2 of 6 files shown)


src/Lorentz/Contracts/BaseDAO.hs
```diff
@@ -38,7 +38,7 @@ baseDaoContract = defaultContract $ contractName "FA2 smart contract" $ do
 fa2Handler :: Entrypoint FA2.Parameter Storage
 fa2Handler =
   entryCase @FA2.Parameter (Proxy @PlainEntrypointsKind)
-    ( #cTransfer /-> transfer
+    ( #cTransfer /-> transfer defaultPermissionsDescriptor
```
sras, line 41: It appears the as per the spec, the contract implements the FA2 specified default permissions description, so at least that aspect does not have to be configurarble, so we don't have to do check the permissions descriptor values while doing permission checks during transfer. So we probably don't need this code https://github.com/tqtezos/baseDAO/pull/22/files#diff-454ccfe639f4a10802990f520295d821b3d444efa51dac4690b092cbdb1cd158R74-R99, and the default behavior can be just hardcoded.

sras:
@PruStephan I think we should also implement permissions_descriptor entrypoint as well here https://github.com/tqtezos/baseDAO/blob/52f2dfbeada956f4378e2ea11dedd828f555aa4c/src/Lorentz/Contracts/BaseDAO/FA2.hs#L231

src/Lorentz/Contracts/BaseDAO.hs
PruStephan: Okay, but we still have to do this checks, don't we? 
```
  checkAdmin OwnerOrOperatorTransfer = do
      dup
      Lorentz.sender
      if IsEq
      then nop
      else do
        dupTop2
        pair
        Lorentz.sender
        pair
        dig @3
        getField #sOperators
        dig @2
        mem
        if_ (dug @2) (failUsing [mt|#fA2_NOT_OPERATOR|])
```

PruStephan: About this 
> I think we should also implement permissions_descriptor entrypoint as well here

I think, that since it is not part of FA2 it should be left for another issue or something.

sras: > Okay, but we still have to do this checks, don't we?

Yes. 

> I think, that since it is not part of FA2 it should be left for another issue or something.

You are right. 

The entrypoint is part of FA2,

https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#exposing-permissions-descriptor

but as the spec say, FA2 allows us to omit the entrypoint if the contract implements default permissions. And we implement default permissions here so we don't need to include the entrypoint here. 

So I suppose you can remove the `permissionsDescriptor` function.

PruStephan: How to fail in Lorentz, with a custom error, that requires some params like  `failCustom_ #fA2_INSUFFICIENT_BALANCE` with `("required" :! Natural, "present" :! Natural)` arguments

gromakovsky:
> I think we should also implement permissions_descriptor entrypoint as well here 

AFAIU we should remove it since we are using the default one.

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -5,29 +5,257 @@ module Lorentz.Contracts.BaseDAO.FA2
...
+  nil; pair
+  where
+    transferItem :: forall s. TransferItem & Storage & s :-> Storage & s
+    transferItem = do
+      swap
+      getField #sAdmin
+      Lorentz.sender
+      if IsEq
+      then push True
+      else push False
```
sras, line 36: I suppose we can just use `eq` here. It should push the result to stack automatically.

PruStephan: Changed

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -129,5 +353,5 @@ assertNeq0or1 = do
     dup
     push @Natural 1
     if IsEq
-    then failUsing [mt|FA2_OPERATION_PROHIBITED|]
-    else failUsing [mt|FA2_TOKEN_UNDEFINED|]
+    then failUsing [mt|fA2_OPERATION_PROHIBITED|]
```
sras, line 356: Here we seem to fail on both branches of the conditional, which seems wrong to me. Also I am not sure there is a `fA2_OPERATION_PROHIBITED` error in FA2 spec. Also, since all tests pass now, it seems that we have some uncovered cases in tests.

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -5,29 +5,255 @@ module Lorentz.Contracts.BaseDAO.FA2
...
+      drop @TransferDestination
+      dup
+      push @Natural 0
+      if IsEq
+      then nop
+      else do
+        dup
+        push @Natural 1
+        if IsEq
+        then failCustom_ #fROZEN_TOKEN_NOT_TRANSFERABLE
```
sras, line 71: >#fROZEN_TOKEN_NOT_TRANSFERABLE

It seems that this error is not mentioned in the spec.

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: This one is custom. The spec doesn't say anything about error in case there is a `tId` equal 1 in `updateOperators` entrypoint

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: It is however, mentioned in tests

src/Lorentz/Contracts/BaseDAO/FA2.hs
sras: If it is custom, then I think it shouldn't be prefixed with `fA2`.  Also,

> Here we seem to fail on both branches of the conditional, which seems wrong to me.
> Since all tests pass now, it seems that we have some uncovered cases in tests.

src/Lorentz/Contracts/BaseDAO/FA2.hs
sras: Yes, appear to be a missed TODO

https://github.com/tqtezos/baseDAO/blob/master/test/Test/BaseDAO/FA2.hs#L270

I think we can include it in this PR.

gromakovsky: If some error is missing in our baseDAO specification, please add it there. It says the following:
> The list of erros may be inaccurate and incomplete, it will be updated during the implementation.

It's hard to list all errors in advance.

src/Lorentz/Contracts/BaseDAO/FA2.hs
```diff
@@ -5,29 +5,255 @@ module Lorentz.Contracts.BaseDAO.FA2
...
+  , pdCustom = Nothing
+  }
+
 transfer :: Entrypoint TransferParams Storage
-transfer = doNothing
+transfer = do
+  iter transferItem
+  nil; pair
+  where
+    transferItem :: forall s. TransferItem & Storage & s :-> Storage & s
```
rinn7e, line 29: I believe it is better to use explicit stack here.

```haskell
transferItem :: forall s. '[TransferItem, Storage] :-> '[Storage]
```

The reason being that, using `TransferItem & Storage & s` is allowed when the stack is,  for example: `'[TransferItem, Storage, Storage]`.

This result in `[Storage, Storage]` where one is different from another, and can be hard to track down if we somehow use wrong `Storage` down later on.

 Using `'[TransferItem, Storage] :-> '[Storage]` guarantee that we get only 1 storage left, so we are sure that this is the `Storage` with updated state.

I think this applies to __all functions that modify storage value__.

I believe using `TransferItem & Storage & s` is more fitted in a case where we simply want to check some condition and not updating storage. We do not care about the storage value in this case.

rinn7e (changes requested):

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: How to apply it correctly? I mean, if I update functions with that, I get a bunch of errors, as compiler can't deduce stack correctly

rinn7e: @PruStephan Which function that it does not work? I tried with `transferItem` , it works fine for me:

```
    transferItem :: '[TransferItem, Storage] :-> '[Storage]
    transferItem = do
      swap
      getField #sAdmin
      Lorentz.sender
      eq
      dig @2
      getField #tiTxs
      swap
      getField #tiFrom
      swap
      drop @TransferItem
      dug @2
      iter transferOne
      dropN @2
```

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: Maybe here we can fail with the same `#fROZEN_TOKEN_NOT_TRANSFERABLE` since it actually the same situation

src/Lorentz/Contracts/BaseDAO/FA2.hs
rinn7e: Change your `transferOne` type to this: 

```haskell
transferOne :: '[TransferDestination, Bool, Address, Storage] :-> '[Bool, Address, Storage]
```

Then you can inspect `debitFrom` and `creditTo` type.

```haskell
transferOne = do
      ... 
      stackType @'[Storage, Address, (TokenId, Natural), Address, (TokenId, Natural), Address, Bool];
      debitFrom;

      stackType @'[Storage, Address, (TokenId, Natural), Address, Bool];
      creditTo;

      stackType @'[Storage, Address, Bool];
```

As you can see, there are more items at the end, you have to explicitly list all of them:

```haskell
debitFrom
  ::  '[Storage, Address, (TokenId, Natural), Address, (TokenId, Natural), Address, Bool]
  :-> '[Storage, Address, (TokenId, Natural), Address, Bool]
```

Same applies to `creditTo`.

I know it may be not pretty this way, but we can optimize it later on if necessary.

rinn7e: Now that I think about it, we shouldn't use explicit stack on `debitFrom` and `creditTo` since they are meant to be reusable. 

So forget about my above comment. Only use explicit stack on function that modify storage and is not meant to be resuable.

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: Or maybe there would be better to change this exception for something like `#fROZEN_TOKEN_OPERATION_PERMITED` to use it everywhere, where operation can't be done with frozen token

src/Lorentz/Contracts/BaseDAO/FA2.hs
PruStephan: Updated

rinn7e (approved):

PR #27, opened 2020-10-26
#26 update transfer check

### LINKED ISSUE #26: Update transfer check

Our specification says that tokens of type 1 can't be transferred at all, `transfer` always fails. However, recently we were told that the admin should be able to transfer frozen tokens.
> tokens of types 0 (unfrozen) can be transferred by owner, operator and admin, tokens of types 1 (frozen) can be transferred only by admin

# Acceptance criteria

1. Admin can transfer frozen tokens.
2. The specification is updated accordingly.

Note that now it's possible that I freeze 100 tokens to vote, then voting ends and I should get my 100 tokens back (i. e. they should be unfrozen), but I can have less than 100 frozen tokens if the admin transfers my frozen tokens somewhere. I think that in such case (when someone has less frozen tokens than should be unfrozen) we should unfreeze all available tokens. It's a topic for another issue though (#14).

PruStephan:
## Description

Previously, none could transfer frozen token, but the latest spec stated that admin can. This PR adds this opportunity to transfer entrypoint

## Related issue(s)

Resolves #26 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -196,6 +196,7 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 | `NO_PENDING_ADMINISTRATOR_SET` | Throws when trying to authorize as the pending administrator whilst is not set for a contract |
 | `NOT_TOKEN_OWNER`              | Trying to configure operators for a different wallet which sender does not own                |
 | `FAIL_PROPOSAL_CHECK`          | Throws when trying to propose a proposal that does not pass `proposalCheck`               |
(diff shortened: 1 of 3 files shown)


gromakovsky (approved):
LGTM

PruStephan:
I rebased on the lattest master and squashed all commits, can I merge now? (Just to be sure.....)

gromakovsky:
Yes, I don't mind.

PR #33, opened 2020-11-02
[#15] Implement non-FA2 token functions

### LINKED ISSUE #15: Implement non-FA2 token functions

rinn7e:
Problem: There are other token entrypoints that are missing:
`mint`, `burn`, `transfer_contract_tokens`, and `token_address`.

Solution: Implements all those missing entrypoints, and add tests for
them.

## Related issue(s)

Resolves #15 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -194,6 +194,7 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 | `NOT_PENDING_ADMINISTRATOR`    | Authorized sender is not the current pending administrator   |
 | `NO_PENDING_ADMINISTRATOR_SET` | Throws when trying to authorize as the pending administrator whilst is not set for a contract |
 | `NOT_TOKEN_OWNER`              | Trying to configure operators for a different wallet which sender does not own                |
+| `FAIL_TRANSFER_CONTRACT_TOKENS` | Trying to cross-transfer DAO tokens to another contract that does not exist or is not a valid FA2 contract. |
 | `FAIL_PROPOSAL_CHECK`          | Throws when trying to propose a proposal that does not pass `proposalCheck`               |
 | `FROZEN_TOKEN_NOT_TRANSFERABLE`| Transfer entrypoint is called for frozen token by a non-admin sender|
 | `PROPOSAL_INSUFFICIENT_BALANCE`| Throws when trying to propose a proposal without having enough unfrozen token                |
@@ -436,6 +437,7 @@ Functions related to token transfers, but not present in FA2. They do not have `
 data MintParam = MintParam
   { to_ :: Address
   , amount :: Natural
+  , token_id :: TokenId
   }
 
 data Parameter proposalMetadata
@@ -446,7 +448,10 @@ Parameter (in Michelson):
 ```
 (pair
   (address %to_)
-  (nat %value)
+  (pair
+    (nat %amount)
+    (nat %token_id)
+  )
 )
 ```
 
@@ -457,8 +462,9 @@ Parameter (in Michelson):
 
 ```haskell
 data BurnParam = BurnParam
-  { to_ :: Address
+  { from :: Address
   , amount :: Natural
+  , token_id :: TokenId
   }
 
 data Parameter proposalMetadata
@@ -468,8 +474,11 @@ data Parameter proposalMetadata
 Parameter (in Michelson):
 ```
 (pair
-  (address %to_)
-  (nat %value)
+  (address %from)
+  (pair
+    (nat %amount)
+    (nat %token_id)
+  )
 )
 ```
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -190,7 +190,7 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 
 | Error                          | Description                                            |
 |--------------------------------|--------------------------------------------------------|
-| `NOT_ADMINISTRATOR`            | The sender is not the administrator                          |
+| `NOT_ADMIN`            | The sender is not the administrator                          |
 | `NOT_PENDING_ADMINISTRATOR`    | Authorized sender is not the current pending administrator   |
 | `NO_PENDING_ADMINISTRATOR_SET` | Throws when trying to authorize as the pending administrator whilst is not set for a contract |
 | `NOT_TOKEN_OWNER`              | Trying to configure operators for a different wallet which sender does not own                |
@@ -456,7 +456,7 @@ Parameter (in Michelson):
 ```
 
 - Produces the given amounts of tokens to the wallet associated with the given address.
-- Fails with `NOT_ADMINISTRATOR` if the sender is not the administrator.
+- Fails with `NOT_ADMIN` if the sender is not the administrator.
 
 ### **burn**
 
@@ -483,7 +483,7 @@ Parameter (in Michelson):
 ```
 
 - Reduce the given amounts of tokens to the wallet associated with the given address.
-- Fails with `NOT_ADMINISTRATOR` if the sender is not the administrator.
+- Fails with `NOT_ADMIN` if the sender is not the administrator.
 - Fails with `FA2_INSUFFICIENT_BALANCE` if the wallet associated with the given address
 does not have enough tokens to burn.
 
@@ -531,7 +531,7 @@ Parameter (in Michelson):
 ```
 
 - This entrypoint can be used by the administrator to transfer tokens owned (or operated) by this contract in another FA2 contract.
-- Fails with `NOT_ADMINISTRATOR` if the sender is not the administrator.
+- Fails with `NOT_ADMIN` if the sender is not the administrator.
 - If the outermost address passed to this entrypoint is a smart contract with FA2 `transfer` entrypoint, this entrypoint is called with supplied argument.
 That is, the list of operations returned from the baseDAO contract should contain one `TRANSFER_TOKENS` operation calling the `transfer` entrypoint.
 Otherwise the call fails.
@@ -556,7 +556,7 @@ address
 
 - Initiate transfer of the role of administrator to a new address.
 
-- Fails with `NOT_ADMINISTRATOR` if the sender is not the administrator.
+- Fails with `NOT_ADMIN` if the sender is not the administrator.
 
 - The current administrator retains his privileges up until
   `accept_ownership` is called.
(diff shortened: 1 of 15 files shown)


rinn7e:
@gromakovsky Just wanted to let you know that I've added all the entrypoints and modify some places of the spec. 

It's not finished yet, I'll be working on adding tests for them next.

src/Lorentz/Contracts/BaseDAO/NonFA2.hs
```diff
@@ -0,0 +1,75 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Proposal related entrypoints
+module Lorentz.Contracts.BaseDAO.NonFA2
```
gromakovsky, line 5: `NonFA2` sounds too general because for example `Management` is non-FA2 as well. Here we have non-FA2 **token** functions. What if we call this module `Lorentz.Contracts.BaseDAO.Token` and move FA2 functionality into `Lorentz.Contracts.BaseDAO.Token.FA2`?

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -345,6 +353,44 @@ data VoteParam = VoteParam
...
+instance TypeHasDoc BurnParam where
+  typeDocMdDescription = "Describes whose account, which token id and in what amount to burn"
+
+data MintParam = MintParam
+  { mTo :: Address
+  , mValue :: Natural
+  , mTokenId :: FA2.TokenId
+  }
+  deriving stock (Generic, Show)
+  deriving anyclass (IsoValue, HasAnnotation)
```
gromakovsky, line 377: AFAIU in auto-generated `HasAnnotation` field annotations will have `m` prefix, right? In the specification there is no such prefix.

gromakovsky (changes requested):

test/Test/BaseDAO/Token.hs
```diff
@@ -0,0 +1,106 @@
...
+                } ]
+            } ]
+      param = DAO.TransferContractTokensParam
+        { DAO.tcContractAddress = toAddress dao
+        , DAO.tcParams = transferParams
+        }
+
+  -- TODO: Figure out why FA2_NOT_OPERATOR occurs
+  callFrom (AddressResolved owner1) dao (Call @"Transfer_contract_tokens") param
+   & expectCustomError_ #fA2_NOT_OPERATOR
```
rinn7e, line 95: I've dropped the prefix for various params to match the spec.

I also rebased and use `nettestScenario` instead.

I've used `contractCalling` to implement `transferContractTokens` entrypoints, but there are 2 problems that I found:
1. `contractCalling` returns `Maybe a`, in case of error, which error should I use in this case.
2. From my understanding, the above test should work fine. However this causes `FA2_NOT_OPERATOR`. 

So my question is, when we use `contractCalling` from a `contractA` to call `Transfer` entrypoint of `contractB`. Is the sender address the contractA's address itself or the original sender of contractA ?

src/Lorentz/Contracts/BaseDAO/Token.hs
```diff
@@ -0,0 +1,75 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Proposal related entrypoints
```
gromakovsky, line 4: Looks wrong.

src/Lorentz/Contracts/BaseDAO/Token.hs
```diff
@@ -0,0 +1,75 @@
...
+    push zeroMutez
+  transferTokens
+  nil; swap; cons
+  pair
+
+tokenAddress
+  :: forall store pm. (HasAnnotation pm, NiceParameter pm, StorageC store pm)
+  => Entrypoint TokenAddressParam store
+tokenAddress = do
+  -- dip ensureNotMigrated -- TODO
```
gromakovsky, line 71: 1. Does it not work?
2. Shouldn't `ensureNotMigrated` be called in other entrypoints?

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -329,6 +340,9 @@ instance (TypeHasDoc pm, IsoValue pm) => TypeHasDoc (ProposeParams pm) where
   typeDocHaskellRep = concreteTypeDocHaskellRep @(ProposeParams ())
   typeDocMichelsonRep = concreteTypeDocMichelsonRep @(ProposeParams ())
 
+instance HasAnnotation pm => HasAnnotation (ProposeParams pm) where
+  annOptions = defaultAnnOptions { fieldAnnModifier = dropPrefixThen toSnake }
```
gromakovsky, line 344: I think it makes sense to extract `defaultAnnOptions { fieldAnnModifier = dropPrefixThen toSnake }` into a top-level constant in this module since it's used in many places.

test/Test/BaseDAO/Token.hs
```diff
@@ -0,0 +1,106 @@
...
+import Util.Named
+
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Test.Common
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+test_BaseDAO_Token :: TestTree
+test_BaseDAO_Token = testGroup "BaseDAO FA2 tests:"
```
gromakovsky, line 24:
```suggestion
test_BaseDAO_Token = testGroup "BaseDAO non-FA2 token tests:"
```

test/Test/BaseDAO/Token.hs
```diff
@@ -0,0 +1,106 @@
...
+  [ testGroup "Admin:"
+    [ nettestScenario "can burn tokens from any accounts" adminBurnScenario
+    , nettestScenario "can mint tokens from any accounts" adminMintScenario
```
gromakovsky, line 27:
```suggestion
    , nettestScenario "can mint tokens to any accounts" adminMintScenario
```

test/Test/BaseDAO/Token.hs
gromakovsky: > contractCalling returns Maybe a, in case of error, which error should I use in this case.

I think there are 2 options:
1. Add a new error and fail with it.
2. Take `contract [TransferItem]` parameter instead of `address`. If you have a `contract`, you don't need to call `contractCalling`.

> So my question is, when we use contractCalling from a contractA to call Transfer entrypoint of contractB. Is the sender address the contractA's address itself or the original sender of contractA ?

It should be `contractA`'s address, so I am surprised. I think the test should pass. I don't know why it fails, apparently it needs deeper investigation.

gromakovsky (changes requested):
Also apparently now some tests fail on CI due to gas exhaustion or too large operations :disappointed:

src/Lorentz/Contracts/BaseDAO/Token.hs
rinn7e: I was mean as a TODO to add `ensureNotMigrated` to all entrypoints except `transferContractTokens`. I believe, it is needed for all entrypoint. Is that correct, Ivan?

test/Test/BaseDAO/Token.hs
rinn7e: @gromakovsky  For point 1, I have added the new error. 

For point 2, I discovered that we use wrong logic to check the operators in `transfer` entrypoint. I have added annotation and fix the bug as well as adding some regression tests. 

Continue from our discussion in slack: https://serokell.slack.com/archives/C019PCGKTK2/p1604402597008100?thread_ts=1604393896.004100&cid=C019PCGKTK2

The too big contract size error is still an issue even after removing proposalMetadata entrypoint. I have searched but no solution coming up. Maybe someone works at tezos could help. The final solution is probably to split this contract into 2 but I'm not sure how hard it will be.

docs/specification.md
```diff
@@ -194,6 +194,7 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 | `NOT_PENDING_ADMINISTRATOR`    | Authorized sender is not the current pending administrator   |
```
gromakovsky, line 194: Don't forget to update `NOT_ADMINISTRATOR` above.

docs/specification.md
```diff
@@ -446,7 +448,10 @@ Parameter (in Michelson):
 ```
 (pair
   (address %to_)
```
gromakovsky, line 450: Why do we have `from` in `burn`, but `to_` (with `_`) in `mint`?

src/Lorentz/Contracts/BaseDAO/Token.hs
gromakovsky: Yes, it's correct. `transferContractTokens` should still be callable after migration because the contract may have some FA2 tokens after migration (and we don't want to lose them).

gromakovsky:
It looks ok now and tests pass, that's great. Please rebase to make better commit history (e. g. rename of `NOT_ADMINISTRATOR` and change of the port should be in separate commits) and I'll make a final pass (also see 2 small comments of mine).

docs/specification.md
rinn7e: I've updated this.

docs/specification.md
rinn7e: I've changed to use `from_` instead now.

rinn7e:
> It looks ok now and tests pass, that's great. Please rebase to make better commit history (e. g. rename of `NOT_ADMINISTRATOR` and change of the port should be in separate commits) and I'll make a final pass (also see 2 small comments of mine).

I've rebased, clean up the commits and handle other cases. You can check this now, Ivan.

test/Test/BaseDAO/Proposal.hs
```diff
@@ -57,7 +57,9 @@ test_BaseDAO_Proposal = testGroup "BaseDAO propose/vote entrypoints tests:"
           flushWithBadConfig
       -- TODO [#15]: admin burn proposer token and test "flush"
 
-      , nettestScenario "flush and run decision lambda" flushDecisionLambda
+      -- TODO: Improve this when contract size is smaller
```
gromakovsky, line 60: Can you add #38 reference please?

gromakovsky (approved):

PR #37, opened 2020-11-03
[#30] Use `advanceTime` from `Nettest` to properly test `flush` entrypoint 

### LINKED ISSUE #30: Use `advanceTime` from `Nettest` to properly test `flush` entrypoint when it is implemented

Currently, it is not possible to do a `wait 3 seconds` in `Nettest`. We need this to actually check if a voting period of a proposal is over or not. Currently, there is a flag to force voting period to be over to be able to test the entrypoint.

# Acceptance criteria
- Remove the workaround: `cForceVotingPeriod`
- Use `advanceTime` fron `Nettest` to wait for a proposal's voting period to be over.
- All TODO related #30 is removed.

rinn7e:
## Description

Problem: Previously, it is not possible to do a `wait 3 seconds` in `Nettest`.
We need this to actually check if a voting period of a proposal is over or not.

Solution: Use `advanceTime` to properly test `flush` entrypoint. Remove
the workaround.

## Related issue(s)

Resolves #30 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/test/Test/BaseDAO/Proposal.hs
+++ b/test/Test/BaseDAO/Proposal.hs
@@ -13,6 +13,7 @@ import Morley.Nettest
 import Morley.Nettest.Tasty
 import Test.Tasty (TestTree, testGroup)
 import Tezos.Crypto (blake2b)
+import Time (sec)
 
 import qualified Lorentz.Contracts.BaseDAO.Types as DAO
 import Test.BaseDAO.ProposalConfig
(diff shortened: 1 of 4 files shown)


test/Test/BaseDAO/Proposal.hs
```diff
@@ -396,6 +404,7 @@ flushDecisionLambda = uncapsNettest $ do
         }
   callFrom (AddressResolved owner2) dao (Call @"Vote") [upvote]
 
+  advanceTime (sec 10)
```
rinn7e, line 407: @gromakovsky I have removed the workaround and use `advanceTime` instead. It seems to work on emulator but in the real network, it seems to not advance the time at all.

Hope to get your advice on whether or not I use the function correctly.

rinn7e: Actually the reason was because I set the `voting_period` to short (2 seconds). I've updated this to (20 seconds), and it works fine now.

Hope to get your review on this.

test/Test/BaseDAO/Proposal.hs
```diff
@@ -231,6 +251,11 @@ flushNotAffectOngoingProposals :: (Monad m) => NettestImpl m -> m ()
 flushNotAffectOngoingProposals = uncapsNettest $ do
   ((owner1, _), _, dao, admin) <- originateBaseDaoWithConfig config
 
+  -- Note: Cannot set to few seconds, since in real network, each
+  -- calls takes some times to run. 20 seconds seem to be the ideal.
+  callFrom (AddressResolved admin) dao (Call @"Set_voting_period") 20
+
+  advanceTime (sec 3)
```
gromakovsky, line 258: It can wait for more than 3 seconds and in some cases it can wait for more than 20. AFAIU here the idea is to call `flush` before the voting period is over. Can we set voting period to a higher value, like 2 minutes for example?

test/Test/BaseDAO/Proposal.hs
```diff
@@ -195,6 +194,27 @@ insufficientTokenVote = uncapsNettest $ do
...
+
+  -- | Create sample proposal
+  key1 <- createSampleProposal 1 owner1 dao
+
+  callFrom (AddressResolved admin) dao (Call @"Set_voting_period") 20
+
+  let params = DAO.VoteParam
+        { vVoteType = True
+        , vVoteAmount = 2
+        , vProposalKey = (key1 :: ByteString)
```
gromakovsky, line 209: Is it necessary to specify the type explicitly here?

gromakovsky (commented):
Mostly LGTM, just a couple of questions.

test/Test/BaseDAO/Proposal.hs
rinn7e: It's not, I've removed this, Ivan.

test/Test/BaseDAO/Proposal.hs
rinn7e: Make sense, I've updated this.

gromakovsky (approved):

PR #40, opened 2020-11-05
[#38] Reduce contract size

### LINKED ISSUE #38: Optimize the contract

In #33 we are already observing that the contract is quite big and in some tests we fail to originate it because we exceed operation size limit (16384 bytes). If we originate it with empty storage, it usually works, but:
1. It works only for trivial DAO where proposal metadata is `()` and instructions in runtime config are trivial.
2. Maybe we'll need to add more functionality later.
3. It's inconvenient that we can't originate the contract with a small (but not empty) storage.
4. Apparently there are some easy ways to make the contract smaller.

# Acceptance criteria

1. Analysis is performed to figure out how the contract can be made smaller.
  * In particular, we'd like to "profile" the contract, i. e. figure out which parts of it are _the most expensive_. Splitting the contract into parts is a bit complicated because for example `creditTo` is used in multiple entrypoints. So it should be just some approximation. We are particularly interested in typechecking gas cost and total size of instructions in bytes. Maybe this profiling will indicate very inefficient places. Even if it doesn't this profiling is still generally useful for us.
2. The contract is made smaller according to the results of that analysis.

Martoon-00:
## Description

Applies optimizations to reduce contract size mentioned in #38.

This yet depends on [MR in Morley](https://gitlab.com/morley-framework/morley/-/merge_requests/651)

## Related issue(s)

Resolves (partially) #38

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Misc
  - [x] Revisions of the libraries we depend on are either on hackage or in `master` branch of the respective repository.

### DIFF
--- a/src/Lorentz/Contracts/BaseDAO.hs
+++ b/src/Lorentz/Contracts/BaseDAO.hs
@@ -16,10 +16,11 @@ module Lorentz.Contracts.BaseDAO
   ) where
 
 import Lorentz
-import Lorentz.Contracts.BaseDAO.Token
-import Lorentz.Contracts.BaseDAO.Token.FA2
+
 import Lorentz.Contracts.BaseDAO.Management
 import Lorentz.Contracts.BaseDAO.Proposal
+import Lorentz.Contracts.BaseDAO.Token
+import Lorentz.Contracts.BaseDAO.Token.FA2
 import Lorentz.Contracts.BaseDAO.Types as BaseDAO
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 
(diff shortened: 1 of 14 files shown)


src/Lorentz/Contracts/BaseDAO.hs
```diff
@@ -65,3 +70,17 @@ ensureZeroTransfer = do
   amount
   push zeroMutez
   if IsEq then nop else failCustom_ #fORBIDDEN_XTZ
+
+pushFuncContext
```
rinn7e, line 74: Would be nice to have some description of this function.

rinn7e (changes requested):
The rest looks good to me. Would be nice to make the build in CI pass so that we know if there is anything wrong within the tests or not.

Martoon-00:
As far as I got, we have to wait for Morley's CI, as our CI here recognizes only commits from Morley's master. So waiting )

Martoon-00:
Okay, started to depend on released versions again and this will break CI, but before that I got CI passed and it be checked.

Also, I added a commit about tests on operation size limit, it is small but debatable; hope it makes sense.

gromakovsky:
Does it compile locally now?

Martoon-00:
> Does it compile locally now?

Yep, sure.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -267,7 +267,7 @@ submitVote = do
 
   constructT @(Proposal pm)
     ( fieldCtor $ do
-        -- | Update upvote
+        -- Update upvote
```
gromakovsky, line 272: Hm, if it's invalid Haddock and CI doesn't check it, I think we should add a CI check to prevent invalid Haddock in future. See https://gitlab.com/morley-framework/morley/-/merge_requests/620/diffs for example.

src/Lorentz/Contracts/BaseDAO.hs
```diff
@@ -65,3 +70,17 @@ ensureZeroTransfer = do
...
+  :: (KnownValue pm)
+  => Proxy pm
```
gromakovsky, line 75: Do we really need to pass `Proxy` if we have type applications?

test/Test/BaseDAO/Limits.hs
```diff
@@ -0,0 +1,66 @@
...
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.HUnit (testCase)
+
+import Lorentz.OpSize
+import Lorentz.Value
+import Tezos.Address
+
+import qualified Lorentz.Contracts.BaseDAO as DAO
+
+originationOpSize :: _ => DAO.Config pm -> DAO.Storage pm -> OpSize
```
gromakovsky, line 25: Why do we need `_ =>`?

Martoon-00: That constraint is extremely big and for the module-local method we don't care.

src/Lorentz/Contracts/BaseDAO.hs
Martoon-00: I don't like when mandatory arguments are passed via just `TypeApplications` because, when the argument is missing, this causes slightly unclear errors like `Cannot deduce 'instance IsoValue pm0'`. But ok, this method is used only in one place anyway, so there should be no problem in removing `Proxy`.

gromakovsky (approved):

rinn7e (approved):

PR #41, opened 2020-11-06
[#36] Add `GameDAO` contract as an example of `BaseDAO`

### LINKED ISSUE #36: Provide better example contract

Our `baseDAO` executable can print one simple DAO contract. This contract is parameterized with `defaultConfig` which is the simplest possible config. It doesn't check anything about proposals, its decision lambda does nothing, it doesn't slash returned amount. I think we need to provide a better example that could be practically useful. The initial document we received contains 4 DAOs: Registry DAO, BakerDAO, DEXDAO, GameDAO. We can implement one of them or come up with something else (inspired by these 4).

# Acceptance criteria

`baseDAO` executable can print something more practically useful than it prints now. I think we can still have the existing DAO and rename it to `trivialDAO` or something like that.

rinn7e:
## Description
Problem: Our `baseDAO` executable can print one simple DAO contract.
This contract is parameterized with `defaultConfig` which is the
simplest possible config. It doesn't check anything about proposals,
its decision lambda does nothing, it doesn't slash returned amount.
I think we need to provide a better example that could be practically useful.
The initial document we received contains 4 DAOs: Registry DAO, BakerDAO,
DEXDAO, GameDAO. We can implement one of them or come up with something else (inspired by these 4).

Solution: Use `GameDAO` from the google doc specification as example.
Add proper `proposalMetadata`, `proposalCheck`, and
`rejectedProposalValue` that makes sense for `GameDAO`.

## Related issue(s)

Resolves #36 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -21,6 +21,7 @@ import Util.CLI (mkCLOptionParser)
 import Util.Named
 
 import qualified Lorentz.Contracts.BaseDAO as DAO
+import qualified Lorentz.Contracts.GameDAO as GameDAO
 
 programInfo :: DGitRevision -> Opt.ParserInfo CmdLnArgs
 programInfo gitRev = Opt.info (Opt.helper <*> versionOption <*> argParser contracts gitRev) $
@@ -50,16 +51,23 @@ contracts = ContractRegistry $ Map.fromList
   [ "BaseDAO" ?:: ContractInfo
     { ciContract = DAO.baseDaoContract DAO.defaultConfig
     , ciIsDocumented = True
-    , ciStorageParser = Just baseDaoStorageParser
+    , ciStorageParser = Just $ baseDaoStorageParser @()
+    , ciStorageNotes = Nothing
+    }
+  , "GameDAO" ?:: ContractInfo
+    { ciContract = GameDAO.gameDaoContract
+    , ciIsDocumented = True
+    , ciStorageParser = Just $
+        baseDaoStorageParser @GameDAO.GameDaoProposalMetadata
     , ciStorageNotes = Nothing
     }
   ]
 
-baseDaoStorageParser :: Opt.Parser (DAO.Storage ())
+baseDaoStorageParser :: forall m. Opt.Parser (DAO.Storage m)
 baseDaoStorageParser = do
   adminAddress <-
     addressOption Nothing (#name .! "admin")
-    (#help .! "Administrator of the BaseDAO contract")
+    (#help .! "Administrator of the DAO contract")
   votingPeriod <-
     mkCLOptionParser (Just $ 60 * 60 * 24 * 7) (#name .! "voting-period")
     (#help .! "Period after which proposals can be finished")
--- a/src/Lorentz/Contracts/GameDAO.hs
+++ b/src/Lorentz/Contracts/GameDAO.hs
@@ -1,0 +1,196 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | A simple DAO for a Moba-like game.
+module Lorentz.Contracts.GameDAO
+  ( GameDaoProposalMetadata (..)
+  , Storage
+  , Parameter
+  , ProposalType (..)
+  , BalanceChange (..)
+  , ItemChange (..)
+  , HeroChange (..)
+  , NewContents
+  , NewContent (..)
+
+  , gameDaoContract
+  , config
+  ) where
+
+import Universum hiding ((>>), drop, swap)
+import Lorentz
+import qualified Lorentz.Contracts.BaseDAO as DAO
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Util.Markdown
+
+data GameDaoProposalMetadata = GameDaoProposalMetadata
+  { pmProposalType :: ProposalType
+  , pmProposalDescription :: MText
+  , pmConsumerAddr :: Address
+  }
+  deriving stock (Generic)
+  deriving anyclass IsoValue
+
+instance HasAnnotation GameDaoProposalMetadata where
+  annOptions = DAO.baseDaoAnnOptions
+
+data ProposalType = BalanceType BalanceChange | NewType NewContents
+  deriving stock (Generic)
+  deriving anyclass (IsoValue)
+
+instance HasAnnotation ProposalType where
+  annOptions = defaultAnnOptions { fieldAnnModifier = toSnake }
+
+instance TypeHasDoc ProposalType where
+  typeDocMdDescription = "There are 2 types of proposal that Game DAO supports. \
+    \'balance_type' is for small update to existing mechanic, items and heroes. \
+    \'new_type' is for proposing new contents all together."
+
+data BalanceChange = BalanceChange
+  { bcItemChanges :: [ItemChange]
+  , bcHeroChanges :: [HeroChange]
+  }
+  deriving stock (Generic)
+  deriving anyclass (IsoValue)
+
+instance HasAnnotation BalanceChange where
+  annOptions = DAO.baseDaoAnnOptions
+
+instance TypeHasDoc BalanceChange where
+  typeDocMdDescription = "Describe the update of all items and heroes in the proposal."
+
+data ItemChange = ItemChange
+  { icItemName :: MText
+  , icChangelogs :: [MText]
+  }
+  deriving stock (Generic)
+  deriving anyclass (IsoValue)
+
+instance TypeHasDoc ItemChange where
+  typeDocMdDescription = "Describe the update of a specific item."
+
+instance HasAnnotation ItemChange where
+  annOptions = DAO.baseDaoAnnOptions
+
+data HeroChange = HeroChange
+  { hcHeroName :: MText
+  , hcChangelogs :: [MText]
+  }
+  deriving stock (Generic)
+  deriving anyclass (IsoValue)
+
+instance TypeHasDoc HeroChange where
+  typeDocMdDescription = "Describe the update of a specific hero."
+
+instance HasAnnotation HeroChange where
+  annOptions = DAO.baseDaoAnnOptions
+
+type NewContents = [NewContent]
+
+data NewContent = NewContent
+  { ncContentName :: MText
+  , ncContentDescription :: [MText]
+  }
+  deriving stock (Generic)
+  deriving anyclass (IsoValue)
+
+instance TypeHasDoc NewContent where
+  typeDocMdDescription = "Describe a new content. The content should have a \
+    \name and detail description."
+
+instance HasAnnotation NewContent where
+  annOptions = DAO.baseDaoAnnOptions
+
+instance TypeHasDoc GameDaoProposalMetadata where
+  typeDocMdDescription = "GameDAO's metadata. This fields affect how proposal \
+    \got accepted and how many tokens will be frozen"
+
+----------------------------------------------------------------------------
+
+type instance ErrorArg "fAIL_DECISION_LAMBDA" = ()
+
+instance CustomErrorHasDoc "fAIL_DECISION_LAMBDA" where
+  customErrClass = ErrClassActionException
+  customErrDocMdCause = "Trying to execute decision lambda but result in errors."
+
+----------------------------------------------------------------------------
+
+type Parameter = DAO.Parameter GameDaoProposalMetadata
+type Storage = DAO.Storage GameDaoProposalMetadata
+
+-- | For "balance change", the proposer needs to freeze 10 tokens
+-- For "new content", the proposal needs to freeze 50 tokens
+gameDaoProposalCheck
+  :: forall s. (DAO.ProposeParams GameDaoProposalMetadata) : s
+  :-> Bool : s
+gameDaoProposalCheck = do
+  getFieldNamed #ppFrozenToken; swap
+  toField #ppProposalMetadata
+  toField #pmProposalType
+  stackType @(ProposalType : "ppFrozenToken" :! Natural : s)
+  caseT
+    ( #cBalanceType /-> do drop; push (10 :: Natural)
+    , #cNewType /-> do drop; push (50 :: Natural)
+    )
+  toNamed #requireValue
+
+  if #requireValue <=. #ppFrozenToken then
+    push True
+  else
+    push False
+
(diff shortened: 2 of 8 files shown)


rinn7e:
```
At the core of most games is not only continual updates but changes to a ‘meta’ or the most
optimal way to play a game to continually make it interesting. For a game to truly be
decentralized, the way in which a game is played and the assets associated with a game shouldbe subject to change based on the whims of a community that keeps the game alive. In the
case of a GameDAO, individual players of the game would be able to earn voting rights by
hitting particular milestones in a game and then vote on proposals that look to change the
game’s assets and playstyle. An example of this could be a card battle game where new cards
represented as NFTs can be added to the game, or the way in which you can modify ‘damage’
bonuses in the game.
Process:
1. The DAO would be instantiated and would distribute assets to players based on
particular attributes such as play-time or games played.
2. Proposals would be made on Agora and would specify the new assets to be
acknowledged or changes to gameplay in the game’s contract.
3. Audits would be necessary in order to garner voter support and would happen externally
to the DAO itself.
4. Proposal periods and voting periods would be rapid due to the introduction of new cards
and changing of a particular meta if there’s enough interest in a game.
5. Voting assets would continually be distributed based on the parameters set in terms of
how a user engages with the game itself.
```

Keeping all this point in mind, I implement a simple DAO for moba-like game. 

The proposer can propose balance changes or new content. Proposing new content require more tokens which mean the the proposer should be an experienced player thus having more token.

`cRejectedProposalValue` simply cut the frozen token in half the proposer.

I could not come up with an interesting `cDecismLambda` decision. 

I've submit the michelson code as well as the documentation of the contract. This will be removed if everything is good. 

Not sure if I should add tests for this or not, hope to get your advice on this @gromakovsky.

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -0,0 +1,163 @@
...
+  deriving anyclass (IsoValue)
+
+instance TypeHasDoc NewContent where
+  typeDocMdDescription = "Describe a new content. The content should have a name and detail description."
+
+instance HasAnnotation NewContent where
+  annOptions = DAO.baseDaoAnnOptions
+
+instance TypeHasDoc GameDaoProposalMetadata where
+  typeDocMdDescription = "GameDAO's metadata. This fields affect how proposal got accepted and how many tokens will be freezed"
```
gromakovsky, line 94:
```suggestion
  typeDocMdDescription = "GameDAO's metadata. This fields affect how proposal got accepted and how many tokens will be frozen"
```

app/Main.hs
```diff
@@ -52,6 +53,12 @@ contracts = ContractRegistry $ Map.fromList
     , ciStorageParser = Just baseDaoStorageParser
     , ciStorageNotes = Nothing
     }
+  , "GameDAO" ?:: ContractInfo
+    { ciContract = GameDAO.gameDaoContract
+    , ciIsDocumented = True
+    , ciStorageParser = Nothing
```
gromakovsky, line 59: I think it's better to provide a parser here, it should be similar to `baseDaoStorageParser` AFAIU.

gromakovsky (commented):
> Not sure if I should add tests for this or not

I think we need at least one test to make sure this GameDAO contract can be originated at least (i. e. doesn't exceed the operation size limit).

app/Main.hs
rinn7e: I have updated this, as well as adding tests for GameDAO. I'll wait to see if it passes the network test.

rinn7e: Unfortunately, the origination failed due to the large operation size. We probably have to wait for #40 to be merged.

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -0,0 +1,173 @@
...
+
+config :: DAO.Config GameDaoProposalMetadata
+config = DAO.defaultConfig
+  { DAO.cDaoName = "Game DAO"
+  , DAO.cDaoDescription = [md|A simple DAO for a Moba-like game.|]
+  , DAO.cProposalCheck = gameDaoProposalCheck
+  , DAO.cRejectedProposalReturnValue = gameDaoRejectedProposalReturnValue
+  , DAO.cDecisionLambda = do
+      -- TODO: Come up with interesting decision lambda
+      -- Can it call http endpoint?
```
Martoon-00, line 159: :joy_cat: :joy_cat: :joy_cat: 

That would be really interesting if it could

test/Test/GameDAO.hs
```diff
@@ -0,0 +1,125 @@
...
+          { icItemName = [mt|Ganto|]
+          , icChangelogs =
+              [ [mt|Ganto now has 10 charges.|]
+              , [mt|Ganto can now target enemey to reduce health regen.|]
+              ]
+          }]
+      , bcHeroChanges = []
+      }
+  , pmProposalDescription = [mt|Balance Item|]
+  }
```
Martoon-00, line 100: :pogchamp:

I recently asked @fersel whether can we have some gamedev-related project, and looks like here we have such partially.

src/Lorentz/Contracts/GameDAO.hs
Martoon-00: Maybe instead of `http` endpoint we could have a separate contract that stores game logs, or maybe top 5 proposals? and in `decisionLambda` we can call that contract to include the recent events.

test/Test/GameDAO.hs
```diff
@@ -0,0 +1,125 @@
...
+
+sampleMetadataContent :: GameDaoProposalMetadata
+sampleMetadataContent = GameDaoProposalMetadata
+  { pmProposalType = NewType [ NewContent
+      { ncContentName = [mt|TireDide|]
```
gromakovsky, line 105: :smile:

app/Main.hs
```diff
@@ -61,6 +68,13 @@ baseDaoStorageParser = do
     (#help .! "Administrator of the BaseDAO contract")
```
gromakovsky, line 68: Hm, if we change `the BaseDAO contract` to just `a DAO contract`, probably we'll be able to use this parser for both DAOs, right? I'd like to do it because we may (and will) have more shared fields configurable and thus more code will be copy-pasted.

src/Lorentz/Contracts/GameDAO.hs
gromakovsky: Calling another contract sounds interesting indeed.

test/Test/Common.hs
```diff
@@ -123,8 +133,8 @@ mkFA2View a c = FA2.FA2View (mkView a c)
 
 -- | Helper function to check a user balance of a particular token
 checkTokenBalance
-  :: MonadNettest caps base m
-  => FA2.TokenId -> TAddress (DAO.Parameter TestProposalMetadata)
+  :: (NiceStorage (Storage pm),NiceParameterFull (Parameter pm), MonadNettest caps base m)
```
gromakovsky, line 136:
```suggestion
  :: (NiceStorage (Storage pm), NiceParameterFull (Parameter pm), MonadNettest caps base m)
```

test/Test/Common.hs
```diff
@@ -72,16 +72,25 @@ originateBaseDaoWithBalance config balFunc = do
...
+  --   st = mkStorage admin mempty mempty
+
+  --   originateData :: forall pm. OriginateData (Parameter pm)
+  --   originateData = OriginateData
+  --     { odFrom = nettestAddress
+  --     , odName = "BaseDAO"
+  --     , odBalance = toMutez 0
+  --     , odStorage = mkStorage admin mempty mempty
+  --     , odContract = DAO.baseDaoContract config
+  --     }
```
gromakovsky, line 86: Is it going to be removed?

gromakovsky (changes requested):

app/Main.hs
rinn7e: I've rebased this MR with the latest master. The tests are still failing due to this error:

```
tezos-client unexpectedly failed with error code 1. Stdout:

          Node is bootstrapped, ready for injecting operations.

          

          Stderr:

          Fatal error:

            The operation will burn ꜩ13.403 which is higher than the configured burn cap (ꜩ13.389).

             Use `--burn-cap 13.403` to emit this operation.
```

This is a new error, I'm not sure how to fix this.

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -0,0 +1,202 @@
...
+  deriving anyclass (IsoValue)
+
+instance TypeHasDoc NewContent where
+  typeDocMdDescription = "Describe a new content. The content should have a name and detail description."
+
+instance HasAnnotation NewContent where
+  annOptions = DAO.baseDaoAnnOptions
+
+instance TypeHasDoc GameDaoProposalMetadata where
+  typeDocMdDescription = "GameDAO's metadata. This fields affect how proposal got accepted and how many tokens will be frozen"
```
gromakovsky, line 106: This line looks too long btw, could you please wrap it? Same for ` instance TypeHasDoc NewContent`.

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -0,0 +1,202 @@
...
+  :-> (List Operation, store) : s
+decisionLambda consumerAddr = do
+  toField #pMetadata
+  toField #pmProposalDescription
+
+  dip $ do
+    push consumerAddr
+    contractCallingUnsafe @MText DefEpName
+    ifSome nop $ do
+      failCustom_ #fAIL_TRANSFER_CONTRACT_TOKENS
```
gromakovsky, line 177: Is it supposed to be `fAIL_DECISION_LAMBDA`?

app/Main.hs
gromakovsky: :fffuuu:

Most likely it's a consequence of https://gitlab.com/morley-framework/morley/-/issues/288. It happens when you have an address constant in contract code. And apparently that's what we are doing in our decision lambda.

It seems quite hard to fix properly in `morley`, to me it looks more like a bug in Tezos. Anyway, there are two options:
1. Fix/workaround this problem in `morley`. I'll think about it tomorrow, probably there is an easy workaround.
2. Fetch `consumerAddr` from the storage. Currently we can't extend the storage with arbitrary data, but it was suggested by @Martoon-00 and probably we'll implement it soon. Then it will be possible to have `consumerAddr` in storage instead of contract code. Another option that is already possible is to store `consumerAddr` inside each proposal, which might be weird, but probably doesn't matter for illustration purposes.

rinn7e: I store `consumerAddr` in the proposal itself, and now the tests are passed now. I've also made other changes according to your review, you can check this.

gromakovsky: Oh, cool, then I think we can merge it as is now. Maybe later we'll move consumer address somewhere else.

gromakovsky (approved):

rinn7e:
@Martoon-00 Hope to get your re-review on this as well.

test/Test/GameDAO.hs
```diff
@@ -0,0 +1,136 @@
...
+      downvote = DAO.VoteParam
+        { vVoteType = False
+        , vVoteAmount = 1
+        , vProposalKey = key1
+        }
+  callFrom (AddressResolved owner2) dao (Call @"Vote") [upvote, downvote]
+  checkTokenBalance (DAO.frozenTokenId) dao owner2 3
+  checkTokenBalance (DAO.unfrozenTokenId) dao owner2 97
+
+  advanceTime (sec 61)
```
Martoon-00, line 79: If I understand correctly, our CI will run this with tezos-client, and `61` will mean that this test takes a minute. Maybe reduce the voting period a bit to make CI faster?

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -0,0 +1,203 @@
...
+  ::  forall s store. (DAO.StorageC store GameDaoProposalMetadata)
+  =>  (DAO.Proposal GameDaoProposalMetadata) : store : s
+  :-> (List Operation, store) : s
+decisionLambda = do
+  toField #pMetadata
+  getField #pmProposalDescription
+
+  dip $ do
+    toField #pmConsumerAddr
+    contractCallingUnsafe @MText DefEpName
```
Martoon-00, line 176: Minor: AFAIU you can have `TAddress MText` in `pmConsumerAddr`, and then use plain `contractCalling CallDef` here.

Martoon-00 (approved):
Looks good to me, I have only a couple of suggestions which you can ignore.

src/Lorentz/Contracts/GameDAO.hs
rinn7e: I tried to make this change, but it requires to add instance for `TypeHasDoc (TAddress MText)`. When I added it, this errors occur:

```
/home/rinne/project/serokell/baseDAO/src/Lorentz/Contracts/GameDAO.hs:110:10: error:
    • Type is not homomorphic: TAddress MText
    • In the expression:
        Michelson.Typed.Haskell.Doc.$dmtypeDocMdReference @(TAddress MText)
      In an equation for ‘typeDocMdReference’:
          typeDocMdReference
            = Michelson.Typed.Haskell.Doc.$dmtypeDocMdReference
                @(TAddress MText)
      In the instance declaration for ‘TypeHasDoc (TAddress MText)’
    |       
110 | instance TypeHasDoc (TAddress MText) where
```

test/Test/GameDAO.hs
rinn7e: It was made this way due to in real dephi network tests, there can be a lot of delay. However, we found out that it won't be helpful anyway and only enable all time-sensitive tests (flush) in emulator. 

So I think I'll reduce it to smaller number. When we are able to run operations in batch, we should be able enable these tests in real network.

src/Lorentz/Contracts/GameDAO.hs
Martoon-00: Oh, `TAddress` does not have this instance :fffuuu.

>     • In the expression:
>       Michelson.Typed.Haskell.Doc.$dmtypeDocMdReference @(TAddress MText)

`$dmtypeDocMdReference` means default implementation of `typeDocMdReference` so you need to initialize it with some other method. But ok, I think it does not worth doing right now.

test/Test/GameDAO.hs
gromakovsky: Oh, btw, it seems that currently this test runs not only in emulator. I think we should change it here as we did in #44.

rinn7e: I forgot about this. I've made so that it will run only in the emulator instead now.

PR #44, opened 2020-11-06
Fix failing tests in delphi due to too short voting period

rinn7e:
## Description

Problem: When running test in real delphi network, the voting period
(20s) is too short, since each block may take up to 30s.

Solution: Update voting_period to 60s.

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/test/Test/BaseDAO/Proposal.hs
+++ b/test/Test/BaseDAO/Proposal.hs
@@ -45,16 +45,18 @@ test_BaseDAO_Proposal = testGroup "BaseDAO propose/vote entrypoints tests:"
   , testGroup "Admin:"
       [ nettestScenario "can set voting period" setVotingPeriod
       , nettestScenario "can set quorum threshold" setQuorumThreshold
-      , nettestScenario "can flush proposals that got accepted" $
-          flushAcceptedProposals
-      , nettestScenario "can flush proposals that got rejected due to not meeting quorum_threshold" $
-          flushRejectProposalQuorum
-      , nettestScenario "can flush proposals that got rejected due to negative votes" $
-          flushRejectProposalNegativeVotes
+
+      -- TODO [#47]: Disable running in real network due to time-sensitive operations
+      , nettestScenarioOnEmulator "can flush proposals that got accepted" $
+          \_emulated -> flushAcceptedProposals
+      , nettestScenarioOnEmulator "can flush proposals that got rejected due to not meeting quorum_threshold" $
+          \_emulated -> flushRejectProposalQuorum
+      , nettestScenarioOnEmulator "can flush proposals that got rejected due to negative votes" $
+          \_emulated -> flushRejectProposalNegativeVotes
       , nettestScenario "flush should not affecting ongoing proposals"
           flushNotAffectOngoingProposals
-      , nettestScenario "flush with bad 'cRejectedProposalReturnValue'"
-          flushWithBadConfig
+      , nettestScenarioOnEmulator "flush with bad 'cRejectedProposalReturnValue'" $
+          \_emulated -> flushWithBadConfig
       -- TODO [#15]: admin burn proposer token and test "flush"
 
       -- TODO [#38]: Improve this when contract size is smaller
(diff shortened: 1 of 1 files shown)


rvem (approved):

rvem:
btw, have you tested that voting tests are now working fine on delphinet?

rinn7e:
> btw, have you tested that voting tests are now working fine on delphinet?

How should I test it? Can I test it locally?

rvem:
You should be able to test it locally using local-chain and providing the required environment variables (same as in CI). Probably, you'd rather comment out all test except changed to speed everything up

rinn7e:
> You should be able to test it locally using local-chain and providing the required environment variables (similar to CI). Probably, you'd rather comment out all test except changed to speed everything up

Running the test with this: `stack build --fast --test --test-arguments "--nettest-run-network -A localhost -P 8732"` take a long time but then result in this error:
```
Timeout for action call was reached. Probably, something is wrong with testing environment.
```

Running with this `stack build --fast --test --test-arguments "--nettest-run-network  -A delphi.testnet.tezos.serokell.team -P 8732"` immediately return this error:

```
          tezos-client unexpectedly failed with error code 1. Stdout:
          
          Stderr:
          Fatal error: unknown protocol version.
```

Not sure if I miss anything.

rvem:
Does your tezos-client new enough? It should be 7.4 version to be able to work with delphi protocol

test/Test/BaseDAO/Proposal.hs
```diff
@@ -266,7 +266,9 @@ flushAcceptedProposals = uncapsNettest $ do
   key1 <- createSampleProposal 1 owner1 dao
   checkTokenBalance (DAO.frozenTokenId) dao owner1 10
 
-  callFrom (AddressResolved admin) dao (Call @"Set_voting_period") 20
+  -- Use 60s for voting period, since in real network by the time we call
+  -- vote entrypoint 30s is already passed.
+  callFrom (AddressResolved admin) dao (Call @"Set_voting_period") 60
   callFrom (AddressResolved admin) dao (Call @"Set_quorum_threshold") 1
```
gromakovsky, line 272: Hm, we make many calls between `propose` and `vote`. We call `set_voting_period` and `set_quorum_threshold`, also each `checkTokenBalance` makes 2 operations. Can we reorder operations here, so that `vote` is called` immediate after `propose` (or there is at least 1 call between them)?

Note that even if we do it, these tests still seem a bit fragile to me. AFAIU there are no strict guarantees about time between blocks. Sometimes bakers can miss their slots and there can be more than 30 seconds between 2 blocks in testnet. And in mainnet it's 1 minute on average. I doubt we'll ever test contracts in mainnet this way, but I still don't like that we rely on block interval. So:
1. I think we can more or less safely assume that there can be at most 1 minute between 2 blocks. Usually it's 30 seconds, but it's not guaranteed, so let's multiply it by 2. If we reorder actions so that `vote` is called right after `propose`, I think 1 minute should be fine as voting period. If we make `n` operations between `vote` and `propose`, we should have `n + 1` minutes as voting period.
2. Waiting for 1 minute is sad when we run this test in local chain with 1 second between blocks. I think we need to add a new method that returns an approximate time between blocks. I'll create an issue about that.
3. I think the only way to reliably write this test is to put `propose` and `vote` calls in one batch. However, batching is not merged in `cleveland` yet and there are some issues with it. So it won't be available sufficiently soon :disappointed: 

So I think that in order to run this test in real network we need some changes in `cleveland`. At least we need a way to know block interval because we don't want to wait for 1 minute in local chain tests. For this reason I propose to run these tests only in emulator for now. When we add at least block interval getter to `cleveland` (preferably also proper batching), we will restore this test in real network.

AFAIU the same applies to other tests fixed in this PR as well, so let's run them only on emulator, too.

gromakovsky (changes requested):

test/Test/BaseDAO/Proposal.hs
rinn7e: That makes sense, Ivan. I've reorder the operations, and make those tests only run in the emulator for now.

gromakovsky (approved):
Okay, looks fine now. I think we don't need to test it in delphinet anymore because tests run only in emulator. I've just created #47. Can you please refer to it as well? Feel free to merge after that.

Issue #29, opened 2020-10-27
Refine/improve documentation


gromakovsky:
# Clarification and motivation

Currently we don't have enough documentation. We have:
1. Auto generated documentation referred from `README.md`. It's quite small and doesn't even describe entrypoints.
2. `docs/specification.md` which describes what contract needs to be implemented, but doesn't really explain how to use it.
3. Trivial build and usage instructions.

# Acceptance criteria

1. Auto generated documentation is expanded and contains as much as possible. For example: storage description, description of each entrypoint.
2. There are detailed usage instructions. In particular, they should explain how to instantiate parameters of the contract. The contract we developed is a base contract for various DAOs, we should clearly explain how to implement DAOs.
3. `docs/specification.md` is proofread to match our code. Well, it may be too hard to check _everything_ carefully, but at least let's check that error scenarios and corresponding errors are accurate.

gromakovsky:
Also we should check that all errors are present in `docs/specification.md`.

gromakovsky:
> There are detailed usage instructions. In particular, they should explain how to instantiate parameters of the contract. The contract we developed is a base contract for various DAOs, we should clearly explain how to implement DAOs.

Note that #41 implements a more specific DAO (GameDAO).

gromakovsky:
Part (1) is assigned to @PruStephan, he's working on it. I intended to do part (2) myself, but got sick, so I am reassigning it to @Martoon-00. I am going to add part (3), it will be for @Martoon-00 as well, he can have a fresh look.

Martoon-00:
Let me remind that no one has checked that the entrypoints errors are enlisted correctly. 

Probably it's worth to decide who does this, or maybe better, remove that and refer to our autodoc (and then also decide which other parts in the specification we would like to deduplicate - maybe keeping specification as a list of diffs between autodoc and what we want to achieve will be simpler both for us and for readers)

gromakovsky:
@Martoon-00 a few small leftovers for you :nate:
1. There is still one `#29` TODO left, please resolve it.
2. https://github.com/input-output-hk/haskell.nix/pull/943 got merged recently, we should repin `haskell.nix` to master.
3. FA2 metadata handling is currently not implemented, but people ask questions about metadata. Ideally we'd like to do https://github.com/tqtezos/baseDAO/issues/42 first, i. e. implement TZIP-16 metadata from the beginning rather than implement the current FA2 way (which will not be recommended soon).

Issue #42, opened 2020-11-06
Implement TZIP-16 according to TZIP-12


gromakovsky:
# Clarification and motivation

See https://gitlab.com/tzip/tzip/-/merge_requests/94. Now TZIP-12 (FA2) refers to TZIP-16 and has new requirements. We should update `baseDAO` contract accordingly.

# Acceptance criteria

After https://gitlab.com/tzip/tzip/-/merge_requests/94 is merged, `baseDAO` contract should comply with the newest version of FA2.

Martoon-00:
In this issue we should also make sure that `sTokenAddress` storage field can be initialized and later managed by the user - if it makes sense with regard to the final version of TZIP-16.

Martoon-00:
Also note the view for counter - it will be introduced in #56 and should appear in metadata.

Martoon-00:
How do we want to implement this?

### Put metadata in-place (as a separate deployment stage) vs in another contract

I don't consider the option of making metadata part of the contract because we are too limited.

I suppose keeping metadata in the same contract is better and not more difficult for everyone than keeping metadata separately. Probably we will have a dedicated `SetMetadata Metadata` or `UpdateMetadata (Lambda Metadata Metadata)` endpoint callable by admin, and ask the client to use it during contract origination sequence. Using `UpdateMetadata` seems more flexible in terms of possibility to apply bug fixes and in terms of fitting into operation size limit.

### How to initialize metadata?

Some fields will be provided by the client - e.g. name, homepage, we should not care about them.

Other fields are determined by us (description, version, license, authors, interfaces, errors, views). We need to decide in which format will we enlist them and how will we deliver them. Probably we could add an executable to generate all these in Yaml (or something else if they request a different format), and make CI produce an artifact with the most actual metadata.

Martoon-00:
Also

### Do we want to provide automatically generated fields manually for now, or wait for them to be derived from doc items?

AFAIU things like errors is too tiresome to provide manually, it's easier to extract them from autodoc even if that will be a temporary solution implemented in this repository. Other metadata fields are easy to provide by hands.

gromakovsky:
We already had to answer these questions in Stablecoin :)
[...]

PR #46, opened 2020-11-09
[#29] Update specification

Martoon-00:
## Description

Here I check that `specification.md` matches against our implementation, except for entrypoints behaviour and errors that should be easier to check when autodoc is implemented.

## Related issue(s)

Resolves (partially) #29

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -40,9 +40,10 @@ The smart contract is implemented in Haskell eDSL Lorentz which allows us to hav
 1. Type parameters that should be instantiated to specific types by DAO creator.
 2. Michelson constants, including lambdas, that can be passed as Haskell values.
 
-BaseDAO has 1 type parameter `proposalMetadata` that defines the structure of a proposal.
-It contains fields that are required to submit a proposal.
-It can contain any fields.
+BaseDAO has 2 type parameters `proposalMetadata` and `contractMetadata` that define the structure of a proposal and contract-wide storage respectively.
+The former contains fields that are required to submit a proposal.
+The latter keeps global information like information about accepted proposals.
+They can contain any fields that can theoretically appear in storage; the only exception is that `proposalMetadata` cannot contain unpackable data like `big_map`s, since we need to be able to compute hash of every proposal.
 Here is an example of "treasury" style DAO:
 ```haskell
 data ProposalMetadata = ProposalMetadata
@@ -51,12 +52,14 @@ data ProposalMetadata = ProposalMetadata
   , agoraPostId :: Natural
   ...
   }
+
+data ContractMetadata = ContractMetadata  -- empty
 ```
 
-Compile-time value parameters are captured by the `Config` type (parameterized by `proposalMetadata`):
+Compile-time value parameters are captured by the `Config` type (parameterized by `proposalMetadata` and `contractMetadata`):
 
 ```haskell
-data Config proposalMetadata = Config
+data Config contractMetadata proposalMetadata = Config
   { daoName :: Text
   -- ^ Name of the DAO.
   , daoDescription :: Text
@@ -75,9 +78,13 @@ data Config proposalMetadata = Config
   -- For example, if Alice freezes 100 tokens to vote and this lambda divides the value by 2,
   -- only 50 tokens will be unfrozen and other 50 tokens will be burnt.
   , decisionLambda :: Lambda
-      (Proposal proposalMetadata, Storage proposalMetadata)
-      (List Operation, Storage proposalMetadata)
+      (Proposal proposalMetadata, Storage contractMetadata proposalMetadata)
+      (List Operation, contractMetadata)
   -- ^ The decision lambda is executed based on a successful proposal.
+  -- It has access to the proposal, can modify `contractMetadata` and perform aribtrary
+  -- operations.
+  -- In case if some data needs to be gathered from the accepted proposal, we suggest
+  -- adding a `BigMap ProposalKey data` to `contractMetadata` and put the data there.
   , maxProposals :: Natural
   -- ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   , maxVotes :: Natural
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -19,6 +19,47 @@ You can use `stack build` to build `baseDAO` executable.
 
 Run `stack exec -- baseDAO --help` to see available commands.
 
+### Deploying a contract
+
+<!--
+TODO: simplify commands once we use changes in this MR:
+https://gitlab.com/morley-framework/morley/-/merge_requests/652
```
gromakovsky, line 26: Probably it won't be applicable after #41 anyway because there we are adding another contract to the registry.

Martoon-00: Yeah, fair, can remove it now.

README.md
```diff
@@ -19,6 +19,47 @@ You can use `stack build` to build `baseDAO` executable.
...
+stack exec baseDAO -- storage-BaseDAO --admin $admin > storage
+```
+
+Later the contract can be originated with `tezos-client`:
+
+```sh
+tezos-client originate contract baseDAO transferring 0 from alice running BaseDAO.tz --init "$(<storage)" --burn-cap 17.0
+```
+
+How to install `tezos-client` executable can be found at [`tezos-package` repository](https://github.com/serokell/tezos-packaging) page.
```
gromakovsky, line 51: 1. It's called `tezos-packaging`.
2. I think we should refer to official documentation as well: http://tezos.gitlab.io/introduction/howtoget.html. There are other ways that some people may prefer. And probably `tezos-packaging` will be mentioned there soon and we won't need to refer to `tezos-packaging`.

docs/specification.md
```diff
@@ -802,3 +799,75 @@ unit
 - Fails with `NOT_MIGRATION_TARGET` if the sender of this call is not the address of the new version.
 
 [FA2]: https://gitlab.com/tzip/tzip/-/blob/3a6464b1e641008b77a83807a0c102e7602c6af4/proposals/tzip-12/tzip-12.md
+
+# How to write DAOs
```
gromakovsky, line 803: Hm, I think it's not the best place for this chapter. Probably users of the contract shouldn't need to read this document, it's mostly for the customer. We also have autodoc for end users that describes entrypoints.

I would put this "How to write DAOs" chapter into "Usage" instructions in README or into a new document if we find it too big for README.

test/Test/Common.hs
```diff
@@ -77,8 +78,11 @@ originateBaseDaoWithBalance config balFunc = do
       { odFrom = nettestAddress
       , odName = "BaseDAO"
       , odBalance = toMutez 0
-      -- , odStorage = DAO.mkStorage admin bal operators
-      , odStorage = mkStorage admin mempty mempty
+      -- , odStorage = (mkStorage ! #admin ! defaults) {
+      --  sLedger = BigMap bal
+      --  sOperators = operators
+      -- }
```
gromakovsky, line 84: What's up here?

gromakovsky: Ah, okay, it's supposed to be restored once the contract is made smaller.

docs/specification.md
```diff
@@ -614,37 +612,16 @@ Parameter (in Michelson):
 )
 ```
 
-- The proposal is saved and assigned a unique numeric proposal ID.
+- The proposal is saved under `BLAKE2b` hash of proposal value.
```
gromakovsky, line 625: Note that we hash a pair where `SENDER` is included.

docs/specification.md
```diff
@@ -691,12 +668,12 @@ nat
...
 
 ```haskell
-type ProposalId = Natural
+type ProposalKey = ByteString
 
 type VoteType = Bool
 
 data VoteParam = VoteParam
-  { proposalId :: ProposalId
+  { proposalId :: ProposalKey
```
gromakovsky, line 686: `s/proposalId/proposalKey`?

docs/specification.md
Martoon-00: Oh, good catch.

docs/specification.md
```diff
@@ -94,6 +101,8 @@ data Config proposalMetadata = Config
   }
 ```
 
+***TODO:*** Contract metadata is not implemented yet, as well as the signature change of `decisionLambda`.
```
gromakovsky, line 104: Let's refer to the related issue?

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -257,41 +257,36 @@ type VotingPeriod = Natural
...
+  , sProposals = mempty
   , sProposalKeyListSortByDate = []
   }
+  where
+    votingPeriodDef = 60 * 60 * 24 * 7  -- 7 days
+    quorumThresholdDef = 4
+    -- ^ any proposals that have less that 4 votes (by default) will be rejected
+    -- regardless of upvotes
+
+emptyStorage :: Storage pm
```
gromakovsky, line 285: `emptyStorage` seems to be unused btw. Let's remove? It looks useless outside of tests because of `genesisAddress`. And even in tests it seems useless anyway because we don't have access to `genesisAddress` in Nettest.

Martoon-00: Yep, makes sense to me.

docs/specification.md
```diff
@@ -101,7 +101,7 @@ data Config contractMetadata proposalMetadata = Config
   }
 ```
 
-***TODO:*** Contract metadata is not implemented yet, as well as the signature change of `decisionLambda`.
+***TODO [#29]:*** Contract metadata is not implemented yet, as well as the signature change of `decisionLambda`.
```
gromakovsky, line 104: Shouldn't it be #42? At least the metadata part.

gromakovsky (approved):

docs/specification.md
Martoon-00: Ah hm, I meant not FA2 metadata rather the new `contractMetadata` type argument. 

Is it worth using a different name for it to avoid confusion? Though then for consistency we have to rename `proposalMetadata` as well then.

gromakovsky: Hmm, I'd call it `contractExtra` then (for example). `proposalMetadata` sounds less confusing because it's clear that it applies to a proposal (something small within the whole contract) while TZIP-16 is about metadata of a contract.

gromakovsky: I mean, since `proposalMetadata` is already used in many places and is not very confusing, I think we can keep it.

Martoon-00: Okay, I think we better do the replacement.

PR #49, opened 2020-11-12
[#29] Add usage instructions

Martoon-00:
## Description

Add some usage examples for those who are going to use our contract.

## Related issue(s)

Resolves (partially) #29

- Tests
        silently reappearing again.

- Documentation
    - Haddock


README.md
```diff
@@ -19,6 +19,52 @@ You can use `stack build` to build `baseDAO` executable.
 
 Run `stack exec -- baseDAO --help` to see available commands.
 
+### Deploying a contract
+
+You can dump the entire contract code into a `BaseDao.tz` file using the following command:
```
gromakovsky, line 24:
```suggestion
You can dump the entire contract code into a `BaseDAO.tz` file using the following command:
```

README.md
```diff
@@ -19,6 +19,52 @@ You can use `stack build` to build `baseDAO` executable.
...
+
+```sh
+stack exec baseDAO -- print -n baseDAO
+```
+
+For now this can only produce a contract with empty proposal metadata and the simplest possible configuration.
```
gromakovsky, line 30: Perhaps we should also consider the possibility to print GameDAO now that it's merged?

README.md
```diff
@@ -19,6 +19,42 @@ You can use `stack build` to build `baseDAO` executable.
...
+
+### Documentation
+
+You can obtain the actual version of the contract documentation via
+
+```sh
+stack exec baseDAO -- document -n BaseDAO
+```
+
+This command will dump the documentation to `BaseDAO.md` file.
```
gromakovsky, line 56: Above we say where one can open this documentation. I think an unprepared reader might be confused because it's not obvious what's the relation between documentation in this paragraph and aforementioned documentation. We know it's the same documentation, but an arbitrary person doesn't.

src/Lorentz/Contracts/TemplateDAO.hs
```diff
@@ -0,0 +1,103 @@
+-- SPDX-FileCopyrightText: 2019 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: Unlicense
```
gromakovsky, line 3: Licensing information doesn't look right btw.

README.md
```diff
@@ -55,6 +55,16 @@ stack exec baseDAO -- document -n BaseDAO
 
 This command will dump the documentation to `BaseDAO.md` file.
 
+## How to write your own DAOs
+
+This repository contains a [template for a DAO contract](/src/Lorentz/Contracts/TemplateDAO.hs) that can be used as an example.
+
+To create your own DAO contract:
+1. Either fork this repository or add a separate Haskell project that depends on this repository;
+2. Create a new Haskell module as a copy of the mentioned template;
```
gromakovsky, line 64: If one forks this repository, I think there is no need to copy the template, one can modify it in place. One _may_ rename this module if they want, but it's really just an aesthetical thing. Also updating `Main.hs` is not really necessary (maybe except for renaming). So I think there are 2 possible flows:
1. You fork this repo and modify code in place. Modifying in place seems more convenient than creating new modules, there are less actions required on your side. One drawback is that you have to manually sync up with upstream changes, you basically maintain a copy of another repo.
2. You create a new project with `baseDAO` directory, in that case you indeed need to make a copy of the template.

Considering what I wrote above, maybe let's make another _package_ in this repo that would depend on `baseDAO` and serve template purpose? Current template is helpful for (1), but is less useful for (2) because in (2) one still needs to manually create a new package. I think this new template package needs just an executable, config can be put there. Probably it's not a good idea to include this template into `baseDAO` library because people who use approach (2) don't want to depend on this module (it's small, but still some extra code).

gromakovsky (changes requested):

README.md
Martoon-00: Okay, I think this should be the most convenient thing for users (though potentially pretty inconvenient for us).

Then I remove the last commit and leave a reference to the new project instead.

Martoon-00: Okay, I think this should be the most convenient thing for users.

Will drop this commit and add a new one instead.

template/LICENSE
```diff
@@ -0,0 +1,20 @@
+This is free and unencumbered software released into the public domain.
```
gromakovsky, line 1: Where did you take it? Why not https://github.com/tqtezos/baseDAO/blob/master/LICENSE?

template/app/Main.hs
```diff
@@ -0,0 +1,62 @@
+-- SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io>
+-- SPDX-License-Identifier: Unlicense
```
gromakovsky, line 2: It's still incorrect, should be MIT-TQ.

template/app/Main.hs
```diff
@@ -0,0 +1,62 @@
...
+  , Opt.footerDoc usageDoc
+  ]
+  where
+    versionOption = Opt.infoOption ("MyDAO-" <> showVersion version)
+      (Opt.long "version" <> Opt.help "Show version.")
+
+repoSettings :: GitRepoSettings
+repoSettings = GitRepoSettings $ \commit ->
+  "https://github.com/tqtezos/baseDAO/tree/" <> commit
+  -- TODO: update this to refer to your repository, the link will appear in
```
gromakovsky, line 38: Maybe let's provide `TODO` for each action that the user should take? Then in "How to write your own DAOs" we can tell the user to go through all TODOs and do what's written there.

template/package.yaml
```diff
@@ -0,0 +1,111 @@
...
+  -Wno-unsafe
+  -Wno-safe
+  -Wno-missing-local-signatures
+  -Wno-monomorphism-restriction
+  -Wno-implicit-prelude
+
+dependencies:
+  - base >= 4.7 && < 5
+
+library:
```
gromakovsky, line 84: Is the library needed at all? I doubt anyone will ever use it as a library, so putting everything into executable seems simpler. AFAIU it's just one module.

template/package.yaml
```diff
@@ -0,0 +1,111 @@
...
+license:             MIT
+license-file:        LICENSE
+author:              Serokell, Tocqueville Group
+git:                 git@github.com:tqtezos/baseDAO.git
+copyright:           2020 Unlicense
+
+extra-source-files:
+- README.md
+
+description:         Please see the README on GitHub at <https://github.com/tqtezos/baseDAO/template#readme>
```
gromakovsky, line 16: Perhaps someone will put it in their own repo, so this comment will look wrong. I see three options:
1. Write something different not referring to our repo that will hold for any DAO.
2. Not include `description` at all.
3. Add a TODO indicating that the user of this template ought to do something with this line.

Something similar applies to some metadata fields above (e. g. `author`).

gromakovsky (changes requested):

template/LICENSE
Martoon-00: Don't we want people to freely use this without any restrictions? As we do for e.g. config file templates

template/app/Main.hs
Martoon-00: Hm, I can try. Probably it should look as follows:
1. `TemplateDAO.hs` have to be read carefully and updated entirely; so I'll leave a point in the `README` saying this about `TemplateDAO.hs`. 
2. In `Main.hs` I will put these `TODO` and ask to resolve them in `README`.

template/package.yaml
Martoon-00: We can do this, though two things that worry me:
1. Someone can really create an intermediate library via defining polymorphic `ProposalMetadata` and `ContractExtra`. But such cases are probably rare and for advanced users, so we don't care.
2. Again, the whole `TemplateDAO` is a point of interest for the user, while `Main.hs` contains a lot of logic that may be considered complex and non-interesting, and mixing them will increase the cognitive load for the user (at least I feel so).
Maybe we could extract most of `Main.hs` to `CLI.hs` module and leave a three-liner at `Main.hs`, so that the user could concentrate on writing his logic? Of course, in a way that can be overridden if necessary. Then I think having one module is pretty fine.

template/package.yaml
Martoon-00: I think I'll go with 3 and 1 (where possible).

template/LICENSE
gromakovsky: I had the same thought but ended up with conclusion that MIT is very permissive anyway. The difference only matters for those who want to apply a different license here AFAIU. And I am not sure if we want to permit that. On the one hand, Unlicense should decrease usage barrier which is good for us. On the other hand, if someone writes their own cool DAO, we probably want them to keep our copyright notice and make their cool DAO public and available for everybody. Though probably if someone wants to use a different license, they can create a new project without our templates. So I don't know what's better, both options seem acceptable to me now. However, even if you decided to keep Unlicense, please update `SPDX-FileCopyrightText` in some files to TQ, it still seems more correct than Serokell.

template/app/Main.hs
gromakovsky: Or you can move `repoSettings` into `TemplateDAO.hs`, so that only that file will have to be updated.

Btw, I think we should also consider the case when there is no suitable repository link to put here and make it clearer what to do in that case.

template/package.yaml
gromakovsky: >  `TemplateDAO` is a point of interest for the user

If we see it this way, probably we can put everything that should be modified by a user into `TemplateDAO.hs` :shrug:
Or we can have `TemplateDAO.hs` and `CLI.hs` which should be modified.

Right now we have a place in `Main.hs` that needs to be updated, so places-to-be-updated are mixed anyway.

template/app/Main.hs
Martoon-00: > Btw, I think we should also consider the case when there is no suitable repository link to put here and make it clearer what to do in that case.

I also thought about this, but I don't know what to do in this case myself ))

Will try to find a way to provide no git rev.

template/package.yaml
Martoon-00: This looks good, probably all we need to change in CLI is repo reference and contract name, so I'll just put them into `TemplateDAO.hs`.

template/app/Main.hs
gromakovsky: One option is to hardcode `DGitRevisionUnknown` and say that git revision is for advanced users who can tweak it themselves if they want.

Martoon-00: I was afraid that it will trigger some problemes on autodoc side, but gonna check.

Martoon-00: Ah wait, a specially added the `DGitRevisionUnknown` for case when there is no related git repository, everything should be fine :facepalm:

gromakovsky:
I recommend to rebase ASAP to make tests faster.

template/README.md
```diff
@@ -0,0 +1,11 @@
+# DAO template
+
+This package serves as a template for writing new DAO contracts.
+
+## How to write your own DAOs
+
+To create your own DAO contract:
+1. Either fork this repository, or copy the [`template`](/template) folder to your repository;
```
gromakovsky, line 8: Now I see that copying the `template` folder is not so straightforward. It's not enough to just copy this folder, one also needs `stack.yaml` file to make it work, in particular they need to point to the `baseDAO` package somehow. So probably we should provide a custom `stack.yaml` file here. Note that we can create `resolver.yaml` (or `snapshot.yaml`) file with common parts of the root `stack.yaml` and this new custom `stack.yaml`.

This custom `stack.yaml` needs to point to this repo, ideally it should point to the current commit, but it's impossible. Maybe we can have `stack.yaml` here that looks like:
```
resolver: resolver.yaml
extra-deps:
- git:
    https://github.com/bla-bla # baseDAO link
  commit:
    # TODO
```
and ask the user to provide last commit instead of `TODO`.

If this single extra-dep is the only addition to the root `stack.yaml`, probably we can only use the root one and treat this custom one as an additional artifact.

Martoon-00: > Note that we can create resolver.yaml (or snapshot.yaml) file with common parts of the root stack.yaml and this new custom stack.yaml.

If I do so, where should I put it? Looks like I shouldn't put the snapshot into the root of `baseDAO`, because then again the user won't be able to just copy-paste the folder and will get a confusing error about `../resolver.yaml` not existing.

> If this single extra-dep is the only addition to the root stack.yaml, probably we can only use the root one and treat this custom one as an additional artifact.

Do you mean I should try to put an extra dep refering to `baseDAO` in `stack.yaml` of `baseDAO`?

gromakovsky: No, I think we can have `resolver.yaml` and `stack.yaml` in `template/`. `template/resolver.yaml` can be used in both `stack.yaml` files. The root one doesn't need anything else. The one in `template/` also needs `baseDAO` in its `extra-deps`.

AFAIU `extra-deps` listed in `stack.yaml` will be concatenated with `extra-deps` from `resolver.yaml`.

gromakovsky:
I'm not sure in which state this MR is, pls ping me once it's ready.

Martoon-00:
> I'm not sure in which state this MR is, pls ping me once it's ready.

I'm not sure what's going on, but my recent commits are not here and my last commit is also lost :thinking: :thinking: :thinking:

template/README.md
Martoon-00: Okay, I implemented this suggestion.

The only problem with it which I see is that if we leave commit sha uninitialized in `template/stack.yaml`'s `extra-deps`, then building the template project will be tiresome for developers who introduce breaking changes, and for CI.

gromakovsky: Aha, my idea was to not use `template/stack.yaml` to build `template` in **this** repo. I. e. keep `template` in `packages` in root `stack.yaml` and use the root `stack.yaml` for building. This way when we build **this** repo we will take `baseDAO` from **this** repo. I think it should work fine as long as `baseDAO` extra-dep is the only difference between two of them. I mean, if it builds for us with the root `stack.yaml`, it should also build for end user with `template/stack.yaml` as long as they provide the right revision.

But apparently it's complicated because `stack` uses the deepest `stack.yaml` it sees. One option is to make user's life a bit harder by renaming `template/stack.yaml` in this repo to `template/_stack.yaml` and asking the user to rename it. Or by asking them to pass `--stack-yaml` or `--resolver`. But probably user's convenience is more important and we should handle this problem on our side. It's possible to point `stack` to the desired `stack.yaml`, we just need to do it in all places where we `stack build`:
1. Make CI build both packages using the root `stack.yaml`. I don't know precisely how to do because we are using `haskell.nix` there, but probably you'll figure it out or will ask our nix experts.
2. Developers are supposed to use `Makefile`. There we currently build only `baseDAO` and it should work fine. We should add a way to build `template` and this command should point to the right `stack.yaml`.

Does it sound alright?

Martoon-00: > Aha, my idea was to not use template/stack.yaml to build template in this repo. I. e. keep template in packages in root stack.yaml and use the root stack.yaml for building.

Oh okay. I first tried this and got some weird error, but probably this error was caused by a different reason. If I add `template` to `/stack.yaml`, then everything works for me with just `stack build`.

> But apparently it's complicated because stack uses the deepest stack.yaml it sees.

Does it work for you so, or you are referring to my experience? Looks like for me `stack` always thinks that `/stack.yaml` is the one to work with, unless I'm within `template` folder.

gromakovsky: > Does it work for you so, or you are referring to my experience? 

I thought that it works this way, but I didn't test it. Also I thought that it worked this way for you.

> Looks like for me stack always thinks that /stack.yaml is the one to work with, unless I'm within template folder.

Okay, cool then, can you add `template` back to the root `stack.yaml`, so that I could test that version? Also it'd be nice to rebase on master (there is a conflict) and update `Makefile`.

Martoon-00: Yep, should be fixed now.

gromakovsky:
1. Apparently CI still doesn't like something. It can't find `template/snapshot.yaml` for some reason. Probably it's a problem in `haskell.nix`. If you have no idea why it happens, check their issue tracker, maybe it's unsupported :shrug: If you don't find anything useful, feel free to ask our `haskell.nix` experts.
2. Since you dared to add .cabal file (which is good), please also add a CI check that it's up-to-date, I think it's the only way to keep it up-to-date. Fortunately such a check is easy to find in `morley` repos.

template/README.md
gromakovsky: Okay, `make` seems to work fine now. However, `stack build` doesn't work from the `template` folder, which is expected and fine, but I think it's worth mentioning somewhere, maybe in `CONTRIBUTING.md`.

Martoon-00:
>     1. Apparently CI still doesn't like something. It can't find `template/snapshot.yaml` for some reason. Probably it's a problem in `haskell.nix`. If you have no idea why it happens, check their issue tracker, maybe it's unsupported shrug If you don't find anything useful, feel free to ask our `haskell.nix` experts.

I once tried to add a snapshot in another project and I got an impression that this quite can be handled by nix.

I tried to find a trivial way include the snapshot and failed. Probably even if I manage to come up with something, that won't be an ideal solution, so created [OPS-1122](https://issues.serokell.io/issue/OPS-1122).

gromakovsky:
I want to check what else needs to be done to make our documentation good enough, will do one day. Right now I have two items in my head:
1. Fix `GaseDAO` typo in README.
2. Provide a link to `stack`, we shouldn't assume that users are familiar with Haskell world.

Can you please do both in this PR?

gromakovsky:
> Provide a link to stack, 

Probably we should have some "Prerequisites" section.

Martoon-00:
> > Provide a link to stack,
> 
> Probably we should have some "Prerequisites" section.

What will I write there? (assuming that creating a section for one sentence does not make sense).

Like, describe a different ways to install `stack`? Mention IDEs?

gromakovsky:
> BaseDAO is a basic generic smart contracts for DAO on Tezos.

Found one more typo in README (s/contracts/contract). However, probably we should update this sentence to say that BaseDAO is a framework to develop DAO contracts on Tezos. And then we can expand it saying that this repo provides:
1. Generic DAO contract (template) that can be used as basis for many DAOs.
2. Two instantiations of this template: simplest implementation with no custom logic (`BaseDAO` contract) and a simple example with custom logic called `GameDAO`.
3. A template _package_ that one can copy and use to create their own DAO.

> What will I write there? (assuming that creating a section for one sentence does not make sense).

Just one line that you already wrote is enough. However, it applies not only to "Build Instructions", but also to usage instructions and "Writing your own DAOs". So I thought that we would say something like "In order to build and use existing DAO contracts or write your own DAO you will need Stack, see how to obtain it" before proceeding to build and usage instructions.
We can also say that in order to develop your DAO based on BaseDAO one needs to have basic knowledge of Haskell and Lorentz (and we can refer to Lorentz docs).

Martoon-00:
> However, probably we should update this sentence to say that BaseDAO is a framework to develop DAO contracts on Tezos.

Okay, done. I updated the second point to mention `TrivialDAO` instead of `BaseDAO`, hope that's what you meant.

> However, it applies not only to "Build Instructions", but also to usage instructions and "Writing your own DAOs".

Aha, makes sense to me.

README.md
```diff
@@ -12,12 +20,16 @@ Documentation of the actually implemented smart contract is generated automatica
 * [BaseDAO](https://github.com/tqtezos/baseDAO/blob/autodoc/master/BaseDAO.md)
```
gromakovsky, line 20: Probably it should be changed to `TrivialDAO` as well?

gromakovsky (approved):
Okay, let's merge it, but not close #29 yet, I want to make another pass over our documentation.

README.md
Martoon-00: Hm, true

Martoon-00:
I added a commit where I refer back to `BaseDAO`, not `TrivialDAO`, because that is not yet present in `autodoc/master`. Fixing this in #60.

gromakovsky:
Ok, so when are we merging this one?

PR #50, opened 2020-11-14
[#38] Instantiate contract immediately in tests

Martoon-00:
## Description

Problem: due to hitting operation size limit we initialized empty
contract and then added there desired balance distribution and
operators, but that looks like extra work.

Solution: uncomment the old code that creates initial storage with all
the necessary balances and operators, remove contract calls that filled
them.

## Related issue(s)

Related to #38.

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -121,6 +121,7 @@ tests:
     dependencies:
     - baseDAO
     - cleveland
+    - containers
     - fmt
     - HUnit
     - lorentz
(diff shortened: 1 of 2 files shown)


rinn7e (approved):
Seems good to me!

gromakovsky (approved):

PR #51, opened 2020-11-16
[#29] Build and upload GameDAO documentation on CI

gromakovsky:
## Description

Problem: we recently added `GameDAO` contract with its own
documentation.
However, this documentation is not generated and checked by our CI,
it's hard to access it.

Solution: generate, check and upload it as part of CI, the same way
as we do for `BaseDAO`.

## Related issue(s)

#29

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -63,10 +63,11 @@ steps:
   - label: contract doc development
     if: build.branch != "autodoc/master" && build.branch != "master"
     commands:
-      - nix-build ci.nix -A contracts-doc --arg release false
-      - ln -s ./result/BaseDAO.md BaseDAO.md
+    - nix-build ci.nix -A contracts-doc --arg release false
+    - ln -s ./result/*.md .
     artifact_paths:
       - BaseDAO.md
+      - GameDAO.md
 
   - label: contract doc release
     if: build.branch == "master"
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -75,8 +77,10 @@ steps:
         --argstr commitSha "$(git rev-parse HEAD)"
         --argstr commitDate "$(git log HEAD -1 --format=%cd)"
     - ln -s ./result/BaseDAO.md BaseDAO.md
+    - ln -s ./result/GameDAO.md GameDAO.md
```
Martoon-00, line 80: Btw, looks like it is not too hard to do [a batch symlinks creation](https://superuser.com/a/633610/756653), maybe use it?

gromakovsky: Let's see what CI will tell us :zhdun:

.buildkite/pipeline.yml
```diff
@@ -63,20 +63,22 @@ steps:
...
     if: build.branch == "master"
     commands:
     - nix-build ci.nix -A contracts-doc --arg release true
         --argstr commitSha "$(git rev-parse HEAD)"
         --argstr commitDate "$(git log HEAD -1 --format=%cd)"
-    - ln -s ./result/BaseDAO.md BaseDAO.md
+    - ln -s ./result/*.md .
     artifact_paths:
       - BaseDAO.md
+      - GameDAO.md
```
Martoon-00, line 81: Why people move files to `.` and then refer to specific files, instead of referring to the folder :unamused:

If not this, we could decrease the number of necessary updates on contract addition to 1-2.

Martoon-00 (approved):

.buildkite/pipeline.yml
gromakovsky: Maybe because this way it's easier to download these artifacts from CI :shrug:

Martoon-00: Fair, probably that's the reason

PR #52, opened 2020-11-16
[#29] Improve decision lambda

Martoon-00:
## Description

Here I add `contractExtra` type parameter to storage and config, and update the decision lambda respectively.

## Related issue(s)

Resolves (partially) #29

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -10,7 +10,7 @@ import Universum
 
 import qualified Data.Map as Map
 import Data.Version (showVersion)
-import Lorentz (DGitRevision, GitRepoSettings(..), mkDGitRevision)
+import Lorentz (DGitRevision, Default(..), GitRepoSettings(..), mkDGitRevision)
 import Lorentz.ContractRegistry
 import Main.Utf8 (withUtf8)
 import Morley.CLI (addressOption)
@@ -49,22 +49,21 @@ usageDoc = Just $ mconcat
 contracts :: ContractRegistry
 contracts = ContractRegistry $ Map.fromList
   [ "BaseDAO" ?:: ContractInfo
-    { ciContract = DAO.baseDaoContract DAO.defaultConfig
+    { ciContract = DAO.baseDaoContract @() @() DAO.defaultConfig
     , ciIsDocumented = True
-    , ciStorageParser = Just $ baseDaoStorageParser @()
+    , ciStorageParser = Just baseDaoStorageParserDef
     , ciStorageNotes = Nothing
     }
   , "GameDAO" ?:: ContractInfo
     { ciContract = GameDAO.gameDaoContract
     , ciIsDocumented = True
-    , ciStorageParser = Just $
-        baseDaoStorageParser @GameDAO.GameDaoProposalMetadata
+    , ciStorageParser = Just baseDaoStorageParserDef
     , ciStorageNotes = Nothing
     }
   ]
 
-baseDaoStorageParser :: forall m. Opt.Parser (DAO.Storage m)
-baseDaoStorageParser = do
+baseDaoStorageParser :: forall pm ce. Opt.Parser ce -> Opt.Parser (DAO.Storage ce pm)
+baseDaoStorageParser extraParser = do
   adminAddress <-
     addressOption Nothing (#name .! "admin")
     (#help .! "Administrator of the DAO contract")
@@ -74,10 +73,15 @@ baseDaoStorageParser = do
   quorumThreshold <-
     mkCLOptionParser (Just 4) (#name .! "quorum-threshold")
     (#help .! "Total number of votes necessary for successful proposal")
+  extra <- extraParser
   pure $ DAO.mkStorage
     (#admin .! adminAddress)
     (#votingPeriod .? Just votingPeriod)
     (#quorumThreshold .? Just quorumThreshold)
+    (#extra .! extra)
+
+baseDaoStorageParserDef :: Default ce => Opt.Parser (DAO.Storage ce pm)
+baseDaoStorageParserDef = baseDaoStorageParser (pure def)
 
 main :: IO ()
 main = withUtf8 $ do
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -22,6 +22,7 @@ import Util.Named
 
 import qualified Lorentz.Contracts.BaseDAO as DAO
 import qualified Lorentz.Contracts.GameDAO as GameDAO
+import qualified Lorentz.Contracts.TrivialDAO as TrivialDAO
 
 programInfo :: DGitRevision -> Opt.ParserInfo CmdLnArgs
 programInfo gitRev = Opt.info (Opt.helper <*> versionOption <*> argParser contracts gitRev) $
@@ -49,7 +50,7 @@ usageDoc = Just $ mconcat
 contracts :: ContractRegistry
 contracts = ContractRegistry $ Map.fromList
   [ "BaseDAO" ?:: ContractInfo
-    { ciContract = DAO.baseDaoContract @() @() DAO.defaultConfig
+    { ciContract = TrivialDAO.trivialDaoContract
     , ciIsDocumented = True
     , ciStorageParser = Just baseDaoStorageParserDef
     , ciStorageNotes = Nothing
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -40,7 +40,7 @@ The smart contract is implemented in Haskell eDSL Lorentz which allows us to hav
 1. Type parameters that should be instantiated to specific types by DAO creator.
 2. Michelson constants, including lambdas, that can be passed as Haskell values.
 
-BaseDAO has 2 type parameters `proposalMetadata` and `contractMetadata` that define the structure of a proposal and contract-wide storage respectively.
+BaseDAO has 2 type parameters `proposalMetadata` and `contractExtra` that define the structure of a proposal and contract-wide storage respectively.
 The former contains fields that are required to submit a proposal.
 The latter keeps global information like information about accepted proposals.
 They can contain any fields that can theoretically appear in storage; the only exception is that `proposalMetadata` cannot contain unpackable data like `big_map`s, since we need to be able to compute hash of every proposal.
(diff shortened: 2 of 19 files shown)


gromakovsky:
> Resolves (partially) #38

How does it help with optimizations? It's not clear to me :confused:

Martoon-00:
> > Resolves (partially) #38
> 
> How does it help with optimizations? It's not clear to me confused

:fuuu:

Too many PRs everywhere

Martoon-00:
I renamed the commits to include the proper issue id.

gromakovsky:
> see `7a43ae54`'s description

AFAIU this commit is unstable btw, so it should be updated.

app/Main.hs
```diff
@@ -49,22 +49,21 @@ usageDoc = Just $ mconcat
 contracts :: ContractRegistry
 contracts = ContractRegistry $ Map.fromList
   [ "BaseDAO" ?:: ContractInfo
-    { ciContract = DAO.baseDaoContract DAO.defaultConfig
+    { ciContract = DAO.baseDaoContract @() @() DAO.defaultConfig
```
gromakovsky, line 52: In another PR I wanted to suggest hiding all these type parameters and config inside the library and exporting the contract with everything applied. We can call it `simplestDAO` for example. Or `trivialDAO`. So that `baseDaoContract` is a generic thing and there is another thing with a different name that is specifically the simplest version of baseDAO.

docs/specification.md
```diff
@@ -56,10 +56,10 @@ data ProposalMetadata = ProposalMetadata
 data ContractMetadata = ContractMetadata  -- empty
```
gromakovsky, line 56: Let's rename to `ContractExtra`?

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -167,7 +167,7 @@ instance TypeHasDoc MigrationStatus where
     "Migration status of the contract"
 
 -- | Storage of the FA2 contract
```
gromakovsky, line 169: While we are at it, I think we should at least change "the FA2 contract" to "a DAO contract" (because baseDAO is a group of contracts and this storage is used in all of them). Also I think `contractExtra` parameter should be described, but in proper world `proposalMetadata` should have been described already and it's not described which is not your fault. So feel free to keep this short description if you don't want to expand it, but let's change this "the FA2 contract" thing at least (otherwise nobody will ever do it).

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -18,12 +18,23 @@ module Lorentz.Contracts.GameDAO
...
 
-import Universum hiding ((>>), drop, swap)
 import Lorentz
+
 import qualified Lorentz.Contracts.BaseDAO as DAO
 import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Universum hiding (drop, swap, (>>))
 import Util.Markdown
 
+data GameDaoContractExtra = GameDaoContractExtra
```
gromakovsky, line 28: Is it better than `()`?

src/Lorentz/Contracts/BaseDAO/Types.hs
Martoon-00: >While we are at it, I think we should at least change "the FA2 contract" to "a DAO contract"

:peka:

src/Lorentz/Contracts/GameDAO.hs
Martoon-00: I wanted to add something interesting there but forgot. Gonna add now.

Martoon-00:
> > see `7a43ae54`'s description
> 
> AFAIU this commit is unstable btw, so it should be updated.

Hm, I can see it in `origin/master`

docs/specification.md
Martoon-00: :woman_facepalming:

app/Main.hs
Martoon-00: Good, I'd like to try it. Can I apply it right in this PR?

Martoon-00: Sad that this can't be easily applied to our tests because there we use different origination methods / different configs all the time.

gromakovsky:
Ak, ok, I was looking at something wrong apparently.

src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -26,14 +26,16 @@ import Universum hiding (drop, swap, (>>))
 import Util.Markdown
 
 data GameDaoContractExtra = GameDaoContractExtra
+  { ceAcceptedProposalsNum :: Natural
```
gromakovsky, line 29: Haddock?

app/Main.hs
gromakovsky: > Can I apply it right in this PR?

I don't mind. This PR increases number of things necessary to instantiate a specific contract, so hiding this things is related to the essence of this PR.

> Sad that this can't be easily applied to our tests because there we use different origination methods / different configs all the time.


It shouldn't be a problem if we still export the generic thing (`baseDaoContract`).

src/Lorentz/Contracts/GameDAO.hs
Martoon-00: ~What if I say I prefer consistency over haddock~ Meh, okay.

src/Lorentz/Contracts/BaseDAO.hs
```diff
@@ -53,6 +54,9 @@ baseDaoContract config@Config{..} = defaultContract $ contractName cDaoName $ do
       , #cToken_address /-> tokenAddress
       )
 
+simplestDaoContract :: Config () () -> Contract (BaseDAO.Parameter ()) (BaseDAO.Storage () ())
+simplestDaoContract = baseDaoContract
```
gromakovsky, line 58: Hm, I imagined that `defaultConfig` will be bound here, not only type parameters. This `defaultConfig` is what makes this DAO actually simplest. In another config someone may provide a more complicated proposal check or decision lambda thus making DAO not that simple.

Martoon-00: Ah okay, it makes sense to me, I thought you intentionally compared `trivialDAO` with  `baseDaoContract` because you wanted me to specialize the contract.

Martoon-00: I removed the commit with the previous attempt and pushed another one on top.

gromakovsky: We have `ciContract = GameDAO.gameDaoContract`, I imagined we would do the same for the trivial DAO, i. e. export a ready-to-use contract that doesn't need any function applications or type applications. We can even create `Lorentz.Contracts.TrivialDAO` and put this `trivialConfig` there like we do for `GameDAO`.

gromakovsky (approved):
I still have one comment about stuff exported from the library, but it's really just a matter of taste and some bikeshedding which shouldn't block this PR.

test/Test/BaseDAO/Token/FA2.hs
```diff
@@ -378,7 +378,7 @@ adminTransferFrozenScenario = uncapsNettest $ do
 
 transferAfterMigrationScenario :: (Monad m) => NettestImpl m -> m ()
 transferAfterMigrationScenario = uncapsNettest $ do
-  ((owner1, op1), (owner2, newAddress1), dao, admin) <- originateBaseDao @()
+  ((owner1, op1), (owner2, newAddress1), dao, admin) <- originateBaseDao @() ()
```
gromakovsky, line 381: This wall of changes where we just add `()` argument passing makes me a bit sad. Maybe let's define `originateTrivialDao` where we can hardcode this `()`? Sounds like a natural next step since we added a notion of trivial DAO.

It won't eliminate the wall of changes (we still need to `s/originateBaseDao/originateTrivialDao` in 100500 places), but should at least make these tests more resistant to possible future changes.

Martoon-00: What about `originateBaseDaoWithConfig` and `originateBaseDaoWithBalance`? Add their trivial versions as well?

gromakovsky: Does trivial version of `originateBaseDaoWithConfig` make any sense? "Trivial" means that config is fixed, "WithConfig" means it can be supplied externally :confused: 

Regarding `originateBaseDaoWithBalance` it's not obvious in how many places its trivial version will be useful, but if there are many such places (like 3+), probably it makes sense to add such a helper.

Martoon-00: > Does trivial version of originateBaseDaoWithConfig make any sense? "Trivial" means that config is fixed, "WithConfig" means it can be supplied externally confused

Well, it could have a different name (`s/trivial/simple`), but such a function still makes sense. In fact, we call `originateBaseDaoWithConfig` with empty contract extra approximately as often as `originateBaseDao`.

> Regarding originateBaseDaoWithBalance it's not obvious in how many places its trivial version will be useful, but if there are many such places (like 3+), probably it makes sense to add such a helper.

There are exactly 3 such places :smile:

gromakovsky: > Well, it could have a different name (s/trivial/simple), but such a function still makes sense. In fact, we call originateBaseDaoWithConfig with empty contract extra approximately as often as originateBaseDao.

I think "trivial" means 3 things:
1. Empty contract extra.
2. Empty proposal metadata.
3. Trivial config, no special logic.

And usually first two things (contract extra and metadata) are deduced from the config. But not always, there can be a polymorphic non-trivial config. If there are many places where we pass a non-trivial polymorphic config and some constant type parameter(s), we can have a function for that as well.

I see we have a lot of `originateBaseDaoWithConfig @Integer config` in `Test.BaseDAO.Proposal`. Probably it makes sense to have a function defined as `originateBaseDaoWithConfig @Integer config` in that module.

Martoon-00:
I dropped the commit about making `decisionLambda` safer, added another one instead where the lambda is simplified.

test/Test/BaseDAO/Token/FA2.hs
gromakovsky: AFAIU this discussion is not resolved yet.

Martoon-00: Yeah, need some time to finish it

Martoon-00:
Okay, finished. Plz reapprove

gromakovsky (approved):
It's somewhat hard to review changes affecting so many lines, but apparently everything is fine :)

PR #58, opened 2020-11-24
[#38] Optimize `flush` slightly

### LINKED ISSUE #38: Optimize the contract

Martoon-00:
## Description

Extract `unfreezeProposerTokens` since it is used 3 times and costs ~1300 bytes in overall.

**TODO:** this depends on an [MR in Morley](https://gitlab.com/morley-framework/morley/-/merge_requests/679), I'll measure the contract size once everything works.

## Related issue(s)

Resolves (partially) #38

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/Lorentz/Contracts/BaseDAO.hs
+++ b/src/Lorentz/Contracts/BaseDAO.hs
@@ -50,7 +50,7 @@ baseDaoContract config@Config{..} = defaultContract $ contractName cDaoName $ do
     doc $ DDescription priorChecksDoc
     ensureZeroTransfer
 
-  pushFuncContext @ce @pm $ do
+  pushFuncContext config $ do
     unpair
     finalizeParamCallingDoc $ entryCase @(BaseDAO.Parameter pm) (Proxy @PlainEntrypointsKind)
       ( #cCall_FA2 /-> fa2Handler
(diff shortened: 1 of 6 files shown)


rinn7e (approved):
Seems good to me.

gromakovsky:
Can it be finished now? I'm not going to review it, unless you ask me to.

Martoon-00:
By the way, op size after this optimization is now `12840`, right before these changes (with recent vote permits stuff `12388`). Somehow the benefit is less than I expected.

PR #59, opened 2020-11-27
[#56] Permits in vote

### LINKED ISSUE #56: Make it possible to vote using Permit-like interface

See [TZIP-17](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-17/tzip-17.md) for some context. That interface can only be applied to entrypoints where the expected sender is known. E. g. for administrative entrypoints we know who should be the sender (since we store the administrator in the storage). But for `vote` we don't know the expected sender, so we can't use TZIP-17's Permit directly.

However, we want people to be able to vote by making signatures off-chain and asking someone else to make a transaction on their behalf. There are (at least) 2 possible options:
1. Extend the existing `vote` entrypoint by adding `option (pair signature key)` to `VoteParam` (maybe something else should be added there). If it's provided, we use this `key` instead of `SENDER` and check the signature.
2. Add a new entrypoint (`permitVote`) where each `VoteParam` has an additional `pair signature key` field (not optional).

We can implement either of these two. Or something else if you have better ideas. Some criteria to choose the best option:
1. We care about gas costs and contract size, so we should pick the option leading to a smaller contract. Probably the first one should be cheaper because there are less entrypoints. However, the second one can be cheap as well, as long as we extract common part(s) of `vote` and `permitVote` and put them on stack to be reused.
2. We would like to extend TZIP-17 with a standard approach to enable Permit for such entrypoints. It should be possible to recognize "inlined permit". In the second case it can be recognized by `permit` prefix. In the first case probably it can be recognized by a certain parameter type pattern (shape).

# Acceptance criteria

It is possible to vote from a `tz*` address without making a transaction from this `tz*` address.

Martoon-00:
## Description

Add permit to `vote` entrypoint.

This depends on [changes in Morley](https://gitlab.com/morley-framework/morley/-/merge_requests/683).

Now in Michelson `vote` entrypoint argument looks like

```
(list %vote (pair :permit_protected (pair 
                                          (bytes %proposal_key)
                                          (pair 
                                                (bool %vote_type)
                                                (nat %vote_amount)))
                                    (option %permit (pair 
                                                          (key %key)
                                                          (signature %signature))))))))
```

## Related issue(s)

Resolves #56

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/baseDAO.cabal
+++ b/baseDAO.cabal
@@ -4,7 +4,7 @@ cabal-version: 2.2
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: ac81ebe2db072ca09cf1331eb45df29d8279881c629d3d5ec9c171578f19a0b8
+-- hash: 42a0fb564d1e9e5dfbc142ce0ee4e038a6b84214b4b5b2079e263206f27c45dc
 
 name:           baseDAO
 version:        0.1.0.0
@@ -28,6 +28,7 @@ library
       Lorentz.Contracts.BaseDAO
       Lorentz.Contracts.BaseDAO.Doc
       Lorentz.Contracts.BaseDAO.Management
+      Lorentz.Contracts.BaseDAO.Permit
       Lorentz.Contracts.BaseDAO.Proposal
       Lorentz.Contracts.BaseDAO.Token
       Lorentz.Contracts.BaseDAO.Token.FA2
--- a/baseDAO.cabal
+++ b/baseDAO.cabal
@@ -3,6 +3,8 @@ cabal-version: 2.2
 -- This file has been generated from package.yaml by hpack version 0.33.0.
 --
 -- see: https://github.com/sol/hpack
+--
+-- hash: ac81ebe2db072ca09cf1331eb45df29d8279881c629d3d5ec9c171578f19a0b8
 
 name:           baseDAO
 version:        0.1.0.0
@@ -43,7 +45,6 @@ library
   build-depends:
       base >=4.7 && <5
     , containers
-    , data-default
     , lorentz
     , morley
     , morley-ledgers
@@ -66,7 +67,6 @@ executable baseDAO
     , baseDAO
     , lorentz
     , morley
-    , optparse-applicative
     , universum
   default-language: Haskell2010
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -231,6 +231,7 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 | `FORBIDDEN_XTZ`                | Throws when some XTZ was received as part of the contract call |
 | `PROPOSER_NOT_EXIST_IN_LEDGER` | Expect a proposer address to exist in Ledger but it is not found         |
 | `PROPOSAL_NOT_UNIQUE`          | Trying to propose a proposal that is already existed in the Storage.         |
+| `MISSIGNED`                    | Parameter signature does not match the expected one - for permits.   |
 
 
 # Entrypoints
@@ -688,32 +689,48 @@ data VoteParam = VoteParam
   , voteAmount :: Natural
   }
 
+data Permit = Permit
+  { key :: PublicKey
+  , signature :: Signature
+  }
+
 data Parameter proposalMetadata
-  = Vote [VoteParam]
+  = Vote [(VoteParam, Maybe Permit)]
 ```
 
 Parameter (in Michelson):
 
 ```
-(list
-  (pair %vote
-    (nat %proposal_key)
+(list %vote
+  (pair :permit_protected
     (pair
-      (bool %vote_type)
-      (nat %vote_amount)
+      (nat %proposal_key)
+      (pair
+        (bool %vote_type)
+        (nat %vote_amount)
+      )
+    )
+    (option %permit
+      (pair
+        (key %key)
+        (signature %signature)
+      )
     )
   )
 )
 ```
 
-- Sender MUST have unfrozen tokens equal to `voteAmount` or more (1 unfrozen token is needed for 1 vote).
-- Fails with `VOTING_INSUFFICIENT_BALANCE` if the unfrozen token balance of the SENDER
-  is less than specified `vVoteAmount` .
+- This implements permits mechanism similar to the one in TZIP-17 but injected directly to the entrypoint.
+- Vote author is identified by permit information, or if it is absent - `sender` is taken.
+- Author MUST have unfrozen tokens equal to `voteAmount` or more (1 unfrozen token is needed for 1 vote).
+- Fails with `VOTING_INSUFFICIENT_BALANCE` if the unfrozen token balance of the author
+  is less than specified `voteAmount` .
 - Fails with `PROPOSAL_NOT_EXIST` if the proposal key is not associated with any ongoing proposals.
 - Fails with `VOTING_PERIOD_OVER` if the proposal associated with the proposal key is already ended.
 - Fails with `MAX_VOTES_REACHED` if the amount of votes of the associated proposal is already at the max value set by the configuration.
-- The sender's balance in frozen tokens is increased by `voteAmount` and in unfrozen tokens is decreased by `voteAmount`.
-- The entrypoint accepts a list of vote params. As a result, the sender can `vote` on multiple proposals (or the same proposal multiple time) in one entrypoint call.
+- Fails with `MISSIGNED` if permit is incorrect with respect to the provided vote parameter and contract state.
+- The author's balance in frozen tokens is increased by `voteAmount` and in unfrozen tokens is decreased by `voteAmount`.
+- The entrypoint accepts a list of vote params. As a result, it is possible to `vote` on multiple proposals (or the same proposal multiple time) in one entrypoint call.
 
 ### **token_address**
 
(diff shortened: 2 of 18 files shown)


Martoon-00:
I want to squash commits at the end of the day unless someone objects or starts reviewing today.

Martoon-00:
In case no one reviewed this yet, I'd squash commits and rebase on the current master,

gromakovsky:
@Martoon-00 please squash, I want to review tomorrow.

baseDAO.cabal
```diff
@@ -3,6 +3,8 @@ cabal-version: 2.2
 -- This file has been generated from package.yaml by hpack version 0.33.0.
 --
 -- see: https://github.com/sol/hpack
+--
+-- hash: ac81ebe2db072ca09cf1331eb45df29d8279881c629d3d5ec9c171578f19a0b8
```
gromakovsky: Meh, it's sad. In case of this repo conflicts are more likely here, I think we don't want this hash.

Okay, let's keep these hashes within this PR, I'll create a new issue about porting our approach to `.cabal` files maintenance from `morley`.

gromakovsky: #65

src/Lorentz/Contracts/BaseDAO.hs
```diff
@@ -67,7 +66,7 @@ baseDaoContract config@Config{..} = defaultContract $ contractName cDaoName $ do
       )
 
 fa2Handler
-  :: (IsoValue ce, IsoValue pm, HasFuncContext s (BaseDAO.Storage ce pm))
+  :: (IsoValue ce, KnownValue pm, HasFuncContext s (BaseDAO.Storage ce pm))
```
gromakovsky, line 73: Why do we have to change `IsoValue` to `KnownValue`?

docs/specification.md
```diff
@@ -688,32 +689,48 @@ data VoteParam = VoteParam
...
 )
 ```
 
-- Sender MUST have unfrozen tokens equal to `voteAmount` or more (1 unfrozen token is needed for 1 vote).
-- Fails with `VOTING_INSUFFICIENT_BALANCE` if the unfrozen token balance of the SENDER
+- This implements permits mechanism similar to the one in TZIP-17 but injected directly to the entrypoint.
+- Vote author is identified by permit information, or if it is absent - `sender` is taken.
+- Author MUST have unfrozen tokens equal to `voteAmount` or more (1 unfrozen token is needed for 1 vote).
+- Fails with `VOTING_INSUFFICIENT_BALANCE` if the unfrozen token balance of the author
   is less than specified `vVoteAmount` .
```
gromakovsky, line 727: `vVoteAmount` looks like a leftover, should be `voteAmount`.

docs/specification.md
```diff
@@ -688,32 +689,48 @@ data VoteParam = VoteParam
...
 - Fails with `PROPOSAL_NOT_EXIST` if the proposal key is not associated with any ongoing proposals.
 - Fails with `VOTING_PERIOD_OVER` if the proposal associated with the proposal key is already ended.
 - Fails with `MAX_VOTES_REACHED` if the amount of votes of the associated proposal is already at the max value set by the configuration.
-- The sender's balance in frozen tokens is increased by `voteAmount` and in unfrozen tokens is decreased by `voteAmount`.
-- The entrypoint accepts a list of vote params. As a result, the sender can `vote` on multiple proposals (or the same proposal multiple time) in one entrypoint call.
+- Fails with `MISSIGNED` if permit is incorrect with respect to the provided vote parameter and contract state.
+- The author's balance in frozen tokens is increased by `voteAmount` and in unfrozen tokens is decreased by `voteAmount`.
+- The entrypoint accepts a list of vote params. As a result, the it is possible to `vote` on multiple proposals (or the same proposal multiple time) in one entrypoint call.
```
gromakovsky, line 733: > the it is possible

Redundant `the`?

docs/specification.md
```diff
@@ -688,32 +689,48 @@ data VoteParam = VoteParam
...
 
 ```
 (list
-  (pair %vote
-    (nat %proposal_key)
-    (pair
-      (bool %vote_type)
-      (nat %vote_amount)
+  (pair :permit_protected
+    (pair %vote
```
gromakovsky, line 706: This `%vote` annotation looks quite weird to me, shouldn't it apply to the topmost `list`?

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -379,6 +397,131 @@ type HasFuncContext s store =
...
+
+-- | When singing something, you usually want to use this wrapper over your data.
+-- It ensures that replay attack is not possible.
+data DataToSign d = DataToSign
+  { dsChainId :: ChainId
+  , dsContract :: Address
+  , dsNonce :: Nonce
+  , dsData :: d
+  } deriving stock (Generic)
+    deriving anyclass (IsoValue, HasAnnotation)
```
gromakovsky, line 420: Why do you use `anyclass` deriving here, but `baseDaoAnnOptions` for `Permit` below?

test/Test/BaseDAO/Proposal.hs
```diff
@@ -7,13 +7,14 @@ module Test.BaseDAO.Proposal
 
 import Universum hiding (compare, drop, (>>))
 
-import Lorentz
+import Lorentz hiding (map)
```
gromakovsky, line 10: An alternative solution is to use `fmap` :sweat_smile:

gromakovsky (changes requested):

gromakovsky:
> All errors belong to some entrypoint:                                        FAIL

gromakovsky:
Apart from the implemented approach another one was proposed: have a new entrypoint like `votePermitted`. There would be some duplication, but it probably can be solved by lambdas. Did you consider that approach?

With the currently implemetned approach I suppose we can extend TZIP-17 by saying that if an entrypoint have `pair :permit_protected anything (option %permit bla-bla)` argument, it's a variation of this permit interface.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -39,19 +39,31 @@ module Lorentz.Contracts.BaseDAO.Types
...
+  , pattern NoPermit
+
   , unfrozenTokenId
   , frozenTokenId
 
   , baseDaoAnnOptions
   ) where
 
-import Universum hiding (drop, (>>))
+import Universum (Each, Num(..), Show)
```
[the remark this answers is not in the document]
Martoon-00: I added `Show` to `Lorentz` recently, but this is not in morley-0.10.0, that MR is to be merged yet.

src/Lorentz/Contracts/BaseDAO.hs
Martoon-00: There were some reorganiztion in `StoreClass`, and `KnownValue` constraint was moved to `instance StoreHasSubmap BigMap`, so serving `sProposals` map now requires `KnownValue` from `pm` :pled:

docs/specification.md
Martoon-00: Oh, fair

src/Lorentz/Contracts/BaseDAO/Types.hs
Martoon-00: I thought that there is no benefit in annotating it, but probably there is nothing bad too.

test/Test/BaseDAO/Proposal.hs
Martoon-00: Oh right, that's simpler.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -267,6 +283,7 @@ data Parameter proposalMetadata
   | Mint MintParam
   | Transfer_contract_tokens TransferContractTokensParam
   | Token_address TokenAddressParam
+  | GetCounter (View () Nonce)
```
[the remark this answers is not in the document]
Martoon-00: Hm, apparently I misinterpreted the notion of off-chain views.

It seems to me that this entrypoint is not very temporary, in a sense that we probably shouldn't assume that everyone will be able to work with off-chain views in the nearest months. Since without counter permissions feature is unusable, it would be nice to provide some conservative mechanism. Though on the other hand, probably only one product will work with our DAOs - DAO middleware.

Okay, since we are not tied to TZIP-17 in implementing this form of getter, I'll rename it to `GetVotePermissionCounter` and mention it in spec.

Martoon-00:
> Did you consider that approach?

Well, I can implement it and compare gas consumption and operation size if necessary, there should be no problem here, but my opinion is that the current form is more convenient both for us and for users. I thought you see no much difference in these two and let me pick some )

src/Lorentz/Contracts/BaseDAO/Types.hs
gromakovsky: AFAIU it will be annotated in any case, the difference is snake_case vs camlCase.

src/Lorentz/Contracts/BaseDAO/Types.hs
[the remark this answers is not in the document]
gromakovsky: > Since without counter permissions feature is unusable

One can just read it directly from the storage without calling a view entrypoint. For off-chain access it's more convenient.

src/Lorentz/Contracts/BaseDAO/Types.hs
Martoon-00: Ah, mmm, now that does not sound like a good default :thinking: :thinking:

gromakovsky: Why is it not a good default? We want to annotate as much as possible for clarity to make it easier to understand storage/parameter types by looking at Michelson code and using field names sounds like the most natural way to generate annotations.

gromakovsky (approved):

PR #70, opened 2020-12-11
[#29] Final points

### LINKED ISSUE #29: Refine/improve documentation

Martoon-00:
## Description

Some last changes related to #29.

## Related issue(s)

Resolves #29

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -11,6 +11,7 @@ SPDX-License-Identifier: LicenseRef-MIT-TQ
 * [Contract logic](#contract-logic)
 * [Errors](#errors)
 * [Entrypoints](#entrypoints)
+* [TZIP-16 metadata](#tzip-16-metadata)
 
 # Overview
 
@@ -829,3 +830,9 @@ nat
 ```
 
 - For `vote` entrypoint with permit, returns the current suitable counter for constructing permit signature.
+
+# TZIP-16 metadata
+
+We do not provide a proper support for contract metadata until [the proposal](https://gitlab.com/tzip/tzip/-/merge_requests/94) is finalized and merged.
+
+Except for basic fields, we are going to include `GetCounter` off-chain view.
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -79,7 +79,7 @@ data Config contractExtra proposalMetadata = Config
   -- only 50 tokens will be unfrozen and other 50 tokens will be burnt.
   , decisionLambda :: Lambda
       (Proposal proposalMetadata, Storage contractExtra proposalMetadata)
-      (List Operation, contractExtra)
+      (List Operation, Storage contractExtra proposalMetadata)
   -- ^ The decision lambda is executed based on a successful proposal.
   -- It has access to the proposal, can modify `contractExtra` and perform aribtrary
   -- operations.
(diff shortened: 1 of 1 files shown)


Martoon-00:
It seems to me that I depend on a commit in hackage.nix which is further than the merged fix, but the issue is here. What am I doing wrong :thinking:

zhenyavinogradov:
The problem is that `sha256` for haskell.nix is from the old pin, where the snapshot wasn't working, so nix tries to use the cached version of haskell.nix instead of downloading the new one. But anyway, I'd keep using our fork for now, because it's based on the same commit that most of our other projects are using, so it would avoid rebuilding ghc and some libraries. At some point we'll update haskell.nix over all repos at once, and we'll repin it to upstream here as well.

Martoon-00:
Ah okay, I somehow thought sha256 is updated after I do `niv update hackage.nix`.

Anyway, I'll remove that commit for now.

gromakovsky:
> But anyway, I'd keep using our fork for now, because it's based on the same commit that most of our other projects are using, so it would avoid rebuilding ghc and some libraries.

Do you mean that the diff between `haskell.nix` revision used here and the one used in most of our repos is much smaller than between the latest upstream master version and the one used in most of our repos? And in particular this first diff doesn't include ghc and some libraries (i. e. they don't differ), but the second diff does include them?

gromakovsky:
> FA2 metadata handling is currently not implemented, but people ask questions about metadata. Ideally we'd like to do #42 first, i. e. implement TZIP-16 metadata from the beginning rather than implement the current FA2 way (which will not be recommended soon).

Oh, meh, apparently I forgot to finish this item or some part of it was lost :facepalm: 

I wanted to propose to write explicitly (e. g. in `specification.md`) that metadata handling is currently in some intermediate state and will be updated when https://gitlab.com/tzip/tzip/-/merge_requests/94 is merged.

> Resolves (partially) #29

Why partially this time? It kinda contradicts your "final" and "last" :)
Is it because you want to wait for `haskell.nix` update? If there is a good reason not to do it now, I think we can forget about it until someone creates an issue about updating revisions in all our repos.
Or because I wrote some vague "Ideally we'd like to do #42 first"? It was an unfinished item :)

Martoon-00:
> Why partially this time? It kinda contradicts your "final" and "last" :)

I always could create `Final final points` :smile: :pled:.

docs/specification.md
```diff
@@ -831,3 +829,9 @@ nat
 ```
 
 - For `vote` entrypoint with permit, returns the current suitable counter for constructing permit signature.
+
+# TZIP-16 metadata
```
gromakovsky, line 833: I suppose it should be added to ToC.

gromakovsky (approved):
Ok, one final request.

zhenyavinogradov:
@Martoon-00 `niv` usually updates both revision and sha256 automatically, I'm not sure why it didn't.

@gromakovsky Yes, the diff between `haskell.nix` revision used in other repos and the branch used here contains just the fix for the snapshot path and no other changes, so this repo will use the same ghc build.

Issue #31, opened 2020-10-27
Check the updated value of a field in `Storage` in tests to ensure the value is updated properly


rinn7e:
# Clarification and motivation

Currently, it is not possible to do a `checkStorage` on a `Storage` that contains `BigMap`. 
We are not able to check just a field value of a `Storage` that contains `BigMap` as well.

This prevents from check the updated value properly in the tests for proposal-related entrypoints.

Thread for more detail: https://serokell.slack.com/archives/CF2UVU8RX/p1603886730046500?thread_ts=1603706879.484000&cid=CF2UVU8RX

# Acceptance criteria
- When it is possible to check a value of a field `Storage`, use it to check the updated storage in the proposal-related tests to ensure it works properly.
- All TODOs related to #31 is removed.

Martoon-00:
I want to wait to see how https://gitlab.com/morley-framework/morley/-/issues/346 will be resolved unless it takes long, the currently available solutions are not very pleasant.

gromakovsky:
morley#346 is now in progress and it's pretty much required here. Let's wait for it. I am unassigning this issue for now, let's come back when morley#346 is available.

sras:
We have also skipped tests in [this PR](https://github.com/tqtezos/baseDAO/pull/149) for lack of this capability, so we should also implement tests there as well, once this capability is available.

sras:
Resolved by https://github.com/tezos-commons/baseDAO/pull/294

PR #71, opened 2020-12-11
[#69] Allow `flush` to accepts an amout of proposals, add `dropProposals` entrypoint

### LINKED ISSUE #69: Allow `flush` to accepts an amout of proposals, add `dropProposals` entrypoint
When calling `flush`, `decisionLambda` or `rejectedValue` lambda may fail due to unexpected error. We should have a way to drop some proposals and a way to not flush proposals all at one.

# Acceptance criteria
1. Allow `flush` to accepts an amount of proposals it should consume.
2. Add `dropProposals` entrypoint that will drop a number of proposals from the current ongoing proposals.

rinn7e:
## Description
Problem: When calling `flush`, `decisionLambda` or `rejectedValue` lambda may fail due to unexpected error. We should have a way to drop some proposals and a way to not flush proposals all at one.

Solution: 
- Refactor `sProposalKeyListSortByDate` to be a map with `(Timestamp, ProposalKey proposalMetadata)` as the key, and `()` as the value (Since value here is not important).
- Add `dropProposals` entrypoint
- Update `flush` entrypoint to accept a (Maybe Natural)
- Add tests for these new features

## Related issue(s)

Resolves #69

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -251,6 +251,7 @@ Full list:
 * [`vote`](#vote)
 * [`token_address`](#token_address)
 * [`flush`](#flush)
+* [`drop_proposal`](#drop_proposal)
 * [`migrate`](#migrate)
 * [`confirm_migration`](#confirm_migration)
 
@@ -750,15 +751,19 @@ Parameter (in Michelson)
 
 ```haskell
 data Parameter proposalMetadata
-  = Flush ()
+  = Flush (Maybe Natural)
 ```
 
 Parameter (in Michelson):
 ```
-unit
+option nat
 ```
 
-- Finish voting process on all proposals for which voting period is over.
+- Finish voting process on an amount of proposals:
+  - If the parameter is `None`, process on all the proposals for which voting period is over.
+  - If the paramter is `Just n`, process `n` proposals for which voting period is over.
+- The order of processing proposals are from 'the oldest' to 'the newest'. The proposals which
+have the same timestamp due to being in the same block, are processed in the order of their proposal keys.
 - Frozen tokens from voters and proposal submitter associated with those proposals are returned
   in form of unfrozen tokens:
   - If proposal got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
@@ -769,6 +774,24 @@ unit
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
 - The lost of frozen tokens is due to the fact that the administrator has the right to `burn` or `transfer` frozen tokens of any proposers or voters.
 - If proposal is accepted, decision lambda is called.
+
+### **drop_proposal**
+
+```haskell
+data Parameter proposalMetadata
+  = Drop_proposal ProposalKey
+```
+
+Parameter (in Michelson):
+```
+bytes
+```
+
+- Delete a finished and accepted proposal that is not flushed. Tokens frozen for this
+proposal are returned to the proposer and voters in full. The decision lambda is skipped.
+- This entrypoint should only be used when there is a proposal that is stuck due to having a
+failing decision lambda.
+- Fails with `NOT_ADMIN` if the sender is not the administrator.
 
 ## Migrating functions
 
--- a/src/Lorentz/Contracts/BaseDAO/Doc.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Doc.hs
@@ -21,6 +21,7 @@ module Lorentz.Contracts.BaseDAO.Doc
    , setVotingPeriodDoc
    , setQuorumThresholdDoc
    , flushDoc
+   , dropProposalDoc
 
    , burnDoc
    , mintDoc
@@ -145,6 +146,15 @@ flushDoc = [md|
   There is a possibility of some tokens being lost due to administrator perform
   of burn or transfer operation.
   If the proposal is accepted, the decision lambda is called.
+  |]
+
+dropProposalDoc :: Markdown
+dropProposalDoc = [md|
+  Delete a finished and accepted proposal that is not flushed. Tokens frozen for this
+  proposal are returned to the proposer and voters in full. The decision lambda is skipped.
+
+  This entrypoint should only be used when there is a proposal that is stuck due to having a
+  failing decision lambda.
   |]
 
 burnDoc :: Markdown
--- a/src/Lorentz/Contracts/BaseDAO/Proposal.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Proposal.hs
@@ -8,11 +8,13 @@ module Lorentz.Contracts.BaseDAO.Proposal
   , setVotingPeriod
   , setQuorumThreshold
   , flush
+  , dropProposal
   ) where
 
 import Lorentz
 import Lorentz.Contracts.BaseDAO.Doc
-  (flushDoc, proposeDoc, setQuorumThresholdDoc, setVotingPeriodDoc, voteDoc)
+  ( flushDoc, proposeDoc, setQuorumThresholdDoc, setVotingPeriodDoc, voteDoc
+  , dropProposalDoc )
 import Lorentz.Contracts.BaseDAO.Management (authorizeAdmin, ensureNotMigrated)
 import Lorentz.Contracts.BaseDAO.Permit
 import Lorentz.Contracts.BaseDAO.Token.FA2 (creditTo, debitFrom)
@@ -53,15 +55,10 @@ ensureProposalIsUnique
 ensureProposalIsUnique = do
   dupTop2
   sender; swap; pair; toProposalKey
-  swap; stToField #sProposalKeyListSortByDate
-  stackType @([ProposalKey pm] : ProposalKey pm : ProposeParams pm : store : s)
-  iter $ do
-    dip dup
-    if IsEq then
-      failCustom_ #pROPOSAL_NOT_UNIQUE
-    else
-      nop
-  drop
+  stMem #sProposals
+  if Holds then
+    failCustom_ #pROPOSAL_NOT_UNIQUE
+  else nop
 
 toProposalKey
   :: forall pm s. NicePackedValue pm
@@ -180,29 +177,29 @@ addProposal
   => '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
 addProposal = do
   ensureProposalIsUnique
+  now; pair -- pStartDate
   constructT @(Proposal pm)
     ( fieldCtor $ push (0 :: Natural) -- initial upvote
     , fieldCtor $ push (0 :: Natural) -- initial downvote
-    , fieldCtor $ now -- pStartDate
-    , fieldCtor $ getField #ppProposalMetadata
+    , fieldCtor $ do dup; car
+    , fieldCtor $ do dup; cdr; toField #ppProposalMetadata
     , fieldCtor $ sender
-    , fieldCtor $ getField #ppFrozenToken
+    , fieldCtor $ do dup; cdr; toField #ppFrozenToken
     , fieldCtor $ push []
     )
+  stackType @[Proposal pm, (Timestamp, ProposeParams pm), store]
   dip swap
-  stackType @[Proposal pm, store, ProposeParams pm]
   some
-  duupX @3; sender; swap; pair; toProposalKey
-  stackType @[ProposalKey pm, Maybe (Proposal pm), store, ProposeParams pm]
-  stUpdate #sProposals
-
-  stGetField #sProposalKeyListSortByDate
-  duupX @3; sender; swap; pair; toProposalKey
-  stackType @[ProposalKey pm, [ProposalKey pm], store, ProposeParams pm]
-  cons -- Assuming passed proposal start immediately
-  stackType @[[ProposalKey pm], store, ProposeParams pm]
+  duupX @3; cdr; sender; swap; pair; toProposalKey
+  stackType @[ProposalKey pm, Maybe (Proposal pm), store, (Timestamp, ProposeParams pm)]
+  dup; dip $ stUpdate #sProposals
+  stackType @[ProposalKey pm, store, (Timestamp, ProposeParams pm)]
+  dip $ stGetField #sProposalKeyListSortByDate -- Assuming passed proposal start immediately
+  duupX @4; car; pair; -- create (timestamp, propsalKey)
+  push True; swap; update
   stSetField #sProposalKeyListSortByDate
-  swap
+  stackType @[store, (Timestamp, ProposeParams pm)]
+  swap; cdr
 
 propose
   :: forall store ce pm s.
@@ -440,34 +437,6 @@ setQuorumThreshold Config{..} = do
       stSetField #sQuorumThreshold
       nil; pair
 
-checkIfProposalIsOver
-  :: forall store ce pm s. (NiceParameter pm, StorageC store ce pm)
-  => ProposalKey pm : store : s :-> ((Bool, ProposalKey pm), Proposal pm) : store : s
-checkIfProposalIsOver = do
-  dupTop2
-
-  checkIfProposalExist
-  stackType @(Proposal pm : ProposalKey pm : store : s)
-
-  dup; dip swap; dip $ dip swap;
-  stackType @(Proposal pm : ProposalKey pm : store : Proposal pm : s)
-  toField #pStartDate
-  duupX @3
-
-  stToField #sVotingPeriod; int
-  add -- add to the start_date
-  toNamed #expectedEndDate
-  now; toNamed #currentDate
-
-  if #currentDate >=. #expectedEndDate then
-    push True
-  else
-    push False
-
-  dip $ dip swap
-
-  pair; pair
-
 -- Used in "flush". When unfreezing the tokens of voters/proposers
 -- It is possible that the currentBalance is less than the frozenValue due to
 -- the fact that admin can burn/transfer frozen token.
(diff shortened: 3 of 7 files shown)


src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -217,7 +217,7 @@ data Storage (contractExtra :: Kind.Type) (proposalMetadata :: Kind.Type) = Stor
 
   , sExtra                     :: contractExtra
   , sProposals                 :: BigMap (ProposalKey proposalMetadata) (Proposal proposalMetadata)
-  , sProposalKeyListSortByDate :: [ProposalKey proposalMetadata] -- Newest first
+  , sProposalKeyListSortByDate :: Map Timestamp (ProposalKey proposalMetadata) -- Oldest first
```
rinn7e, line 220: This MR is still WIP, currently allow `flush` to accepts an amount of proposals is done. I will work on adding `dropProposals` entrypoint and their tests next.

One thing that I need advice on is that, currently, I have to update `sProposalKeyListSortByDate` from a list to map of proposal created date and its key.

This is due to 2 main reasons:
- There is currently a bug in `handleProposalIsOver`: https://github.com/tqtezos/baseDAO/blob/master/src/Lorentz/Contracts/BaseDAO/Proposal.hs#L656-L663
  - I implemented this naively, this end up reversing the list. 
  Since we process all the proposals all at the same time, it was hard to detect this.

- Using a list is hard to modify. Since we can handle certain amount of a proposals now, using `Map` is much easier to update/remove the proposal key.

Hope to get your views on this @gromakovsky @Martoon-00 .

Martoon-00: I remember that in another project we also sometimes had to use maps for structures that were initially list-like - because of some Michelson limitations. So I think it's a good change if it simplifies things.

However, note that keeping `Map Timestamp a` may be tricky since `NOW` eventually returns the time of operation inclusion into a block, thus probability of collision is high. You will probably have to use `Map Timestamp [a]`. And if so, I wonder whether will we get any benefits from using Map. 

Another solution is to use numeric indices as keys and remember the indices of the first and last active proposals or something like that. And add an index in that map into `Proposal` datatype. This is gonna be even less pleasant than working with `Map Timestamp` though, but the most correct and flexible.

Could you elaborate on the second point: "Using a list is hard to modify"? AFAIU the first point is caused by the lack of tests on storage (#31) and is quite fixable. Regarding the second point, probably the following dareful plain would be possible to implement:
* Keep the list in "oldest goes first" order;
* Now we don't even need to limit the number of active proposals, because when it is high, `propose` will be the most expensive operation and thus the number of proposals will become naturally limited.
* `ensureProposalIsUnique` can probably be implemented via seeking by proposal hash in proposals map;
* `addProposal` will add to the end of the list;
* `flush` will do a series of `IF_CONS` to consume the list in direct order.

This plan probably has its own issues, but if everything is agreed it seems the simplest to implement to me.

What do you think?

rinn7e: > probability of collision is high

I didn't think of this, and this could be a big issue.

> Using a list is hard to modify
This is pretty much due to having to copy to a new list when we add/remove an element form a list. Since it is actually linear, I guess it is not a performance issue. 

I'm good with your points, and I will make these changes instead of using `Map`.

Martoon-00: > This is pretty much due to having to copy to a new list when we add/remove an element form a list. It is simply a bit tricky, and I was worried about the performance. But in our case, it seems to be linear anyway, I guess that wouldn't be a performance problem.

Yeah, right. Well, if we would use `BigMap Index Proposal`, then we could really have `O(log N)` cost of `propose` and `O(n)` cost for `flush(n)`, though this would be trickier to support.

rinn7e: The issue with `BigMap` is that I don't think we can loop through it using `ITER`.

Related to the collision, I'm thinking of using `Map (Timestamp, ProposalKey) ()`, this way we can guarantee that it has no collision, it should also be sorted with timestamp as well, since it seems in michelson, it compare the left of a pair first.

With this in mind, do you think using `Map` could be better? If not, I'll still go with List.

Martoon-00: > The issue with BigMap is that I don't think we can loop through it using ITER.

Yep, but if we store indices of the first and last active proposals, then in all cases we can iterate by indices, in theory.

> Map (Timestamp, ProposalKey) ()

And then `ITER` over it to get proposals in appropriate order? Right this should work now. I though that there is no way to get all elements with timestamp less than a given value from such a map, but apparently `ITER` suits that.

Okay, I think it's reasonable to use map, should be safer at least, and modifying it is indeed much simpler.

rinn7e:
@gromakovsky @Martoon-00  This MR is ready for review now. 

A few things to note:
1. I have updated `sProposalKeyListSortByDate` to be `Map (Timestamp, ProposalKey proposalMetadata) ()`. 
	- This avoid collision, provide faster add/delete.
	- It is sorted by `Timestamp` from oldest to newest first, then ordered by the proposal key `bytestring` 
2. I have added `pIsDeleted` for proposal, to detect whether or not a proposal has been deleted. User that vote on a deleted proposal will get 'Proposal not exist' error, and flush now will mark all ended proposals as deleted.

3. This is not really necessary, but `dropProposals` should not accept 0 as param, so I added a check which will throw 'bad parameter' error. This applies to when pass `Just 0` to flush entrypoint as well.

> Note that as part of this issue we should also specify the order in which proposals are executed. `flush` can execute multiple proposals and the order may matter. The tricky thing here is that we can't just say "the earliest proposal is executed first" because 2 `propose` calls can be included in the same block and have the same `NOW` timestamp.

Related to your comment, with the new update, `flush` will execute base on the timestamp first, then the bytestring generated by the ProposeParams. I'm not sure how to put this into a proper word, and it is hard to predict the position of this bytestring as well.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -604,63 +581,153 @@ unfreezeVoterToken = do
...
+  constructT @(Proposal pm)
+    ( fieldCtor $ getField #pUpvotes
+    , fieldCtor $ getField #pDownvotes
+    , fieldCtor $ getField #pStartDate
+    , fieldCtor $ getField #pMetadata
+    , fieldCtor $ getField #pProposer
+    , fieldCtor $ getField #pProposerFrozenToken
+    , fieldCtor $ getField #pVoters
+    , fieldCtor $ push True
+    )
```
Martoon-00, line 697: Can't we use just `setField` here? Invoking getters so many times is indeed expensive, maybe avoiding this would save us a lot of operation size.

rinn7e: I did it that where since I actually don't know `setField` exist. I have updated this to use `setField` now. 

Tho, he operation size is reduced to 14880, the test still failed.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -727,3 +735,93 @@ flush config = do
...
+  , NicePackedValue a, KnownValue a, NiceComparable a
+  ) => a -> (a : s) :-> (a : s)
+ensureNot input = do
+  dup; toNamed #val
+  push input; toNamed #input
+  if #input ==. #val then
+    failCustom_ #vOTING_PERIOD_OVER -- TODO: forbidden entrypoint parameter
+  else
+    nop
+
```
Martoon-00, line 843: Yep, these checks look like a good improvement to the contract semantics. 

Though a bit sad that for such things we pay with operation size which is a problem now. I suppose we could also offload such checks to the middleware via mentioning in the description something like "passing 0 is semantically no-op but is still expensive, please don't call us with 0".

Martoon-00:
[This place](https://github.com/tqtezos/baseDAO/pull/71/commits/e3b94c63be8e2df0ee90739e5858b81bda711842#diff-4f7aa07adda34fc46cdfc608faad70f7e9c4eb7a1231d1eaafb34eb564770c09R285) looks like one more candidate for optimization, probably we could use just 3 setters there.

Martoon-00:
Could you explain the difference between `pIsDeleted` flag and removing a proposal from the map? Like, previously we could forget about terminated proposals, but why `pIsDeleted` flag wasn't necessary where and got useful only now?

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -604,63 +577,104 @@ unfreezeVoterToken = do
...
+      stackType @(Proposal pm : store : ProposalKey pm : [Operation] : s)
       push (#isAccepted .! False)
       callCachedFunc $ unfreezeProposerToken config
       unfreezeVoterToken
-      drop
-
-    dip $ drop @(ProposalKey pm)
-
+      deleteProposalKeyFromMap
+      dip drop
```
Martoon-00: Maybe we could also optimize here?: these two `else` branches seem to be identical but pretty big. Maybe we could construct a complex condition using `and`/`or` and have only one `if ... then ... else`? That should save a really lot of space.

Though a sad truth is that Lorentz does not have operators like `&&` and adding them there will take too long (no one severely used this functionality until baseDAO, I added [one](https://gitlab.com/morley-framework/morley/-/issues/505) and [two](https://gitlab.com/morley-framework/morley/-/issues/506) now), but probably we could add a helper that handles our specific condition.

rinn7e:
> Could you explain the difference between `pIsDeleted` flag and removing a proposal from the map? Like, previously we could forget about terminated proposals, but why `pIsDeleted` flag wasn't necessary where and got useful only now?

Actually, we probably don't need `pIsDeleted`. I make a mistake by not thinking about `dropProposals` should only deleted ended proposals that have not been flush. I'll make the fix and come back to this.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -727,3 +735,93 @@ flush config = do
...
+      dig @3;
+      car; dip none; update
+      swap; fromNamed #counter; push (1 :: Natural); add; toNamed #counter
+      dip swap
+    else do
+      -- Increase counter, do nothing
+      dip $ dip drop
+      fromNamed #counter; push (1 :: Natural); add; toNamed #counter
+
+  drop; drop
```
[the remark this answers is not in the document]
rinn7e: Yes, currently this is a batch delete. Before it was affecting ongoing proposals, but now this should only delete finished proposals. 

Since the goal was simply to remove stuck proposals that fail due to decisionLambda, we probably will use this entrypoint to drop the first one, eg. `dropProposals 1`.

I agree that passing `ProposalKey` is better, so if you think we should really do it, I'll make the change. 

cc @gromakovsky

rinn7e:
@Martoon-00 I've made a bunch of changes with change a lot of previous logic, hope to get your review again.

1 question relating to dropping the proposals. How should we deal with the frozen token?
Do we treat this as failing proposals, by simply slash the proposer tokens, and return all voter tokens back?

src/Lorentz/Contracts/BaseDAO/Proposal.hs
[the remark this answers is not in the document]
Martoon-00: Probably my inial proposal about dropping the head proposal was not good. Though let's really ask @gromakovsky.

Martoon-00:
> @Martoon-00 I've made a bunch of changes with change a lot of previous logic, hope to get your review again.
> 
> 1 question relating to dropping the proposals. How should we deal with the frozen token?
> Do we treat this as failing proposals, by simply slash the proposer tokens, and return all voter tokens back?

That's a nice observation. I would also ask @gromakovsky about this, but I suspect that yes. If there is a failing proposal, that's a contract owner to blame, and if he says "there was a mistake and now your funds are lost", no one will be happy ))  Better to return everything to the initial state (though that's a bit sad if we have to write a logic for reversing changes, gonna be a nice pot for bugs)

src/Lorentz/Contracts/BaseDAO/Proposal.hs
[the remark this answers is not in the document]
gromakovsky: What if we pass only one `ProposalKey`, compare it with the head one and remove the head one if they match (failing otherwise)? It should simplify the contract and make it smaller without significantly complicating its usage.

gromakovsky:
I see the following options:
1. Return all tokens that were frozen because the proposal was approved (and failed).
2. Return slashed amount of tokens.
3. Let the administrator specify how much to return as part of the `dropProposals` call.

We can combine them by applying one option to the proposer and a different one to the voters. I will discuss it, but probably I would start with doing (1) for both types of entities (proposer, voters). If a proposal got more positive votes than negative, it's a good proposal even if there is a conflicting one that also got enough votes. So we shouldn't punish anyone by returning less tokens.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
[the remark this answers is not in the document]
rinn7e: > What if we pass only one `ProposalKey`, compare it with the head one and remove the head one if they match (failing otherwise)? It should simplify the contract and make it smaller without significantly complicating its usage.

I'm fine with this as well. If we're good, I'll make the changes.

rinn7e:
> I see the following options:
> 
>     1. Return all tokens that were frozen because the proposal was approved (and failed).
> 
>     2. Return slashed amount of tokens.
> 
>     3. Let the administrator specify how much to return as part of the `dropProposals` call.
> 
> 
> We can combine them by applying one option to the proposer and a different one to the voters. I will discuss it, but probably I would start with doing (1) for both types of entities (proposer, voters). If a proposal got more positive votes than negative, it's a good proposal even if there is a conflicting one that also got enough votes. So we shouldn't punish anyone by returning less tokens.

I'm going with the option 1 for now.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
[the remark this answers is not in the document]
rinn7e: I'm thinking that in some scenario where we know that the proposal is going to be accepted, but containing fail lambda, we probably want to drop it beforehand to let the proposer fixed the lambda, and avoiding wasting time till the voting period ended. 

Being able to drop only the ended proposals could be limiting. 

(Tho, this should be rarely since, accidentality write a fail lambda should not happen regularly)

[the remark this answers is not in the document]
gromakovsky: > I'm thinking that in some scenario where we know that the proposal is going to be accepted, but containing fail lambda, we probably want to drop it beforehand to let the proposer fixed the lambda, and avoiding wasting time till the voting period ended.

Below there is a question about unfreezing tokens in case of failing lambda, I think it's relevant here. There I assumed that we drop only accepted proposals and thus it makes sense to unfreeze all frozen tokens. However, if we can drop potentially negative proposals as well, we may want to reconsider unfreezing.

I think if it's known in advance that the proposal will definitely fail, it simply won't get enough positive votes to pass the quorum threshold and won't be accepted. At least with good UI it sounds like a reasonable assumption.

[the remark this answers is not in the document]
gromakovsky: > probably this does not increase safety

I think requiring the admin to explicitly specify what proposal they are removing is a good idea for safety. Or maybe show a list of proposals and "DROP IT" button near the oldest one. Without a good UI it can be easy to make a mistake, especially if there are 2 proposals in the same block. Another case when it can increase safety is when there are multiple people/computers with admin access who can potentially call `dropProposals` at the same time.

> now when we represent the proposals queue as Map, checking whether the given element matches the head one seems expensive

I didn't review code in this PR closely, but if you think it's expensive, we can assume that the administrator will use a good UI (we can write a CLI client that would take `proposalKey` and check it somehow) and won't be accessed by more than 1 person concurrently.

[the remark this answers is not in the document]
Martoon-00: If we will just remove a proposal by key, without checking it is at the head, this won't be less safe, or not? At least, concurrency issues do not seem to be possible this way.

Also, generally, being able to remove non-head proposals seems to make sense, since if we `flush` by batches, then some intermediate element may turn out to be failing, and we may want to remove it immediately. Though with the current definitions detecting which exactly element is failing may be difficult (the client may need to do `for [n .. 1] dryRunFlush` to find the first failing proposal).

rinn7e:
@gromakovsky @Martoon-00 Continue the discussion from this thread: https://github.com/tqtezos/baseDAO/pull/71#discussion_r546663541

I've updated:
- explicitly specify a proposal key for dropProposal
- return all frozen tokens back to voters and proposer
- only affect a proposal that is ended and accepted but not yet flush (or can't be flush)

The only problem now is that the contract size is too large again. 😕

This is due to having more code in `dropProposal` related to checking whether or not it is an accepted proposal. Hope to get some advice on how to optimize this.

rinn7e:
> @gromakovsky @Martoon-00 Continue the discussion from this thread: [#71 (comment)](https://github.com/tqtezos/baseDAO/pull/71#discussion_r546663541)
> 
> I've updated:
> 
>     * explicitly specify a proposal key for dropProposal
> 
>     * return all frozen tokens back to voters and proposer
> 
>     * only affect a proposal that is ended and accepted but not yet flush (or can't be flush)
> 
> 
> The only problem now is that the contract size is too large again. confused
> 
> This is due to having more code in `dropProposal` related to checking whether or not it is an accepted proposal. Hope to get some advice on how to optimize this.

I've fixed the CI issues now. This PR is ready for re-review.

Martoon-00 (commented):
Overall looks good to me, I have just some minor suggestions.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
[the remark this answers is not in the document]
gromakovsky: > the client may need to do for [n .. 1] dryRunFlush to find the first failing proposal

Humanity has invented binary search :trollface:

src/Lorentz/Contracts/BaseDAO/Doc.hs
```diff
@@ -147,6 +148,15 @@ flushDoc = [md|
   If the proposal is accepted, the decision lambda is called.
   |]
 
+dropProposalDoc :: Markdown
+dropProposalDoc = [md|
+  Delete a finished and accepted proposal that is not flushed. The amount of frozen tokens is
+  are returned to the proposer and votes in full. The decision lamda is skip.
```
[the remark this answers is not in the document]
gromakovsky: > votes

voters

src/Lorentz/Contracts/BaseDAO/Proposal.hs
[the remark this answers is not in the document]
Martoon-00: I assumed a smart `for`, even more difficult from implementation point of view :smile_cat:

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -281,12 +286,13 @@ data Parameter proposalMetadata
   -- Admin
   | Set_voting_period VotingPeriod
   | Set_quorum_threshold QuorumThreshold
-  | Flush ()
+  | Flush (Maybe Natural)
```
gromakovsky, line 289: If we hit the operation size limit and have find time dealing with it, we can remove `Maybe` here because the user usually can just pass a large value instead of `Nothing`. Like, if you pass 1000, most likely all proposals will be processed (even if there are more than 1000 in the contract, which is unlikely, processing more than 1000 will likely fail due to gas costs). Anyway, I don't mind having `Maybe` here for now.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -211,7 +212,7 @@ data Storage (contractExtra :: Kind.Type) (proposalMetadata :: Kind.Type) = Stor
 
   , sExtra                     :: contractExtra
   , sProposals                 :: BigMap (ProposalKey proposalMetadata) (Proposal proposalMetadata)
-  , sProposalKeyListSortByDate :: [ProposalKey proposalMetadata] -- Newest first
+  , sProposalKeyListSortByDate :: Map (Timestamp, ProposalKey proposalMetadata) () -- Oldest first
```
gromakovsky, line 215: It looks isomorphic to a set, why are we using a map here?

gromakovsky (changes requested):
Please also update the specification document. In particular, we should explicitly state in which order proposals are processed (and that proposals from the same block will be ordered by their keys).

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -597,60 +564,138 @@ unfreezeVoterToken = do
...
+    push False
+
+isCounterMet
+  :: forall s. Counter : s
+  :-> Bool : Counter : s
+isCounterMet = do
+  unpair
+  swap; dup
+  stackType @(Maybe Natural : Maybe Natural : "currentCount" :! Natural : s)
+  ifSome (do
```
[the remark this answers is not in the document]
rinn7e: I've updated this, as well as other changes.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: Yes, it is isomorphic to a set. This is due to, initially the map was `Map Timestamp ( ProposalKey proposalMetadata)`, but due to key collision problem, it becomes `Map (Timestamp, ProposalKey proposalMetadata) ()`.

Do you think we should change to `Set` or not.

docs/specification.md
```diff
@@ -769,6 +773,25 @@ unit
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
 - The lost of frozen tokens is due to the fact that the administrator has the right to `burn` or `transfer` frozen tokens of any proposers or voters.
 - If proposal is accepted, decision lambda is called.
+- Fails with `NOT_ADMIN` if the sender is not the administrator.
```
gromakovsky, line 776: Oh, that's somewhat unexpected to me. Why don't we let anyone call `flush`? Is there any harm in letting anyone call it? Ideally it should be called automatically when voting period ends, but since it's not possible, we have to explicitly call it. It doesn't seem to use `SENDER` value for anything. And if a non-admin user calls it before the admin, it will be even better.

It's unrelated to this issue, but if you agree with me, I will create an issue about removing this check or you can do it right in this PR because the change should be small I guess.

src/Lorentz/Contracts/BaseDAO/Types.hs
gromakovsky: If we can change it to `Set`, I think we should do it. It should be more efficient I guess (and we are already close to hitting the operation size limit). Also it adds clarity.

rinn7e: I've updated this now.

docs/specification.md
```diff
@@ -750,15 +751,18 @@ Parameter (in Michelson)
...
 ```
-unit
+option nat
 ```
 
-- Finish voting process on all proposals for which voting period is over.
+- Finish voting process on an amount of proposals:
+  - If the parameter is `None`, process on all the proposals for which voting period is over.
+  - If the paramter is `Just n`, process `n` proposals for which voting period is over.
+- The order of processing proposals are from 'the oldest' to 'the newest'.
```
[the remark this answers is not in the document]

docs/specification.md
rinn7e: I'm good with removing the admin restriction.

Martoon-00 (approved):

gromakovsky (approved):
My review was not very thorough, but what I've seen looks ok. Hopefully @martoon has reviewed everything I skipped :)

PR #79, opened 2020-12-17
[#73] Implement RegistryDAO

### LINKED ISSUE #73: Implement Registry DAO and Treasury DAO

We need to develop 2 DAO contracts: registry and treasury. We don't have an exact specification of what they should do, so we should come up with it ourselves.

Below are some notes that I had some time ago
————————————————————————————————————
For Registry DAO in general case we can have something like:
```
data RegistryProposalMetadata k v = RegistryProposalMetadata
  { agoraPostId :: MText
  , diff :: [RegistryUpdate k v]
  }

data RegistryUpdate k v = RegistryUpdateItem 
  { key :: k
  , newValue :: Maybe v 
  , oldValue :: Maybe v
  }

data RegistryContractExtra k v = RegistryContractExtra 
  { registry :: BigMap k (RegistryEntry v)
  }

data RegistryEntry v = RegistryEntry
  { value :: Maybe v
   -- optional statistics on given proposal
  , lastUpdated :: Timestamp
  }
```

So `contractExtra` is a big map from `k` to `v` that can store extra data for each `v` (such as `lastUpdated` timestamp). In a specific Registry DAO contract `k` and `v` should be set to specific types.
`proposalMetadata` in this case contains a list of items updating the `contractExtra` big_map along with some extra information such as `agoraPostId`.
In this case RegistryDAO is a key-value storage where each update need to go through voting

For Treasury we can have:
```
data ProposalMetadata = ProposalMetadata
  { amount :: Mutez
  , recipientId :: Address
  , agoraPostId :: MText
  }
```
`contractExtra` can be empty, i. e. `unit`. Anyone can add funds to the treasury by calling its `default` entrypoint.
Sometimes people can be rewarded for useful activities. Alice can write an Agora post #123. Then someone can propose giving Alice 10 XTZ for this post. This person creates and submits a new proposal. If it's accepted, we transfer 10 XTZ to Alice (or less if there is not enough balance in the treasury contract)
————————————————————————————————————————————

There are tools called Aragon and Daostack, AFAIK they let you create DAOs from templates. Perhaps there are templates for Treasury and Registry DAO. So we should check how these two types of DAO are represented there. After that, we should write down what exact functionality should be present in this DAOs and discuss it. Once we figure out what it should be, we will proceed to implementation.

# Acceptance criteria

1. There is a clear description/specification of what these 2 types of DAO should do and to which extent they should be customizable.
2. These 2 DAO contracts are present in the contract registry and can be printed (as well as their initial storages).

Notes: lambdas present in the `Config` type should be constant for a specific contract. I mean, for Treasury DAO there is a single definition of each lambda. However, we still might want some configuration, e. g. for Treasury DAO we may want to specify maximal `amount` to be transferred. Configuration of these 2 templates should happen at origination time, not compile time. So parameters should be part of the storage. We can put them into `contractExtra` and use them in lambdas.
The only complexity can happen in Registry DAO if we have to make some types there configurable. In the example above we have `k` and `v` parameters which must be instantiated in compile time. So we can't print the contract without knowing these types. But UI doesn't want to use our Haskell code. So our Haskell code should be used before UI. One possible way to solve this problem is to print stub strings as `k` and `v` types and require the UI to to replace these stubs with actual `k` and `v`. Maybe there are better ways.

rinn7e:
## Description

Problem: We need to develop 2 DAO contracts: registry and treasury for
the new DAO frontend. We implement only RegistryDAO in this MR for now.

Solution: Add `RegistryDAO` according to specification, add tests.

Specification: https://gist.github.com/gromakovsky/8fb38c6992bf779e68241a31630f39d5

## Related issue(s)

Resolves first part of #73

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -14,6 +14,7 @@ import Paths_baseDAO (version)
 
 import BaseDAO.CLI
 import qualified Lorentz.Contracts.GameDAO as GameDAO
+import qualified Lorentz.Contracts.RegistryDAO as RegistryDAO
 import qualified Lorentz.Contracts.TrivialDAO as TrivialDAO
 
 repoSettings :: GitRepoSettings
@@ -32,6 +33,13 @@ contracts = daoContractRegistry
     , dciConfig = GameDAO.config
     , dciExtraParser = pure def
     }
+  , DaoContractInfo
+    { dciName = "RegistryDAO"
+    , dciConfig = RegistryDAO.config @ByteString @ByteString
+    , dciExtraParser = pure def
+    }
+  -- Note: When adding a new contract, CI needs to be updated to
+  -- upload the new contract's documentation.
   ]
 
 main :: IO ()
--- a/src/Lorentz/Contracts/BaseDAO/Proposal.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Proposal.hs
@@ -9,6 +9,7 @@ module Lorentz.Contracts.BaseDAO.Proposal
   , setQuorumThreshold
   , flush
   , dropProposal
+  , toProposalKey
   ) where
 
 import Lorentz
@@ -70,14 +71,13 @@ toProposalKey = do
 ------------------------------------------------------------------------
 -- Propose
 ------------------------------------------------------------------------
-
 checkIsProposalValid
-  :: forall store ce pm.
-     Config ce pm -> '[ProposeParams pm, store] :-> '[ProposeParams pm, store]
+  :: forall store ce pm s. (StorageC store ce pm)
+  => Config ce pm -> (ProposeParams pm : store : s) :-> (ProposeParams pm: store : s)
 checkIsProposalValid Config{..} = do
-  dup
+  dupTop2
   cProposalCheck
-  stackType @[Bool, ProposeParams pm, store]
+  stackType @(Bool : ProposeParams pm : store : s)
   if Holds then
     nop
   else
@@ -208,7 +208,7 @@ propose
 propose config = do
   doc $ DDescription proposeDoc
   stackType @(ProposeParams pm : store : s)
-  framed $ checkIsProposalValid config
+  checkIsProposalValid config
   framed $ checkProposalLimitReached config
   framed $ checkProposerUnfrozenToken
   stackType @(ProposeParams pm : store : s)
@@ -532,7 +532,7 @@ unfreezeProposerToken Config{..} isAccepted = do
           stackType @(Natural : Address : store : Proposal pm : ProposalKey pm : [Operation] : s)
 
         False -> do
-          dup
+          dupTop2
           cRejectedProposalReturnValue
           stackType @("slash_amount" :! Natural : Proposal pm : store : ProposalKey pm : [Operation] : s)
           dip $ do
@@ -709,7 +709,9 @@ flush config = do
   -- guards
   dip $ do
     ensureNotMigrated
+
   ensureMaybeIsNot (Just 0 :: Maybe Natural)
+
   push 0; toNamed #currentCount; pair
   nil
   dig @2; stGetField #sProposalKeyListSortByDate
@@ -757,7 +759,6 @@ dropProposal config = do
   else do
     failCustom_ #fAIL_DROP_PROPOSAL_NOT_OVER
 
--- Have to be divided like this due to Maybe a is not comparable
 ensureMaybeIsNot :: forall a s.
   ( NiceParameter a, NiceStorage a
   , NicePackedValue a, NiceComparable a
--- a/src/Lorentz/Contracts/BaseDAO/Types.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Types.hs
@@ -83,15 +83,17 @@ data Config contractExtra proposalMetadata = Config
   , cUnfrozenTokenMetadata :: FA2.TokenMetadata
   , cFrozenTokenMetadata :: FA2.TokenMetadata
 
-  , cProposalCheck :: forall s
-      . (ProposeParams proposalMetadata) : s
+  , cProposalCheck :: forall s store.
+        StorageC store contractExtra proposalMetadata
+    =>  ProposeParams proposalMetadata : store : s
     :-> Bool : s
   -- ^ A lambda used to verify whether a proposal can be submitted.
   -- It checks 2 things: the proposal itself and the amount of tokens frozen upon submission.
   -- It allows the DAO to reject a proposal by arbitrary logic and captures bond requirements
 
-  , cRejectedProposalReturnValue :: forall s
-      . (Proposal proposalMetadata) : s
+  , cRejectedProposalReturnValue :: forall s store.
+        StorageC store contractExtra proposalMetadata
+    =>  Proposal proposalMetadata : store : s
     :-> ("slash_amount" :! Natural) : s
   -- ^ When a proposal is rejected, the value that voters get back can be slashed.
   -- This lambda specifies how many tokens will be received.
@@ -137,9 +139,9 @@ defaultConfig = Config
       , FA2.tmExtras = mempty
       }
   , cProposalCheck = do
-      drop; push True
+      dropN @2; push True
   , cRejectedProposalReturnValue = do
-      drop; push (0 :: Natural); toNamed #slash_amount
+      dropN @2; push (0 :: Natural); toNamed #slash_amount
   , cDecisionLambda = do
       drop; nil
 
--- a/src/Lorentz/Contracts/RegistryDAO.hs
+++ b/src/Lorentz/Contracts/RegistryDAO.hs
@@ -1,0 +1,179 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | A Registry DAO can be used to store key value pair @(k, v)@
+module Lorentz.Contracts.RegistryDAO
+  ( registryDaoContract
+  , config
+  ) where
+
+import Lorentz
+import Universum ((*))
+
+import qualified Lorentz.Contracts.BaseDAO as DAO
+import qualified Lorentz.Contracts.BaseDAO.Proposal as DAO
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Lorentz.Contracts.RegistryDAO.Types
+import Lorentz.Contracts.RegistryDAO.Doc
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+-- | To pass a proposal check, a proposal needs to have:
+-- @
+--   let s = size(pack(proposalMetadata))
+--
+--   submitTokens == ceFrozenScaleValue * s + ceFrozenExtraValue
+--     &&
+--   s < ceMaxProposalSize
+-- @
+registryDaoProposalCheck
+  :: forall s k v store.
+     ( NicePackedValue (RegistryDaoProposalMetadata k v)
+     , IsoRegistryDaoProposalMetadata k v
+     , DAO.StorageC store (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+     )
+  => DAO.ProposeParams (RegistryDaoProposalMetadata k v) : store : s
+  :-> Bool : s
+registryDaoProposalCheck = do
+  getField #ppProposalMetadata
+  packRaw
+  size; toNamed #proposalSize
+
+  duupX @3
+  stToField #sExtra; toFieldNamed #ceMaxProposalSize; dip dup
+  if #ceMaxProposalSize >. #proposalSize then do
+    fromNamed #proposalSize
+    dig @2
+    stackType @(store : Natural : DAO.ProposeParams (RegistryDaoProposalMetadata k v) : s)
+    stToField #sExtra
+
+    getField #ceFrozenScaleValue; dip swap; mul
+    swap; toField #ceFrozenExtraValue; add
+
+    toNamed #requireValue; swap
+    toFieldNamed #ppFrozenToken; swap
+
+    if #requireValue ==. #ppFrozenToken then do
+      push True
+    else do
+      -- The frozen tokens from the proposal has to be exactly equal to the require frozen tokens.
+      push False
+  else do
+    -- submitted proposal size is bigger than max proposal size
+    dropN @3
+    push False
+
+-- | The slash value of a fail proposal is calculated as:
+-- @
+--   slashValue
+--     = ceSlashScaleValue * pProposerFrozenToken / ceSlashDivisionValue
+-- @
+registryDaoRejectedProposalReturnValue
+  :: forall s k v store.
+      (IsoRegistryDaoProposalMetadata k v
+      , DAO.StorageC store (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+      )
+  =>  (DAO.Proposal (RegistryDaoProposalMetadata k v)) : store : s
+  :-> ("slash_amount" :! Natural) : s
+registryDaoRejectedProposalReturnValue = do
+  toField #pProposerFrozenToken
+  swap
+  stToField #sExtra
+
+  getField #ceSlashScaleValue; dip swap; mul
+  swap; toField #ceSlashDivisionValue; swap; ediv
+  ifSome car $
+    push (0 :: Natural)
+  toNamed #slash_amount
+
+-- | When a proposal is accepted, we add/update the new registry entry
+-- to the storage registry.
+-- We also record the corresponding proposal key and the current time
+-- of when it is applied.
+decisionLambda
+  :: forall k v s store.
+  ( DAO.StorageC store (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+  -- shared
+  , IsoRegistryDaoProposalMetadata k v
+  , NicePackedValue (RegistryDaoProposalMetadata k v)
+  )
+  =>  DAO.Proposal (RegistryDaoProposalMetadata k v) : store : s
+  :-> List Operation : store : s
+decisionLambda = do
+  constructT @(DAO.ProposeParams (RegistryDaoProposalMetadata k v))
+    ( fieldCtor $ getField #pProposerFrozenToken
+    , fieldCtor $ getField #pMetadata
+    )
+  dip $ getField #pProposer; pair; DAO.toProposalKey
+  swap
+  toField #pMetadata
+  caseT
+    ( #cNormalProposalType /-> do
+        toField #npDiff
+        iter $ do
+          stackType @(RegistryUpdate k v : (DAO.ProposalKey (RegistryDaoProposalMetadata k v)) : store : s)
+          constructT @(RegistryEntry k v)
+            ( fieldCtor $ getField #ruNewValue
+            , fieldCtor $ duupX @2
+            , fieldCtor $ now
+            )
+          stackType @(RegistryEntry k v : RegistryUpdate k v : (DAO.ProposalKey (RegistryDaoProposalMetadata k v)) : store : s)
+          dip $ dip swap
+          dip $ toField #ruKey
+          some; swap
+          duupX @3; stToField #sExtra
+          stackType @((RegistryDaoContractExtra k v) : k : Maybe (RegistryEntry k v) : store : (DAO.ProposalKey (RegistryDaoProposalMetadata k v)) : s)
+          dig @2; dig @2; stUpdate #ceRegistry
+          stSetField #sExtra
+          swap
+
+        drop @(DAO.ProposalKey (RegistryDaoProposalMetadata k v))
+        nil
+    , #cConfigProposalType /-> do
+        dip drop
+        duupX @2; stToField #sExtra
+        duupX @2; toField #cpFrozenScaleValue; ifSome (setField #ceFrozenScaleValue) nop
+        duupX @2; toField #cpFrozenExtraValue; ifSome (setField #ceFrozenExtraValue)  nop
+        duupX @2; toField #cpSlashScaleValue; ifSome (setField #ceSlashScaleValue) nop
+        duupX @2; toField #cpSlashDivisionValue; ifSome (setField #ceSlashDivisionValue) nop
+        swap; toField #cpMaxProposalSize; ifSome (setField #ceMaxProposalSize) nop
+        stSetField #sExtra
+        nil
+    )
+
+config ::
+  ( IsoRegistryDaoProposalMetadata k v
+  , NicePackedValue (RegistryDaoProposalMetadata k v)
+  )
+  => DAO.Config (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+config = DAO.defaultConfig
+  { DAO.cDaoName = "Registry DAO"
+  , DAO.cDaoDescription = registryDaoDoc
+  , DAO.cProposalCheck = registryDaoProposalCheck
+  , DAO.cRejectedProposalReturnValue = registryDaoRejectedProposalReturnValue
+  , DAO.cDecisionLambda = decisionLambda
+
+  , DAO.cMaxVotingPeriod = 60 * 60 * 24 * 30 -- 1 months
+  , DAO.cMinVotingPeriod = 1 -- value between 1 second - 1 month
+
+  , DAO.cMaxQuorumThreshold = 1000
+  , DAO.cMinQuorumThreshold = 1
+
+  , DAO.cMaxVotes = 1000
+  , DAO.cMaxProposals = 500
+  }
+
+registryDaoContract ::
+  ( NiceParameter (RegistryDaoProposalMetadata k v)
+  , TypeHasDoc (RegistryDaoProposalMetadata k v)
+  , HasAnnotation (RegistryDaoProposalMetadata k v)
+  , KnownValue (RegistryDaoContractExtra k v)
+  , TypeHasDoc (RegistryDaoContractExtra k v)
+
+  -- shared
+  , IsoRegistryDaoProposalMetadata k v
+  , NicePackedValue (RegistryDaoProposalMetadata k v)
+  ) => Contract
+      (DAO.Parameter (RegistryDaoProposalMetadata k v))
+      (DAO.Storage (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v))
+registryDaoContract = DAO.baseDaoContract config
--- a/src/Lorentz/Contracts/RegistryDAO/Types.hs
+++ b/src/Lorentz/Contracts/RegistryDAO/Types.hs
@@ -1,0 +1,185 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | RegistryDAO Types
+module Lorentz.Contracts.RegistryDAO.Types
+  ( RegistryDaoContractExtra (..)
+  , RegistryEntry (..)
+  , RegistryUpdate (..)
+  , RegistryDaoProposalMetadata (..)
+  , NormalProposal (..)
+  , ConfigProposal (..)
+  , IsoRegistryDaoProposalMetadata
+  , AgoraPostId (..)
+  ) where
+
+import Lorentz
+
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+
+
+data RegistryDaoContractExtra k v = RegistryDaoContractExtra
+  { ceRegistry :: BigMap k (RegistryEntry k v)
+  , ceFrozenScaleValue :: Natural
+  , ceFrozenExtraValue :: Natural
+  , ceSlashScaleValue :: Natural
+  , ceSlashDivisionValue :: Natural
+  , ceMaxProposalSize :: Natural
+  }
+  deriving stock (Generic)
+
+instance (NiceComparable k, Ord k, IsoValue k, IsoValue v)
+  => (IsoValue (RegistryDaoContractExtra k v))
+
+instance (HasAnnotation k, HasAnnotation v)
+  => HasAnnotation (RegistryDaoContractExtra k v) where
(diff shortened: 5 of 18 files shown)


rinn7e:
@gromakovsky  @Martoon-00 This MR is almost ready, and can be reviewed.
A few things to clarify:
1. I've implemented this base on: 
  - Issue desc: https://github.com/tqtezos/baseDAO/issues/73
  - Spec: https://gist.github.com/gromakovsky/8fb38c6992bf779e68241a31630f39d5
  - Add homebase discussion
  
2. Currently, I still use type variables `k` `v`. I think this allows more flexibility in the future, even though for now, we only want `bytestring`.

3. The current frozen token amount is calculated as `s = size(pack(proposalMetadata))`. This can be tricky for the user to put the right amount, so maybe the frontend will help calculate this.

4. We use some variables `a, b, c ..` in the lambda. I'm not sure if I should add them to be configured in the config or not. For now, I just hardcoded the value in those lambdas.

5. It is a bit tricky to test `decisionLambda` to be sure whether or not it updates the registry properly, so hope to get some advice on this as well.

Next, I'll work on cleaning the MR and doing some optimzation.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,125 @@
...
+     )
+  => DAO.ProposeParams (RegistryDaoProposalMetadata k v) : s
+  :-> Bool : s
+registryDaoProposalCheck = do
+  let a :: Natural = 1; b :: Natural = 0
+  getField #ppProposalMetadata
+  packRaw;
+  size;
+  push a; mul
+  push b; add
```
Martoon-00, line 32: Looks pretty beautiful, interesting to see that in stack-machine language it is possible to write such concise code :smile_cat:

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,125 @@
...
+    dip $ toField #ruKey
+    some; swap;
+    duupX @3; stToField #sExtra;
+    stackType @((RegistryDaoContractExtra k v) : k : Maybe (RegistryEntry v) : store : Natural : s)
+    toField #ceRegistry
+    stackType @(BigMap k (RegistryEntry v) : k : Maybe (RegistryEntry v) : store : Natural : s)
+    swap; dip swap; update -- Update `ceRegistry`
+    duupX @2; stToField #sExtra; swap; setField #ceRegistry -- -- Update `sExtra`
+    stSetField #sExtra
+    swap
```
Martoon-00, line 85: Minor suggestion: you could provide a shortcut access to the registry map via defining

```
instance StoreHasSubmap (Storage RegistryDaoContractExtra RegistryDaoProposalMetadata) "sRegistry" k (RegistryEntry v) where
  -- approximate implementation
  storeSubmapOps = composeStoreSubmapOps #ceRegistry storeFieldOps storeSubmapOps  
```

Then calling these getters/setters to update the registry map won't be necessary, you can just call `stInsert` on the entire storage.

src/Lorentz/Contracts/RegistryDAO/Types.hs
```diff
@@ -0,0 +1,105 @@
...
+instance (Ord k, TypeHasDoc k, TypeHasDoc v)
+  => TypeHasDoc (RegistryDaoContractExtra k v) where
+    typeDocMdDescription =
+      "TODO"
+    typeDocMdReference = poly2TypeDocMdReference
+    typeDocHaskellRep = concreteTypeDocHaskellRep @(RegistryDaoContractExtra ByteString ByteString)
+    typeDocMichelsonRep = concreteTypeDocMichelsonRep @(RegistryDaoContractExtra ByteString ByteString)
+
+instance (Ord k) => Default (RegistryDaoContractExtra k v) where
+  def = RegistryDaoContractExtra mempty
```
Martoon-00, line 39: You can avoid `Ord k` constraint if use `def`. `mempty` requires `Ord` only because Monoid also has `<>` where `Ord` is actually necessary.

Martoon-00 (commented):
Looks good to me :+1: 

I have just several minor comments.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,125 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | A Registry DAO can be used to store key value pair `(k, v)`
```
gromakovsky, line 4:
```suggestion
-- | A Registry DAO can be used to store key value pair @(k, v)@
```

src/Lorentz/Contracts/RegistryDAO/Types.hs
```diff
@@ -0,0 +1,105 @@
...
+      "TODO"
+    typeDocMdReference = poly1TypeDocMdReference
+    typeDocHaskellRep = concreteTypeDocHaskellRep @(RegistryEntry ByteString)
+    typeDocMichelsonRep = concreteTypeDocMichelsonRep @(RegistryEntry ByteString)
+
+
+data RegistryDaoProposalMetadata k v = RegistryDaoProposalMetadata
+  { pmAgoraPostId :: Natural
+  , pmDiff :: [RegistryUpdate k v]
+  }
```
gromakovsky, line 66: As mentioned in chat, it should have another constructor to update some parameters.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,125 @@
...
+
+-- | Proposal check: `s = size(pack(proposalMetadata))` `(a * s + b)`
+registryDaoProposalCheck
+  :: forall s k v.
+     ( NicePackedValue (RegistryDaoProposalMetadata k v)
+     )
+  => DAO.ProposeParams (RegistryDaoProposalMetadata k v) : s
+  :-> Bool : s
+registryDaoProposalCheck = do
+  let a :: Natural = 1; b :: Natural = 0
```
gromakovsky, line 27: This `a` and `b` parameters should be taken from the storage. They should be configurable on origination and modifiable in runtime by making a proposal. Also I would name they more descriptively, I just didn't bother coming up with better name, but it should be done.

src/Lorentz/Contracts/RegistryDAO/Types.hs
```diff
@@ -0,0 +1,105 @@
...
+    typeDocMdDescription =
+      "TODO"
+    typeDocMdReference = poly2TypeDocMdReference
+    typeDocHaskellRep = concreteTypeDocHaskellRep @(RegistryDaoProposalMetadata ByteString ByteString)
+    typeDocMichelsonRep = concreteTypeDocMichelsonRep @(RegistryDaoProposalMetadata ByteString ByteString)
+
+data RegistryUpdate k v = RegistryUpdateItem
+  { ruKey :: k
+  , ruNewValue :: Maybe v
+  , ruOldValue :: Maybe v -- why we need oldValue
```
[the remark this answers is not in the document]
gromakovsky: Okay, I think in some cases it can really matter indeed. I don't like `Maybe v` because it forces you to specify the old value (I assume `Nothing` here means "value is absent" rather than "I don't care"), but sometimes you just want to forcefully update a value (maybe delete it). So I think it should be `Maybe (Maybe v)`, but it looked ugly, so I decided not to include it :yao:

Maybe let's not include for now, but suggest it as additional feature to be added in the next version?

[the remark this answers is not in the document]
rinn7e: For me, I'm good with not including this for now.

src/Lorentz/Contracts/RegistryDAO.hs
gromakovsky, line 4:
```suggestion
-- | A Registry DAO can be used to store key value pair @(k, v)@
```

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: I'm not exactly sure how to use `composeStoreSubmapOps`. Currently this causes error: 
```
baseDAO    >     • Datatype DAO.Storage
baseDAO    >                  (RegistryDaoContractExtra k v)
baseDAO    >                  (RegistryDaoProposalMetadata k v) has no field "ceRegistry"
```

Martoon-00: Oh right, I forgot that our contract extra is part of the entire DAO storage.

Then I would define this instance not for `Storage ...` but for `RegistryDaoContractExtra`, and would use `stInsert` on the `RegistryDaoContractExtra`.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: I've updated it so that the `a`, `b`, .. are taken from storage now. I also added `ConfigProposal` which will update this value when accepted. 

I've renamed these value to:
```
data RegistryDaoContractExtra k v = RegistryDaoContractExtra
  { ceRegistry :: BigMap k (RegistryEntry v)
  , ceFrozenScaleValue :: Natural -- a
  , ceFrozenExtraValue :: Natural -- b
  , ceSlashScaleValue :: Natural -- c
  , ceSlashDivisionValue :: Natural -- d
  }
```

Not sure about the naming, hope to get advice on this as well.

Next, I'll work on making the tests for this new config proposal.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: Oh, that makes sense now. I added that, and update the code above to:

```haskell
..
          stackType @((RegistryDaoContractExtra k v) : k : Maybe (RegistryEntry v) : store : Natural : s)
          dig @2; dig @2; stUpdate #ceRegistry
          stSetField #sExtra
          swap
```

Martoon-00: Cool, thanks :+1:

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: I've updated the `ConfigProposal` to match the spec now:
```
data ConfigProposal = ConfigProposal
  { cpFrozenScaleValue :: Maybe Natural -- a
  , cpFrozenExtraValue :: Maybe Natural -- b
  , cpSlashScaleValue :: Maybe Natural -- c
  , cpSlashDivisionValue :: Maybe Natural -- d
  , cpMaxProposalSize :: Maybe Natural -- s_max
  }
```

The only thing left is tests for this config proposal and some code cleanup.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -87,15 +87,17 @@ data Config contractExtra proposalMetadata = Config
   , cUnfrozenTokenMetadata :: FA2.TokenMetadata
   , cFrozenTokenMetadata :: FA2.TokenMetadata
 
-  , cProposalCheck :: forall s
-      . (ProposeParams proposalMetadata) : s
+  , cProposalCheck :: forall s store.
+        StorageC store contractExtra proposalMetadata
+    =>  ProposeParams proposalMetadata : store : s
```
gromakovsky, line 92: Apparently it should be changed in `specification.md` as well.

app/Main.hs
```diff
@@ -32,6 +33,11 @@ contracts = daoContractRegistry
     , dciConfig = GameDAO.config
     , dciExtraParser = pure def
     }
+  , DaoContractInfo
+    { dciName = "RegistryDAO"
+    , dciConfig = RegistryDAO.config @ByteString @ByteString
+    , dciExtraParser = pure def
+    }
```
gromakovsky, line 40: I think we should also update CI to upload Registry DAO documentation. I propose to add a comment somewhere around this list to remind about that.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -143,9 +145,9 @@ defaultConfig = Config
       , FA2.tmExtras = mempty
       }
   , cProposalCheck = do
-      drop; push True
+      drop; drop; push True
```
gromakovsky, line 148: Btw, there should be `dropN @2` and it's likely more efficient.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -29,21 +29,29 @@ registryDaoProposalCheck
...
+    stackType @(store : Natural : DAO.ProposeParams (RegistryDaoProposalMetadata k v) : s)
+    stToField #sExtra
+
+    getField #ceFrozenScaleValue; dip swap; mul
+    swap; toField #ceFrozenExtraValue; add
+
+    toNamed #requireValue; dig @1
+    toFieldNamed #ppFrozenToken; swap
+
+    if #requireValue <=. #ppFrozenToken then do
```
gromakovsky, line 48: I have recently updated it to require strict equality. I think there is no good reason to freeze more than required, so it makes sense to prevent it. I mean, probably the only reason one can freeze 100 tokens when freezing 90 is enough is by mistake.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,159 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | A Registry DAO can be used to store key value pair @(k, v)@
+module Lorentz.Contracts.RegistryDAO
+  ( registryDaoContract
+  , config
+  )
+  where
```
gromakovsky, line 9:
```suggestion
  ) where
```

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,159 @@
...
+
+import Lorentz
+
+import qualified Lorentz.Contracts.BaseDAO as DAO
+import Lorentz.Contracts.RegistryDAO.Types
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Universum hiding (drop, swap, (>>), some)
```
gromakovsky, line 16: AFAIU, `Lorentz` is supposed to be a self-contained prelude and it shouldn't be necessary to import `Universum` (or any other prelude) as well. There is [Lorentz.Prelude](https://hackage.haskell.org/package/lorentz-0.8.0/docs/Lorentz-Prelude.html) which exports a lot of prelude things. So I'd like to better understand what exactly we are importing from `Universum` here and whether it can be added to `Lorentz.Prelude`. And if it's 1-3 things, I think it's better to list them explicitly than listing 4 hidden functions.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,159 @@
...
+import Util.Markdown
+
+-- | Proposal check: `s = size(pack(proposalMetadata)) && s < s_max` `(a * s + b)`
```
gromakovsky, line 19: Analog of markdown's ` is @, so please use it instead. You can do `make haddock-no-deps` to check rendered documentation.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,159 @@
...
+  :: forall s k v store.
+     ( NicePackedValue (RegistryDaoProposalMetadata k v)
+     , IsoRegistryDaoProposalMetadata k v
+     , DAO.StorageC store (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+     )
+  => DAO.ProposeParams (RegistryDaoProposalMetadata k v) : store : s
+  :-> Bool : s
+registryDaoProposalCheck = do
+  getField #ppProposalMetadata
+  packRaw;
```
gromakovsky, line 31: There are some redundant trailing `;`s

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,159 @@
...
+
+
+config ::
+  ( IsoRegistryDaoProposalMetadata k v
+  , NicePackedValue (RegistryDaoProposalMetadata k v)
+  )
+  => DAO.Config (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+config = DAO.defaultConfig
+  { DAO.cDaoName = "Registry DAO"
+  , DAO.cDaoDescription = [md|Registry DAO description|]
```
gromakovsky, line 131: It doesn't look like a good description, can you please write a sentence or two explaining what this DAO does?

src/Lorentz/Contracts/RegistryDAO/Types.hs
```diff
@@ -0,0 +1,162 @@
...
+instance IsoRegistryDaoProposalMetadata k v =>
+    StoreHasSubmap (RegistryDaoContractExtra k v) "ceRegistry" k (RegistryEntry v) where
+  storeSubmapOps = storeSubmapOpsDeeper #ceRegistry
+
+
+data RegistryEntry v = RegistryEntry
+  { reValue :: Maybe v
+   -- optional statistics on given proposal
+  , reLastAgoraPostId :: Natural
+  , reLastUpdated :: Timestamp
```
gromakovsky, line 64: It doesn't match what's written in https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md. We decided to store `ProposalKey` as this "optional statistics", one can query metadata by this key because we don't remove old proposals.

src/Lorentz/Contracts/RegistryDAO/Types.hs
```diff
@@ -0,0 +1,162 @@
...
+    typeDocMdDescription =
+      "TODO"
+    typeDocMdReference = poly2TypeDocMdReference
+    typeDocHaskellRep = concreteTypeDocHaskellRep @(RegistryDaoProposalMetadata ByteString ByteString)
+    typeDocMichelsonRep = concreteTypeDocMichelsonRep @(RegistryDaoProposalMetadata ByteString ByteString)
+
+
+-- | A registry proposal. It will update registry list in contract storage when got accepted.
+data NormalProposal k v = NormalProposal
+  { npAgoraPostId :: Natural
```
gromakovsky, line 106: I would very much prefer to wrap this `Natural` into a newtype because we should never use it as a number, i. e. never apply arithmetic operations to it. Wrapping it into a newtype can prevent us from doing that.

gromakovsky (changes requested):

src/Lorentz/Contracts/RegistryDAO/Types.hs
rinn7e: > Alternatively, we can merge them and store `option v` values.
It should be treated as an implementation detail.
There is additional data to implement configuration lambdas, it's an implementation detail as well.

In this case, I used `option v` alternative instead. I believe this serves the same purpose as having 2 maps.

gromakovsky: Yes, but we don't need to store `reLastAgoraPostId`, we can store `ProposalKey` and one can figure out the post ID by looking up the proposal with given key.

rinn7e: I see, I got this now. I'll update this to use `lastProposalKey` instead of `reLastAgoraPostId`.

src/Lorentz/Contracts/BaseDAO/Types.hs
rinn7e: Updated this.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: That makes sense, I've updated this now.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: Fixed this.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: I see, I only import (*) which is use in voting period to make it more understandable: e.g. `DAO.cMaxVotingPeriod = 60 * 60 * 24 * 30 -- 1 months`.

I updated it to be `import Universum ((*))`

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: I've improved the haddock comments for most functions now.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: Fixed this.

src/Lorentz/Contracts/RegistryDAO.hs
rinn7e: I used the spec description for this now.

src/Lorentz/Contracts/RegistryDAO/Types.hs
rinn7e: Sure, I've updated this to use a new type now.

src/Lorentz/Contracts/RegistryDAO/Types.hs
rinn7e: I've updated this to be a proposal key instead. For now, I'll focus on making the contract smaller to fix the CI in https://github.com/tqtezos/baseDAO/pull/71

app/Main.hs
```diff
@@ -32,6 +33,12 @@ contracts = daoContractRegistry
     , dciConfig = GameDAO.config
     , dciExtraParser = pure def
     }
+  -- TODO: Update CI to upload RegistryDAO documentation
```
gromakovsky, line 36: Oh, that's not what I meant :slightly_smiling_face: 

I'd like you to update CI to upload RegistryDAO _now_ and also add a comment here saying that one should update CI config to upload new documentation when they add a new contract to this list. For example, when we add TreasuryDAO, we want to upload its documentation.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,177 @@
...
+      push True
+    else do
+      -- The frozen tokens from the proposal has to be exactly equal to the require frozen tokens.
+      push False
+  else do
+    -- submitted proposal size is bigger than max proposal size
+    dropN @3
+    push False
+
+-- | The slash value of a fail proposal is calcualted as:
```
gromakovsky, line 64:
```suggestion
-- | The slash value of a fail proposal is calculated as:
```

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,177 @@
...
+
+  getField #ceSlashScaleValue; dip swap; mul
+  swap; toField #ceSlashDivisionValue; swap; ediv
+  ifSome car $
+    push (0 :: Natural)
+  toNamed #slash_amount
+
+-- | When a proposal is accepted, we add/update the new registry entry
+-- to the storage registry.
+-- We also record the its associated agora post id and the current time
```
gromakovsky, line 89:
```suggestion
-- We also record the corresponding proposal key and the current time
```

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,177 @@
...
+registryDaoRejectedProposalReturnValue
+  :: forall s k v store.
+      (IsoRegistryDaoProposalMetadata k v
+      , DAO.StorageC store (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+      )
+  =>  (DAO.Proposal (RegistryDaoProposalMetadata k v)) : store : s
+  :-> ("slash_amount" :! Natural) : s
+registryDaoRejectedProposalReturnValue = do
+  toField #pProposerFrozenToken
+  swap;
```
gromakovsky, line 78:
```suggestion
  swap
```

test/Test/RegistryDAO.hs
```diff
@@ -0,0 +1,150 @@
...
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+-- | Testing a registry-like DAO. Ex. DNS registry
+test_RegistryDAO :: TestTree
+test_RegistryDAO = testGroup "RegistryDAO Tests"
+  [ testGroup "Proposal creator:"
+      [ nettestScenario "can propose a valid proposal" validProposal
+      , nettestScenario "can propose a valid configuration proposal" validConfigProposal
+      ]
+    -- TODO: How to test whether or not registry got updated
```
gromakovsky, line 30: Can we test it only on emulator by checking the updated storage?

gromakovsky (changes requested):

src/Lorentz/Contracts/RegistryDAO.hs
gromakovsky: Other contracts serve just PoC purpose, so I think it doesn't matter and we don't need to change them.

test/Test/RegistryDAO.hs
rinn7e: Yes we can do it. We still have to wait for this issue to be able to check the updated storage: https://gitlab.com/morley-framework/morley/-/issues/346

app/Main.hs
rinn7e: Got this, I've added this to the CI now.

rinn7e:
The originated operation size for `RegistryDAO` now is `16702`, well over the max_size `16384`. I'm pretty much stuck on how to reduce this. 😕

gromakovsky:
Let's merge #71, rebase and then think what to do next (with updated numbers).

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,179 @@
...
+registryDaoProposalCheck = do
+  getField #ppProposalMetadata
+  packRaw
+  size; toNamed #proposalSize
+
+  duupX @3
+  stToField #sExtra; toFieldNamed #ceMaxProposalSize; dip dup
+  if #ceMaxProposalSize >. #proposalSize then do
+    fromNamed #proposalSize
+    dig @2;
```
gromakovsky, line 46:
```suggestion
    dig @2
```

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,179 @@
...
+    stackType @(store : Natural : DAO.ProposeParams (RegistryDaoProposalMetadata k v) : s)
+    stToField #sExtra
+
+    getField #ceFrozenScaleValue; dip swap; mul
+    swap; toField #ceFrozenExtraValue; add
+
+    toNamed #requireValue; dig @1
```
gromakovsky, line 53: `dig @1` is likely equivalent to `swap` which should be more efficient.

test/Test/RegistryDAO.hs
gromakovsky: So can you add a test or two which would use `eiCheckStorage` from `EmulatedImpl`? It doesn't require the storage to have no big_map.

rinn7e: From what I understand `eiCheckStorage` check the whole storage right? Is it possible to simply check a field of the storage, e.g. `sExtra`?

gromakovsky: Oh, that's stupid actually. I don't see a problem to provide a more general method which would let one check a specific field (e. g. by taking a predicate). Probably it just wasn't needed and nobody bothered to expose it. Adding such a method to `EmulatedImpl` and making it available here will likely take some time. While it's not available, I propose to use the Integrational interface to write a test or two with storage checks. It has more functions to check storage, we can use them. Let's not write too many such tests because we will have to update them to use Nettest at some point.

src/Lorentz/Contracts/BaseDAO/Proposal.hs
```diff
@@ -709,7 +709,10 @@ flush config = do
   -- guards
   dip $ do
     ensureNotMigrated
+
+  -- TODO: Disable for now due to too large contract size
```
gromakovsky, line 713: So are we disabling it or not?

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -0,0 +1,179 @@
...
+
+import qualified Lorentz.Contracts.BaseDAO as DAO
+import qualified Lorentz.Contracts.BaseDAO.Proposal as DAO
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Lorentz.Contracts.RegistryDAO.Types
+import Lorentz.Contracts.RegistryDAO.Doc
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+-- | To pass a proposal check, a proposal need to have:
```
gromakovsky, line 21:
```suggestion
-- | To pass a proposal check, a proposal needs to have:
```

gromakovsky (approved):

gromakovsky (approved):

src/Lorentz/Contracts/RegistryDAO/Types.hs
```diff
@@ -0,0 +1,183 @@
...
+  => HasAnnotation (NormalProposal k v) where
+    annOptions = DAO.baseDaoAnnOptions
+
+instance (TypeHasDoc k, TypeHasDoc v)
+  => TypeHasDoc (NormalProposal k v) where
+    typeDocMdDescription =
+      "Describe a proposal for add/update/delete an item in the registry."
+    typeDocMdReference = poly2TypeDocMdReference
+    typeDocHaskellRep = concreteTypeDocHaskellRep @(NormalProposal ByteString ByteString)
+    typeDocMichelsonRep = concreteTypeDocMichelsonRep @(NormalProposal ByteString ByteString)
```
Martoon-00, line 131: Since you mention `ByteString`s as examples, you also need to include them into dependencies, otherwise there will be a reference to a description of this type but the desription itself won't be included. The same for `RegistryUpdate` type.

You can check how it is done for View: https://gitlab.com/morley-framework/morley/blob/master/code/lorentz/src/Lorentz/Macro.hs#L661

Martoon-00 (approved):
Have only one comment about the autodoc, overall looks good :+1:

Issue #57, opened 2020-11-24
Initialize LIGO version development


gromakovsky:
# Clarification and motivation

We need to develop another version of BaseDAO: LIGO version of it. One difference between LIGO and Haskell/Indigo is that LIGO doesn't have compile-time polymorphism that Haskell has, there is no straightforward way to write a polymorphic contract that can be instantiated in compile-time (e. g. its type parameters). So configuration should happen in runtime.
Instead of type parameters we can use `map string bytes` or something similar. We can represent virtually any proposal type as such a map. E. g. if there are 2 fields `foo :: nat` and `bar :: bool`, we can represent it as a map with 2 keys (`"foo"` and `"bar"`) and values packed into `bytes`. We need to unpack them dynamically.

Since we have dynamic unpacking of type parameters, we can make other parameterization dynamic as well. For example, we can put decision lambda into storage instead of inlining it into contract code. This way we can have only one dynamic smart contract that is fully customizable without modification of source code.

Within this issue we shouldn't implement the whole contract (that would be too much), but create some ground on which we can build the contract.

# Acceptance criteria

1. There is a specification for the LIGO version of BaseDAO. Note that there are some substantial changes compared to the Lorentz version, but also they have a lot of stuff in common. So I am not sure how to structure it better (probably they should be in different files, but share common parts), please figure it out while working on the task.
2. There is a stub LIGO contract that is built by our CI and has a simple test-suite written in Haskell using `cleveland` (that also runs on CI).

Martoon-00:
> Note that there are some substantial changes compared to the Lorentz version

There can I see them?

gromakovsky:
They are not specified precisely, but the "Clarification and motivation" section gives some ideas. One of the goals of this issue is to specify them precisely.

The main change is that LIGO version will have dynamic types because it doesn't have static polymorphism (and sometimes compile-time polymorphism might be not desired).

Martoon-00:
I see. So semantically we want the same thing, but code may differ, and so some interesting properties will arise.

Martoon-00:
To achieve polymorphism, can't we proceed as follows?: have one `dao_config.ligo` file which will be the source of all DAO-custom logic, and all other modules of the contract will inline it.

In order to build the contract, we will put modules of baseDAO and `dao_config.ligo` into one folder. Makefile will hide all the building complexity. 

This seems to require just some small effort on providing the infra, and will let us write contracts normally.

gromakovsky:
[...]

Issue #72, opened 2020-12-11
Cleanup Config


gromakovsky:
# Clarification and motivation

1. We have `unfrozenTokenMetadata` and `frozenTokenMetadata`, but they are currently not used. Currently metadata is supposed to be in `sTokenAddress`. When we switch to TZIP-16 metadata, it will be stored differently, but it still won't be embedded into contract code. So I think it shouldn't be part of `Config`. `Config` contains everything necessary to produce a `.tz` file with smart contract code. Metadata is not needed for that, it's needed for origination.
2. AFAIU things like `daoName` and `daoDescription` are currently used only for documentation. They will also be part of metadata, but (see point above) this metadata won't be needed for the contract itself, it will go into storage.

# Acceptance criteria

1. Remove `unfrozenTokenMetadata` and `frozenTokenMetadata` from configuration.
2. Make it clearer that some configuration values are used only for documentation. I think we can keep them in one data type (`Config`), but let's at least provide comments.

Martoon-00:
Note: the first point is done in #98.

PR #81, opened 2020-12-18
[#57] Initialize ligo contract

Martoon-00:
## Description

Stub for our LIGO version of the contract.

- [x] Decide on which syntax to use
- [x] Fill the most basic parts

## Related issue(s)

Initializes #57
Mostly solves #97 as well (except from the point about `:metadata`) 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


ligo/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+# Where to put build files
+OUT ?= out
+
+.PHONY: all trivialDAO build clean clean_all
+
+all: build
+
+# Compile LIGO contract into its michelson representation.
+build: src
```
gromakovsky, line 18: The way we do in `stablecoin` (where the target is an actual generated file, not a phony target) looks like a more proper usage of Makefiles to me, but I don't insist on doing that way.

ligo/Makefile
```diff
@@ -0,0 +1,32 @@
...
+# Ligo executable
+LIGO ?= ligo
+
+# Compile code
+BUILD = $(LIGO) compile-contract --syntax cameligo
+
+# Where to put build files
+OUT ?= out
+
+.PHONY: all trivialDAO build clean clean_all
```
gromakovsky, line 13: `trivialDAO` looks like a leftover.

ligo/Makefile
```diff
@@ -0,0 +1,32 @@
+# SPDX-FileCopyrightText: 2020 TQ Tezos
+# SPDX-License-Identifier: MIT
+
+# Ligo executable
+LIGO ?= ligo
+
+# Compile code
+BUILD = $(LIGO) compile-contract --syntax cameligo
```
gromakovsky, line 8: What syntax are we using? I recall this message in our chat:
> Regrading ligo syntax, let's pick Pascal. The tutorial is written on this dialect so using it should be simpler

ligo/Makefile
```diff
@@ -0,0 +1,32 @@
...
+# Compile LIGO contract into its michelson representation.
+build: src
+	mkdir -p $(OUT)
+	# ============== Compiling contract ============== #
+	$(BUILD) src/base_DAO.mligo base_DAO_contract --output-file $(OUT)/baseDAO.tz
+	# ============== Compilation successful ============== #
+	# See "$(OUT)/baseDAO.tz" for compilation result #
+
+	# strip the surrounding braces and indentation,
+	# note that dollar char is escaped as part of Makefile
```
gromakovsky, line 26: Why are we doing it?

.buildkite/pipeline.yml
```diff
@@ -56,6 +56,11 @@ steps:
       ./result/bin/baseDAO-test --nettest-run-network
       -A localhost -P 8735
 
+  - label: ligo
+    if: *not_scheduled
+    commands:
+    - nix-build ci.nix -A build-ligo
```
gromakovsky, line 62: I think we should also put the resulting contract into artifacts.

ligo/test/Ligo/BaseDAO/Contract.hs
```diff
@@ -0,0 +1,33 @@
...
+  let path = mPath
+          ?: error ("Specify path to contract in " <> show envKey <> " var")
+             -- ↑ Note that our working directory will be './ligo/test'
+  (_uc, tc) <- importContract path
+  return $  tc
+  where
+    rewrapExc = handleAny $ throwM . DisplayExceptionInShow
+
+baseDAOContractLigo :: Contract (ToT Parameter) (ToT FullStorage)
+baseDAOContractLigo = fetchContract "BASEDAO_LIGO_PATH"
```
gromakovsky, line 32: Can we use TH like we do in https://github.com/tqtezos/stablecoin? It sounds safer and cleaner. Maybe we should even add a helper to `morley` to import a contract in compile-time.

Martoon-00:
I added a few small fixes to LIGO types + there are tests (which do not pass and are not checked though yet)

ligo/Makefile
Martoon-00: Do you know how do they use their way? Just `make stablecoin.tz` (as in CI), or their approach allows something smarter?

ligo/Makefile
Martoon-00: Eventually Kirill Kuvshinov opened my eyes and now I see that PascalLigo is not that nice, better to use camelligo.

ligo/Makefile
Martoon-00: LIGO prints a contract as 

```
{ parameter ...;
  storage ...;
  code ...; }
```

This makes the contract more difficult to read (for instance I needed to do `diff baseDAOLorentz.tz baseDAOLigo.tz`, and it got broken because of the braces). Since I needed to implement this, probably it makes sense to leave there (though I don't like these extra lines being printed on each `make build`, not sure what to do with this).

.buildkite/pipeline.yml
Martoon-00: I tried hard but failed, will offload this to Ops team (hopefully after this PR is merged, because I don't want to block anyone).

ligo/test/Ligo/BaseDAO/Contract.hs
Martoon-00: Apparently in this case this makes perfect sense.

ligo/Makefile
gromakovsky: IIRC it was your lowest priority, but ok, I don't have a preference anyway. It means we won't be able to copy-paste FA2 implementation from Stablecoin, but ok, we will port it to a different LIGO syntax, shouldn't be hard.

ligo/Makefile
gromakovsky: I don't think there is anything smarter, but in `Makefile` targets are supposed to be actual files produced by the corresponding commands https://www.gnu.org/software/make/manual/make.html#Rule-Introduction

> A target is usually the name of a file that is generated by a program; examples of targets are executable or object files. A target can also be the name of an action to carry out, such as ‘clean’ (see Phony Targets).

AFAIU there is no big advantage in non-phony targets, it just seems to be more conventional. Well, this convention is from the 20-th century and maybe nobody cares about it nowadays.

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
...
+  ; symbol : string
+  ; name : string
+  ; decimals : nat
+  ; extras : (string, string) map
+  }
+
+type operator =
+  { owner : address
+  ; operator : address
+  }
```
pasqu4le, line 21: It seems to be missing a `token_id` field, at least compared to its Lorentz counterpart.

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
...
+type operators = (operator, unit) big_map
+
+type ledger_key = address * token_id
+type ledger_value = nat
+type ledger = (ledger_key, ledger_value) big_map
+
+// -- DAO base types -- //
+
```
pasqu4le, line 29: I see that there is no equivalent for the `Nonce` type, I suppose that's because it's less useful without the options to make `newtype`s in LIGO, but I'd argue even just a `nat` type synonym could be nice to have (as we do for the other ones).

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
...
+  ; quorum_threshold : nat
+  ; extra : contract_extra
+  ; proposals : (proposal_key, proposal) big_map
+  ; proposals_key_list_sort_by_date : proposal_key list
+  ; permits_counter : nat
+  }
+
+// -- Parameter -- //
+
+type transfer_destination =
```
pasqu4le, line 77: Even tho this is indeed part of the parameter, it's also part of FA2's interface for which there is a "group" above, do you think it would make sense to move this (and all the following types up to `fa2_parameter`) under the `FA2 types` banner?

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
...
+  | Remove_operator of operator_param
+type update_operators_param = update_operator list
+
+type fa2_parameter =
+    Transfer of transfer_params
+  | Balance_of of balance_request_params
+  | Token_metadata_registry of token_metedata_registry_param
+  | Update_operators of update_operators_param
+
+type transfer_ownership_param = address
```
pasqu4le, line 126:
```suggestion
type transfer_ownership_param = [@annot:newOwner] address
```
Again only to match the Lorentz counterpart, I think this should work.

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
...
+
+type migrate_param = address
```
pasqu4le, line 128:
```suggestion
type migrate_param = [@annot:newAddress] address
```

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
...
+  | Vote of vote_param_permited list
+  | Set_voting_period of voting_period
+  | Set_quorum_threshold of quorum_threshold
+  | Flush of unit
+  | Burn of burn_param
+  | Mint of mint_param
+  | Transfer_contract_tokens of transfer_contract_tokens_param
+  | Token_address of token_address_param
+  | GetVotePermitCounter of vote_permit_counter_param
+
```
pasqu4le, line 192: I had written a couple of comments already but then realized this is a bit behind master and the PR needs a rebase.

Listing them here just in case they come useful, compared to the Lorentz version LIGO:
- lacks a `Drop_proposal` equivalent here in `parameter`
  (and as a consequence a corresponding entrypoint function)
- has `unit` as the argument of `Flush`, but it should actually be `nat option`

Maybe some of the comments above also are just about outdated stuff, sorry if that's the case.

pasqu4le (commented):
One last comment about commit messages:

1. I'd remove the `of` from the first line `Start LIGO version of development`
2. `is far not trivial` should be corrected to `is far from trivial` or `is far from being trivial`
3. lacks a "problem/solution` description

ligo/Makefile
Martoon-00: I updated the Makefile, now it does not rebuild the contract if it has already been built. Though the Makefile now looks a bit strange :roll_eyes: because unlike in stablecoin, I put the produced file in gitignored `out` folder, not directly at the `ligo` folder.

ligo/src/base_DAO.mligo
```diff
@@ -0,0 +1,30 @@
...
+      | Propose (p) -> propose(p, store)
+      | Vote (p) -> vote(p, store)
+      | Set_voting_period (p) -> set_voting_period(p, store)
+      | Set_quorum_threshold (p) -> set_quorum_threshold(p, store)
+      | Flush (p) -> flush (p, config, store)
+      | Burn (p) -> burn(p, store)
+      | Mint (p) -> mint(p, store)
+      | Transfer_contract_tokens (p) -> transfer_contract_tokens(p, store)
+      | Token_address (p) -> token_address(p, store)
+      | GetVotePermitCounter (p) -> (failwith("not necessary") : return)
```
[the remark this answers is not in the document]
Martoon-00: This constructor will be removed soon or later because we will have an off-chain view instead, the constructor probably should not never in the LIGO version.

ligo/src/base_DAO.mligo
```diff
@@ -0,0 +1,30 @@
...
+  in (result.0, (result.1, config))
```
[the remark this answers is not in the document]
Martoon-00: Oh this is pretty weird, usually my editor handles trailing newlines. Gonna check what's going on.

ligo/src/permit.mligo
```diff
@@ -0,0 +1 @@
+// Corresponds to Permit.hs module
```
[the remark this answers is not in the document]
Martoon-00: Yes, in Haskell this module contained only helpers, so there is nothing to put here yet.

ligo/src/types.mligo
```diff
@@ -0,0 +1,230 @@
+// Corresponds to Types.hs module
+
+// -- FA2 types -- //
+
+#if !TYPES_H
+#define TYPES_H
```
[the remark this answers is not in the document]
Martoon-00: Since the only way for one module in LIGO to depend on another module is C-like inlining (literally inlining to form one big file), we can import the same module several times and have to deal with duplication, and the recommended way to do this is again C-like preprocessor magic. This is described e.g. on [this fancy page](https://tezosacademy.io/reason/chapter-preprocessor), and I think this applies to any syntax.

Usually, it is possible to inline the same module multiple times without problems, but in our case we have sum types in `types.mligo`, and so importing module twice ends with "conflicting constructor name".

ligo/src/types.mligo
Martoon-00: Could you point me to where do you see that? I can't find it :/

ligo/src/types.mligo
Martoon-00: Okay, added the necessary types to make Lorentz and LIGO versions match.

I only skip `DataToSign`, because IMO there is no sense in a non-polymorphic version of this type; I'll let the implementor of signing logic decide whether to use a record here or just pairs.

ligo/src/types.mligo
Martoon-00: Yes, this was a questionable point. Since this type is a dedicated parameter part, should I also move `fa2_parameter` above?

ligo/src/types.mligo
Martoon-00: Nope, it does not work :/

I suspect that LIGO cannot produce type annotations at all.

ligo/src/types.mligo
Martoon-00: Yes, these are just the very recent additions, should automatically be handled on rebase.

Martoon-00:
One this that remains regarding all this is to decide what do we want to do with mirroring LIGO types?

At the moment I add the mirrored `Parameter`, `Storage` and `Config`.

The easiest option would be to use the new `ligoLayout` right in `Lorentz.Contract.BaseDAO.Types` and remove the mirrors. The shortcoming here is that such layout is suboptimal and operation size will increase - though I think that we don't care about operation size that much now when it is big anyway.

Another option is finishing the way I started, leave the mirrored types, and implement `IsoLigo` instances.

Probably this should be resolved already without me (unless we have a definite dicision today).

ligo/src/types.mligo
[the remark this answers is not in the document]
pasqu4le: Oh I see, so this is a "if it was imported already don't import it again" kinda thing. Got it

P.s. that's indeed a fancy page :sweat_smile:

ligo/src/types.mligo
pasqu4le: Oh you are right, there isn't one.

My bad, I got confused with `operator_param`,

ligo/src/types.mligo
pasqu4le: Not sure, I'm more for the "yes" only because being part of FA2 is more specific than being part of the parameter, but his is indeed subjective.

pasqu4le:
> decide what do we want to do with mirroring LIGO types?

If I understand the problem I think we should consider "bending" the LIGO types to match the Lorentz ones (as they were before these changes) as much as possible.

I say because the `IsoLigo` solution seems excessively cumbersome and even tho we can now easily reproduce LIGO-compatible types in Lorentz we can also produce Lorentz-compatible types in LIGO (unless there's a problem I'm not seeing).

It would take longer (to modify the LIGO types) but this way we'd be able to avoid modifying the Lorentz one and we'd still have that reduced operation cost.

ligo/src/types.mligo
pasqu4le: I thought I had replied to this, but apparently not.

I was for moving that too anyway and it seems you have done so as well, so all good :slightly_smiling_face:

pasqu4le:
This ended up requiring quite a bit more effort than expected.

I finally managed to make all the types between the LIGO version and the Haskell (for tests) version match after rebasing on the current `master`.

Note: I ended up keeping the `IsoLigo` approach, but using `ligoLayout` for the mirror types.

Everything should now be up-to-date and complete (except for the specification that I still have to write).
@rinn7e your review would be appreciated, I no longer expect to make changes here.

ligo/src/types.mligo
```diff
@@ -0,0 +1,238 @@
...
+  { request : balance_request_item
+  ; balance : nat
+  }
+type balance_request_params =
+  [@layout:comb]
+  { requests : balance_request_item list
+  ; callback : balance_response_item list contract
+  }
+
+type token_metedata_registry_param = address contract
```
rinn7e, line 60:
```suggestion
type token_metadata_registry_param = address contract
```

rinn7e (approved):
Just a small typo. The rest looks good to me.

ligo/test/Ligo/BaseDAO/Types.hs
```diff
@@ -0,0 +1,274 @@
...
+    , T.sVotingPeriod = sVotingPeriod
+    }
+
+data Config = Config
+  { cDaoName :: MText
+  , cDaoDescription :: MText
+  , cUnfrozenTokenMetadata :: FA2.TokenMetadata
+  , cFrozenTokenMetadata :: FA2.TokenMetadata
+  , cProposalCheck :: '[ProposeParamsL] :-> '[Bool]
+  , cRejectedProposalReturnValue :: '[ProposalL] :-> '["slash_amount" :! Natural]
```
rinn7e, line 237: This is probably missed, the latest master now require `cProposalCheck` and `cRejectedProposalReturnValue                                                                             ` to have a storage type in the stack as well:   
https://github.com/tqtezos/baseDAO/blob/master/src/Lorentz/Contracts/BaseDAO/Types.hs#L89-L97

rinn7e (changes requested):

ligo/test/Ligo/BaseDAO/Types.hs
```diff
@@ -0,0 +1,274 @@
...
+    , T.sProposals = BigMap . Map.map fromLigo $ unBigMap sProposals
+    , T.sProposalKeyListSortByDate = sProposalKeyListSortByDate
+    , T.sQuorumThreshold = sQuorumThreshold
+    , T.sTokenAddress = sTokenAddress
+    , T.sVotingPeriod = sVotingPeriod
+    }
+
+data Config = Config
+  { cDaoName :: MText
+  , cDaoDescription :: MText
```
rinn7e, line 233: `cDaoName` and `cDaoDescription` are only used in Lorentz autodoc feature. We probably do not need it here.

ligo/test/Ligo/BaseDAO/Types.hs
pasqu4le: :ok_hand: well spotted, added now

ligo/test/Ligo/BaseDAO/Types.hs
pasqu4le: Makes sense to me as well, removed now

ligo/src/types.mligo
```diff
@@ -0,0 +1,235 @@
...
+  ; symbol : string
+  ; name : string
+  ; decimals : nat
+  ; extras : (string, string) map
+  }
+
+type operator =
+  { owner : address
+  ; operator : address
+  }
```
rinn7e, line 24: I'm not exactly sure why. I was testing the fa2 ligo contract and it seems like it needs `layout:comb` here.

```
type operator =
  [@layout:comb]
  { owner : address
  ; operator : address
  }
```

So do we really need this or maybe I did something wrong.

pasqu4le: Yes that look correct. The test for the matching types passes in both cases because they are both `address`, but without `layout:comb` LIGO reorders them alphabetically, so they end up inverted (compared to how we expect them to be).
Fixed now

ligo/test/Ligo/BaseDAO/Types.hs
gromakovsky: Related: https://github.com/tqtezos/baseDAO/issues/72

.buildkite/pipeline.yml
pasqu4le: I haven't added CI test execution either because of this (they need the resulting contract), I also spent some time on it without luck and prefer to do so in a separate PR so that this one can be merged.

pasqu4le:
Note: I simply added the specification in a separate file (`specification-ligo.md`) for now.

It should be referenced somewhere, but then again we also need to write a proper README for the LIGO version because there are too many differences to use the existing one, so I'll leave that to a future PR.

docs/specification-ligo.md
```diff
@@ -0,0 +1,897 @@
...
+
+- The storage of the contract must have annotations for all fields
+  and must be documented to make its interpretation easy for users.
+
+# Configuration
+
+BaseDAO is not a concrete smart contract, but rather a framework to implement various DAOs.
+It can be configured at origination for any specific needs.
+
+In order to do so the contract has two type synonym for `(string, bytes) map`
```
rinn7e, line 38:
```suggestion
In order to do so the contract has two type synonyms for `(string, bytes) map`
```

docs/specification-ligo.md
```diff
@@ -0,0 +1,897 @@
...
+
+- The contract must store tokens of two types: frozen (`token_id` is 1) and unfrozen (`token_id` is 0).
+
+- The storage of the contract must have annotations for all fields
+  and must be documented to make its interpretation easy for users.
+
+# Configuration
+
+BaseDAO is not a concrete smart contract, but rather a framework to implement various DAOs.
+It can be configured at origination for any specific needs.
```
Martoon-00, line 36: Is it true for LIGO version?

docs/specification-ligo.md
```diff
@@ -0,0 +1,897 @@
...
+# Contract logic
+
+This chapter provides a high-level overview of the contract's logic.
+
+- The contract maintains a ledger of address and its balance (frozen and unfrozen tokens)
+- The contract manages a special role called "Administrator".
+- The contract stores a list of proposals that can be in one of the states: "ongoing", "rejected", or "accepted".
+- Migration form a two-step process:
+  + After migration start the contract `migration_status` state is set to `MigratingTo targetAddress` state.
+  + After successful migration confirmation `migration_status` is set to `MigratedTo newContractAddress` state. All the subsequent operations, except `transfer_contract_tokens`, will fail with the `MIGRATED` tag.
```
rinn7e, line 139:
```suggestion
- Migration forms a two-step process:
  + After the migration starts, the contract `migration_status` state is set to `MigratingTo targetAddress` state.
  + After successful migration confirmation, `migration_status` is set to `MigratedTo newContractAddress` state. All the subsequent operations, except `transfer_contract_tokens`, will fail with the `MIGRATED` tag.
```

rinn7e (approved):
Just a few suggestions, the rest looks good from me.

docs/specification-ligo.md
pasqu4le: Fair point, changed it to:
```
BaseDAO is a concrete smart contract, but also a framework to implement various DAOs.
```

docs/specification-ligo.md
pasqu4le: Fixed

docs/specification-ligo.md
pasqu4le: Fixed here and in `specification.md` as well

PR #82, opened 2020-12-19
[#80] TZIP-16 metadata

### LINKED ISSUE #80: Incorporate TZIP-16

[TZIP-16](https://gitlab.com/tzip/tzip/-/blob/7671eb1fa5fdd0a56afa9387757f3a4a024cc2dc/proposals/tzip-16/tzip-16.md) specifies how to incorporate metadata into Michelson smart contracts. It requires `(big_map %metadata string bytes)` in the storage and specifies some standard fields that should be optionally provided.

We need it because:
1. Our original requirements say that `name` and `description` must be provided via TZIP-16 interface.
2. TZIP-12 will recommend using it soon, see https://gitlab.com/tzip/tzip/-/merge_requests/94.
3. We may want to attach additional metadata specific to certain DAOs (e. g. key and value types in Registry DAO).

# Acceptance criteria

1. Any BaseDAO contract satisfies TZIP-16 requirements (mandatory ones).
2. Our CLI exposes a command to print metadata JSON. We will use it for metadata that should be provided by our application, e. g. views. I guess it can be empty for now.

If you think there is something else related to TZIP-16 metadata that can be useful and implemented already, feel free to do it. Note: TZIP-12 metadata will be handled in #42.

Martoon-00:
## Description

Include metadata into our contract that refers to another contract, and provide means to fill baseDAO metadata in that other contract. 

Depends on a [PR in morley-metadata](https://gitlab.com/morley-framework/morley-metadata/-/merge_requests/6).

## Related issue(s)

Resolves #80

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -870,6 +870,17 @@ DAO developer can provide arbitrary entrypoints that will be callable by their n
 
 # TZIP-16 metadata
 
-We do not provide a proper support for contract metadata until [the proposal](https://gitlab.com/tzip/tzip/-/merge_requests/94) is finalized and merged.
+Note: we do not provide a proper support for contract metadata until [the proposal](https://gitlab.com/tzip/tzip/-/merge_requests/94) is finalized and merged.
 
-Except for basic fields, we are going to include `GetCounter` off-chain view.
+This contract implements TZIP-16.
+
+The DAO contract itself won't store the metadata, rather a dedicated contract will contain that.
+Motivation:
+* The baseDAO contract is very large and approaches the Tezos hard limits.
+  * Providing the metadata during origination would significantly reduce the amount of functionality we can put into contract;
+  * Inserting the metadata after origination still requires mechanisms to manage the metadata, increasing the contract size.
+
+## Deployment with metadata
+
+The deployment of contract with metadata has an extra step: a dedicated contract for carrying metadata has to be originated first.
+Then the baseDAO contract should include the reference to a metadata key in the contract in order to be compliant with TZIP-16.
--- a/scripts/test.bats
+++ b/scripts/test.bats
@@ -30,3 +30,7 @@ exe="baseDAO"
 @test "Document BaseDAO" {
   "$exe" document -n TrivialDAO
 }
+
+@test "Print metadata" {
+  "$exe" print-metadata > metadata.json
+}
--- a/scripts/test.bats
+++ b/scripts/test.bats
@@ -34,3 +34,7 @@ exe="baseDAO"
 @test "Print metadata" {
   "$exe" print-metadata > metadata.json
 }
+
+@test "Print MetadataCarrier" {
+  "$exe" print -n MetadataCarrier
+}
--- a/scripts/test.bats
+++ b/scripts/test.bats
@@ -22,7 +22,9 @@ exe="baseDAO"
 }
 
 @test "Print initial BaseDAO storage" {
-  "$exe" storage-TrivialDAO --admin KT1VdyhR9JsmfpxdGf2hMBo1f4Z2TCogp8Th
+  "$exe" storage-TrivialDAO --admin KT1VdyhR9JsmfpxdGf2hMBo1f4Z2TCogp8Th \
+                            --metadata-host-address KT1VdyhR9JsmfpxdGf2hMBo1f4Z2TCogp8Th \
+                            --metadata-key ""
 }
 
 @test "Document BaseDAO" {
--- a/src/BaseDAO/CLI.hs
+++ b/src/BaseDAO/CLI.hs
@@ -21,7 +21,6 @@ import Data.Aeson.Encode.Pretty (encodePretty)
 import qualified Data.Map as Map
 import Data.Version (Version, showVersion)
 import qualified Lorentz.Contracts.Spec.TZIP16Interface as TZIP16
-import Main.Utf8 (withUtf8)
 import qualified Options.Applicative as Opt
 import Options.Applicative.Help.Pretty (Doc, linebreak)
 
@@ -31,6 +30,7 @@ import Lorentz.Doc
 import Lorentz.Value
 import Morley.CLI (addressOption)
 import Util.CLI
+import Util.Main (wrapMain)
 import Util.Named
 
 import qualified Lorentz.Contracts.BaseDAO as DAO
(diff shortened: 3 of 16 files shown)


docs/specification.md
```diff
@@ -833,6 +833,47 @@ nat
...
 
-Except for basic fields, we are going to include `GetCounter` off-chain view.
+This contract implements TZIP-16.
+
+The DAO contract itself won't store the metadata, rather a dedicated contract will contain that.
+Motivation:
+* The baseDAO contract is very large and approaches the Tezos hard limits.
+  * Providing the metadata during origination would significantly reduce the amount of functionality we can put into contract;
+  * Inserting the metadata after origination still requires mechanisms to manage the metadata;
+    this way there is no guarantee that it remains constant (as users will probably expect).
```
gromakovsky, line 845: Well, we can make it callable only once, but this "insert after origination" approach increases the size of the main contract (e. g. because there is an extra entrypoint) which is not good (not critically, but still).

docs/specification.md
```diff
@@ -833,6 +833,47 @@ nat
...
+
+## Deployment with metadata
```
gromakovsky, line 847: Shouldn't it be in the usage instructions? This specification is supposed to be more or less language/tool independent, but here we are referring to `stack` and implementation details.

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -0,0 +1,46 @@
...
+
+-- | Parts of metadata that can be filled just knowing the contract,
+-- without user's input.
+knownBaseDAOMetadata :: Metadata (ToT MetadataStorage)
+knownBaseDAOMetadata = mconcat
+  [ version "0.0.1"
+  , license $ License "LicenseRef-MIT-TQ" Nothing
+  , authors [Author "Serokell", Author "Tocqueville Group"]
+  , homepage "https://github.com/tqtezos/baseDAO"
+  , interfaces [tzip 12, tzip 16, tzip 17]
```
gromakovsky, line 27: AFAIK `tzip 16` shouldn't be mentioned because it's redundant (one should get `%metadata` before accessing `interfaces` and it already implies that the contract follows TZIP-16).

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -0,0 +1,46 @@
...
+
+-- For metadata storage does not matter
+type MetadataStorage = Storage () ()
+
+-- | Parts of metadata that can be filled just knowing the contract,
+-- without user's input.
+knownBaseDAOMetadata :: Metadata (ToT MetadataStorage)
+knownBaseDAOMetadata = mconcat
+  [ version "0.0.1"
+  , license $ License "LicenseRef-MIT-TQ" Nothing
```
gromakovsky, line 24: I _think_ it should be simply `MIT`. Like here: https://spdx.org/licenses/MIT.html

docs/specification.md
```diff
@@ -841,11 +841,13 @@ The DAO contract itself won't store the metadata, rather a dedicated contract wi
...
+  * Inserting the metadata after origination still requires mechanisms to manage the metadata, increasing the contract size.
 
 ## Deployment with metadata
 
+The deployment of contract with metadata has an extra step: a dedicated contract for carrying metadata has to be originated first.
+Then the baseDAO contract should include the reference to a metadata key in the contract in order to be compliant with TZIP-16.
+
 The proposed way of contract deployment is the following:
 
 ### 1. Construct baseDAO metadata
```
gromakovsky, line 853: So now it's in two places, is it intended?

Martoon-00: Wtf :fuu:

gromakovsky:
Okay, seems to look fine now, I'll take another look when the upstream MR is merged and everything is cleaned up here (you can squash commits for example).

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
Martoon-00: Btw, examples in TZIP-16 mention TZIP-16 as one of implemented interfaces:
https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/examples/example-002.json

Though indeed this does not mean we have to follow this.

gromakovsky: They say: "Each string should allow the consumer of the metadata to know which interfaces and behaviors the contract claims to obey (other than the obvious TZIP-016)."

I think "other than the obvious TZIP-016" means we don't have to include it. Probably it's ok to include it (not an error), but it seems useless.

gromakovsky:
Btw, I guess #79 will be merged very soon, so it makes sense to also provide Registry metadata (`k` and `v` types) as part of this PR (I hope it's something relatively small).

Martoon-00:
Looks green now.

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -0,0 +1,43 @@
...
+import Lorentz.Contracts.BaseDAO.Types
+
+-- For metadata storage does not matter
+type MetadataStorage = Storage () ()
+
+-- | Parts of metadata that can be filled just knowing the contract,
+-- without user's input.
+knownBaseDAOMetadata :: Metadata (ToT MetadataStorage)
+knownBaseDAOMetadata = mconcat
+  [ version "0.0.1"
```
gromakovsky, line 22: Can't we take it from `Paths_*`?

src/BaseDAO/CLI.hs
```diff
@@ -73,36 +79,98 @@ data DaoContractInfo =
...
+  -> Opt.Parser a
+  -> String
+  -> Opt.Mod Opt.CommandFields a
+mkCommandParser commandName parser desc =
+  Opt.command commandName $ Opt.info parser $ Opt.progDesc desc
+
 -- | Put this to @main@ to make the executable serve given contracts.
 serveContractRegistry :: String -> DGitRevision -> ContractRegistry -> Version -> IO ()
 serveContractRegistry execName gitRev contracts version =
   withUtf8 $ do
```
gromakovsky, line 156: While we are at it, can you change it to `wrapMain` please and remove `(die . displayException)`?

scripts/test.bats
```diff
@@ -22,7 +22,9 @@ exe="baseDAO"
 }
 
 @test "Print initial BaseDAO storage" {
-  "$exe" storage-TrivialDAO --admin KT1VdyhR9JsmfpxdGf2hMBo1f4Z2TCogp8Th
+  "$exe" storage-TrivialDAO --admin KT1VdyhR9JsmfpxdGf2hMBo1f4Z2TCogp8Th \
```
gromakovsky, line 25: Please add a test that `print-metadata` works.

gromakovsky:
> Btw, I guess #79 will be merged very soon, so it makes sense to also provide Registry metadata (k and v types) as part of this PR (I hope it's something relatively small).

Actually it doesn't sound trivial to do. Apparently the best interface is to have `baseDAO metadata --name RegistryDAO` (like we have `print`, `document` and other commands). So maybe it should be added to `ContractRegistry`. But `ContractRegistry` is not aware of `morley-metadata` and seems to be more general. I guess we shouldn't resolve this problem in this PR, one is supposed to extend metadata provided by our code anyway. I'll create a separate issue.

gromakovsky (approved):

PR #85, opened 2020-12-22
[#73] Implement TreasuryDAO

### LINKED ISSUE #73: Implement Registry DAO and Treasury DAO

rinn7e:
## Description
Problem: We need to develop 2 DAO contracts: registry and treasury for
the new DAO frontend. We implement only RegistryDAO in this MR for now.

Solution: Add `TreasuryDAO` according to specification, add tests.
## Related issue(s)

Resolves #73 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

Problem: We need to develop 2 DAO contracts: registry and treasury for
the new DAO frontend. We implement only RegistryDAO in this MR for now.

Solution: Add `RegistryDAO` according to specification, add tests.

### DIFF
--- a/src/Lorentz/Contracts/TreasuryDAO.hs
+++ b/src/Lorentz/Contracts/TreasuryDAO.hs
@@ -1,0 +1,208 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | A Treasury DAO can be used to store key value pair @(k, v)@
+module Lorentz.Contracts.TreasuryDAO
+  ( treasuryDaoContract
+  , config
+  )
+  where
+
+import Lorentz
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+import Universum ((*))
+
+import qualified Lorentz.Contracts.BaseDAO as DAO
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import Lorentz.Contracts.BaseDAO.Management (ensureNotMigrated)
+import Lorentz.Contracts.TreasuryDAO.Doc
+import Lorentz.Contracts.TreasuryDAO.Types
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+treasuryDaoProposalCheck
+  :: forall s store.
+     ( DAO.StorageC store TreasuryDaoContractExtra TreasuryDaoProposalMetadata
+     )
+  => DAO.ProposeParams TreasuryDaoProposalMetadata : store : s
+  :-> Bool : s
+treasuryDaoProposalCheck = do
+  getField #ppProposalMetadata
+  packRaw;
+  size; toNamed #proposalSize
+
+  duupX @3
+  stToField #sExtra; toFieldNamed #ceMaxProposalSize; dip dup
+  if #ceMaxProposalSize >. #proposalSize then do
+    fromNamed #proposalSize
+    duupX @3;
+    stackType @(store : Natural : DAO.ProposeParams TreasuryDaoProposalMetadata : store : s)
+    stToField #sExtra
+
+    getField #ceFrozenScaleValue; dip swap; mul
+    swap; toField #ceFrozenExtraValue; add
+
+    toNamed #requireValue; duupX @2
+    toFieldNamed #ppFrozenToken; swap
+
+    if #requireValue ==. #ppFrozenToken then do
+      -- Check if the proposal is xtz or not
+      toField #ppProposalMetadata
+      caseT
+        ( #cNormalProposalType /-> do
+            toField #npTransfers
+            push True; swap -- track whether or not it contains proper xtz amount
+            iter $ do
+              caseT
+                ( #cXtzTransferType /-> do
+                    toFieldNamed #xtAmount;
+                    duupX @3; stToField #sExtra
+                    dupTop2;
+                    toFieldNamed #ceMaxXtzAmount
+                    if #ceMaxXtzAmount >=. #xtAmount then do
+                      toFieldNamed #ceMinXtzAmount
+                      if #ceMinXtzAmount <=. #xtAmount then do
+                        nop
+                      else do
+                        drop; push False
+                    else do
+                      dropN @3
+                      push False
+
+                , #cTokenTransferType /-> do
+                    drop;
+                )
+            dip drop
+            -- drop; drop; push True
+        , #cConfigProposalType /-> do
+            dropN @2; push True
+        )
+    else do
+      dropN @2
+      push False
+  else do
+    -- submitted proposal size is bigger than max proposal size
+    dropN @3
+    push False
+
+treasuryDaoRejectedProposalReturnValue
+  :: forall s store.
+      ( DAO.StorageC store TreasuryDaoContractExtra TreasuryDaoProposalMetadata
+      )
+  =>  DAO.Proposal TreasuryDaoProposalMetadata : store : s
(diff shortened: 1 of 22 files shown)


src/Lorentz/Contracts/TreasuryDAO.hs
```diff
@@ -0,0 +1,194 @@
...
+                    push zeroMutez; swap
+                    transferTokens
+                    dip swap; cons; swap
+                  ) $ do
+                    drop; drop; push True
+            )
+
+        if Holds then do
+          -- drop; nil
+          -- Execute fail for now. TODO: Waiting for new issue that deals with fail decision lambda
```
gromakovsky, line 148: The issue is #87 I guess.

test/Test/TreasuryDAO.hs
```diff
@@ -0,0 +1,194 @@
...
+flushTokenTransfer = uncapsNettest $ do
+  ((owner1, _), (owner2, _), dao, admin) <-
+      originateBaseDaoWithBalance def config
+          (\owner1_ owner2_ ->
+              [ ((owner1_, DAO.unfrozenTokenId), 200)
+              , ((owner2_, DAO.unfrozenTokenId), 100)
+              ]
+          )
+  callFrom (AddressResolved admin) dao (Call @"Set_voting_period") 20
+  callFrom (AddressResolved admin) dao (Call @"Set_quorum_threshold") 1
```
gromakovsky, line 75: Btw, I noticed this pattern that we originate with a certain config/storage and then call these two entrypoints to update it. Can't we originate the contract with the right voting period and threshold from the beginning to make less calls? Our tests are already quite slow, so we should avoid making unnecessary calls when possible.

rinn7e: I've updated this now. The PR is mostly ready for review. The only problem is that, we hit the operation size limit again: `oversized_operation\",\"size\":16933,\"max_size\":16384`. 😔

test/Test/TreasuryDAO.hs
```diff
@@ -0,0 +1,186 @@
...
+  checkTokenBalance (DAO.frozenTokenId) dao owner1 0
+  checkTokenBalance (DAO.unfrozenTokenId) dao owner1 210
+  checkTokenBalance (DAO.frozenTokenId) dao owner2 0
+  checkTokenBalance (DAO.unfrozenTokenId) dao owner2 90
+
+flushXtzTransfer :: (Monad m) => NettestImpl m -> m ()
+flushXtzTransfer = uncapsNettest $ do
+  ((owner1, _), (owner2, _), dao, admin) <- originateBaseDaoWithConfig def config
+
+  sendXtz (toAddress dao)
```
gromakovsky, line 111: AFAIU we transfer to `%default` here. Is there a `%default` entrypoint?

rinn7e: There is no default entrypoint. I'm waiting for this: `https://github.com/tqtezos/baseDAO/pull/88`

Should I try sending xtz to another entrypoint in the meantime ? (Have to remove this check: https://github.com/tqtezos/baseDAO/blob/master/src/Lorentz/Contracts/BaseDAO.hs#L103)

gromakovsky: No, I think it's ok to wait, I was just wondering how it works, but AFAIU it doesn't.

gromakovsky:
Okay, now it seems to be not too big and fails with an expected error:
> Parameter Unit ()  does not match the type of KT1BDYZoaUxYZLBbBvsL6pFtxQmH88VBaNyD.

test/Test/TreasuryDAO.hs
rinn7e: I've now rebased on #88, the tests work fine now. 

2 things to note:

- I have to remove `ensureZeroTransfer` for now, since it affects the custom entrypoints as well which results in now allowing to transfer xtz.

- The current way to add custom entrypoints seem to need at least more than 2 entrypoints since it expects a sum-type.

So I have to declare it this way for now to make it work:
```
data TreasuryDaoExtraInterface
  = Default
  | None -- Not needed
```

Hope to get some advice on how to solve/improve this.

Martoon-00: > The current way to add custom entrypoints seem to need at least more than 2 entrypoints since it expects a sum-type.

Another option is using [this magic type wrapper](https://gitlab.com/morley-framework/morley/blob/martoon%2F%23513-custom-generics-ligo/code/lorentz/src/Lorentz/Entrypoints/Helpers.hs#L36) to make only one entrypoint work. 

Apparently its existence is pretty hard to detect, if you have ideas where it is worth mentioning it, please tell.

gromakovsky (approved):

PR #88, opened 2020-12-24
[#86] Add custom entrypoints

### LINKED ISSUE #86: Make it possible to add custom entrypoint(s)

For Treasury DAO we want to be able to add custom entrypoints to the contract. Specifically, we need only `%default` entrypoint that would do nothing and simply receive XTZ. But later we may want to add more entrypoints to other DAOs.

# Acceptance criteria

1. It's sufficiently easy to add a custom entrypoint to the BaseDAO contract. Ideally, it should be easy to add an arbitrary set of entrypoints with arbitrary types and arbitrary code. But that's not strictly required and supporting only `unit %default` which does nothing should be enough to close this issue.
2. There is a proof of concept demonstrating that (1) is indeed sufficiently easy. It can be done by updating GameDAO or creating another custom DAO or by targeting #85 and updating TreasuryDAO from there.

Martoon-00:
## Description

Allow DAOs have custom entrypoints.

If this implementation is justified, I'll update the spec.

## Related issue(s)

Resolves #86

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/Lorentz/Contracts/GameDAO.hs
+++ b/src/Lorentz/Contracts/GameDAO.hs
@@ -58,6 +58,19 @@ data GameDaoProposalMetadata = GameDaoProposalMetadata
 instance HasAnnotation GameDaoProposalMetadata where
   annOptions = DAO.baseDaoAnnOptions
 
+-- | Extra entrypoints supported by our DAO.
+data GameDaoExtraInterface
+  = Default
+  | ResetProposalsCounter
+  deriving stock (Generic)
+  deriving anyclass (IsoValue, HasAnnotation)
+
+instance TypeHasDoc GameDaoExtraInterface where
+  typeDocMdDescription = "Additional entrypoints of Game DAO."
+
+instance ParameterHasEntrypoints GameDaoExtraInterface where
+  type ParameterEntrypointsDerivation GameDaoExtraInterface = EpdPlain
+
 data ProposalType = BalanceType BalanceChange | NewType NewContents
   deriving stock (Generic)
   deriving anyclass (IsoValue)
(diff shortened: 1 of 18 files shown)


src/Lorentz/Contracts/GameDAO.hs
```diff
@@ -203,13 +225,34 @@ decisionLambda = do
...
+        doc $ DDescription "An entrypoint that performs something smart (or not)."
+        stGetField #sExtra
+        push 0
+        setField #ceAcceptedProposalsNum
+        stSetField #sExtra
+        nil
+    )
+
+config :: DAO.Config GameDaoContractExtra GameDaoProposalMetadata GameDaoExtraInterface
+config = DAO.Config
```
gromakovsky, line 246: Does it mean we can't use `defaultConfig` anymore (as we previously did)?

gromakovsky:
There is a small drawback that now we add a redundant entrypoint if we don't need any custom one, but I think it's sufficiently minor for now. Overall this approach looks ok.

src/Lorentz/Contracts/GameDAO.hs
Martoon-00: We can, I just find this pattern not good because one can forget to initialize some fields (e.g. compiler didn't warn me that I forgot to add an implementation for new entrypoints). And values that are inherited from default config (like voting periods) are redefined here anyway.

Mm, actually nothing stops me from putting this change in a separate commit.

gromakovsky:
Ok, fine, it needs a specification update and a rebase.

gromakovsky (approved):

rinn7e (approved):

Issue #65, opened 2020-12-09
Ensure cabal files are up-to-date and don't have hashes


gromakovsky:
# Clarification and motivation

We have `baseDAO.cabal` in this repo which should be manually updated to be in sync with `package.yaml`. Someone may forget to update it. Also we don't want it to include a line like `-- hash: ac81ebe2db072ca09cf1331eb45df29d8279881c629d3d5ec9c171578f19a0b8` because it causes conflicts. And we want everyone to use the same `hpack` version.

# Acceptance criteria

1. There is a script to update `baseDAO.cabal`, it pins a specific `hpack` version.
2. CI checks that `baseDAO.cabal` matches what's produces by the script.
3. Optional: do the same for `cabal.project` and `cabal.project.freeze`.

Example: https://gitlab.com/morley-framework/morley

PR #89, opened 2020-12-25
[#65] Fix cabal file validation in CI

rvem:
## Description
`diff` was not in the PATH and the error checking was improper, so the job was always passing

## Related issue(s)

Relates #65

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/scripts/ci/validate-cabal-files.sh
+++ b/scripts/ci/validate-cabal-files.sh
@@ -37,14 +37,16 @@ stack2cabal
 for file in "${files_to_verify[@]}"; do
     set +e
     diff_res=$(diff "$file" "${file}_backup")
-    if [ "$diff_res" != "" ]; then
+    diff_exit_code="$?"
+    set -e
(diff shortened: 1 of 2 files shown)


scripts/ci/validate-cabal-files.sh
```diff
@@ -37,14 +37,14 @@ stack2cabal
 for file in "${files_to_verify[@]}"; do
     set +e
     diff_res=$(diff "$file" "${file}_backup")
+    set -e
     if [ "$diff_res" != "" ]; then
```
gromakovsky, line 41: Isn't it more proper to check the exit code here?

rvem: :monkahmm:

gromakovsky (approved):
One minor thing: perhaps it'd be nice to also print something like "Cabal files are up-to-date" in successful case to be more confident the script succeeded because everything actually went find or because we simply did something wrong.

PR #90, opened 2020-12-29
Fix RegistryDAO tests failling during scheduled delphinet tests

rinn7e:
## Description

Problem: `validConfigProposal` test in RegistyDAO contains
time-sensitive operations (`flush`). This currently does not work well
in delphinet tests.

Solution: Only run the test in emulator for now.

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e:
@gromakovsky I disabled the failing test now since it used `flush` and I forgot to enable it only for the emulator. This should fix the issue. }

I trigger a manual delphinet test, just to be sure, but it may take quite some times: https://buildkite.com/serokell/basedao/builds/611#e7fc3b2b-80a1-4b6d-b4dc-348cff355d3d/64-79

gromakovsky (approved):

PR #94, opened 2021-01-04
Fix delphinet failing due to short voting period

rinn7e:
## Description

Problem: The recent change makes the default storage having short voting
period. This results in a few tests failing due to time sensitive operation.

Solution: Set a longer voting period for those tests.

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/test/Test/BaseDAO/Proposal.hs
+++ b/test/Test/BaseDAO/Proposal.hs
@@ -129,7 +129,7 @@ nonUniqueProposal = uncapsNettest $ do
 
 voteValidProposal :: (Monad m) => NettestImpl m -> m ()
 voteValidProposal = uncapsNettest $ do
-  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () config
+  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () voteConfig
 
   -- | Create sample proposal (first proposal has id = 0)
   key1 <- createSampleProposal 1 owner1 dao
@@ -161,7 +161,7 @@ voteNonExistingProposal = uncapsNettest $ do
 
 voteMultiProposals :: (Monad m) => NettestImpl m -> m ()
 voteMultiProposals = uncapsNettest $ do
-  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () config
+  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () voteConfig
 
   -- | Create sample proposal
   key1 <- createSampleProposal 1 owner1 dao
@@ -186,7 +186,7 @@ voteMultiProposals = uncapsNettest $ do
 
 insufficientTokenVote :: (Monad m) => NettestImpl m -> m ()
 insufficientTokenVote = uncapsNettest $ do
-  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () config
+  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () voteConfig
 
   -- | Create sample proposal
   key1 <- createSampleProposal 1 owner1 dao
@@ -228,7 +228,7 @@ voteOutdatedProposal = uncapsNettest $ do
 
 voteWithPermit :: (Monad m) => NettestImpl m -> m ()
 voteWithPermit = uncapsNettest $ do
-  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () config
+  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () voteConfig
 
   -- | Create sample proposal
   key1 <- createSampleProposal 1 owner1 dao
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -74,6 +74,22 @@ steps:
     commands:
     - nix-build ci.nix -A haddock --no-out-link
 
+  - label: MANUAL scheduled delphinet test
```
pasqu4le, line 77: IIUC you have added this here to test the changes yourself only and should be removed before merging, am I correct?

test/Test/BaseDAO/Proposal.hs
```diff
@@ -129,7 +129,9 @@ nonUniqueProposal = uncapsNettest $ do
 
 voteValidProposal :: (Monad m) => NettestImpl m -> m ()
 voteValidProposal = uncapsNettest $ do
-  ((owner1, _), (owner2, _), dao, _) <- originateBaseDaoWithConfig @Integer @Empty () config
+  ((owner1, _), (owner2, _), dao, admin) <- originateBaseDaoWithConfig @Integer @Empty () config
+  callFrom (AddressResolved admin) dao (Call @"Set_voting_period") 120
+  callFrom (AddressResolved admin) dao (Call @"Set_quorum_threshold") 4
```
pasqu4le, line 134: I have tried to look into the source of the problem, please confirm with me if this sounds correct to you as well.

I noticed that by default in `BaseDAO.CLI` we set `votingPeriod` and `quorumThreshold` to the values you are setting here, however we don't do so for tests because in `originateBaseDaoWithBalance` the `#votingPeriod` and `#quorumThreshold` are different.

Looking at recent changes it seems this was introduced in [this commit](https://github.com/tqtezos/baseDAO/commit/530ec00df307c5de1e3a0339f5686c76be8aa253).

My question is: instead of adding these two calls in some of these tests, couldn't we modify the `originateData` in `originateBaseDaoWithBalance` to use these values instead?
Or is there another reason to not do this that I am missing?

rinn7e: Some of the tests prefer short voting period since it makes it run faster. 
So I make it as the default and only set longer voting period if necessary.

The failing tests are not detected previously because I thought all time sensitive tests only run in emulator until batch operation is implemented in cleveland. 

But this is not the case, all the failing tests here are vote related tests, and they require longer voting period. They however, do not causes long running time (due to not having to use `advanceTime`), so I keep them in the real network tests and simply set a longer voting period.

pasqu4le: I see, saving time with shorter voting periods when possible makes sense.

However I'd still prefer to set this beforehand instead of making a call to set the values, I think it should be handled with configuration instead.

So I suggest this:
- in `originateBaseDaoWithBalance`'s `mkStorage` call instead of setting the `votingPeriod` to a default value of `20`, let's set it to the `cMinVotingPeriod` value from `config`
- then let's change the `Test.BaseDAO.ProposalConfig` `config` to set the `votingPeriod` to `20` instead of the 7 days from `DAO.defaultConfig`. (Note: I think we should do the same for the other configs in this module as well)
- create a new config in `Test.BaseDAO.ProposalConfig` for these voting tests, which is identical to `config` except having the `votingPeriod` set to `120`.

Plus, we do the same for the `quorumThreshold`.

This way we'd still have most tests to use the lower voting period and these ones to use the bigger one, but without making more calls for it. What do you think?
If you agree please make this in a separate commit, so we can check that there are indeed no problems in doing it this way and revert otherwise.

P.s. while we are at it, we could also change `dropProposal` and perhaps `flushProposalIntegrational` to not set these values with calls but by changing the config they are created with (this time in place)

rinn7e: > - then let's change the `Test.BaseDAO.ProposalConfig` `config` to set the `votingPeriod` to `20` instead of the 7 days from `DAO.defaultConfig`. (Note: I think we should do the same for the other configs in this module as well)

I think you mean `minVotingPeriod` right, since votoingPeriod is stored in `Storage` instead of `Config`. Tho, if we use `minVotingPeriod`, it can still work since we use that field to set the storage. 

I think it would be nice to be able to set the field of the storage directly (instead of using `minVotingPeriod` field as a passer). This will require quite a lot of refactoring. 

So what's your advice on this, should we simply set `minVotingPeriod` in the config, or modify `originateBaseDao` to be able to accepts a custom storage.

pasqu4le: > I think you mean minVotingPeriod

Yeah, that's correct.

IMO we should do this: here let's use `minVotingPeriod` in `mkStorage` since it's easier to do and I think makes sense (these are small values and after all indeed these tests **need** a minimum value of `120`) and let's create a separate issue to consider and eventually make the changes to pass a different `votingPeriod` using the config.

In fact if you agree, please create the issue yourself and add a TODO mentioning it in `originateBaseDaoWithBalance` as well.

test/Test/Common.hs
```diff
@@ -68,6 +68,7 @@ originateBaseDaoWithConfig contractExtra config = do
         ]
     )
 
+-- TODO: [#96] Add a way to customize storage
 originateBaseDaoWithBalance
```
rinn7e, line 72: @pasqu4le I've made the change as you suggested, also added an issue as well: https://github.com/tqtezos/baseDAO/issues/96

You can check this to see if it fits what you have in mind. In the meanwhile, the delphinet tests probably take 1 or 2 hours to finish.

.buildkite/pipeline.yml
rinn7e: Yeah, I'll drop this change when the tests succeed and the MR is ready to be merged.

test/Test/Common.hs
pasqu4le: LGTM

pasqu4le (approved):
LGTM so I'll approve, however we'll have to wait for the tests to run before merging (don't forget to remove the temporary commit as well).

P.s. please modify the "Related issue(s)" part to `None` instead of `Resolves #`

PR #98, opened 2021-01-07
[#42] Implement TZIP-12 with TZIP-16's edits

### LINKED ISSUE #42: Implement TZIP-16 according to TZIP-12

### LINKED ISSUE #107: Update FA2 part of the LIGO version to non-legacy requirements

In #99 we implemented FA2 entrypoints in LIGO according to the legacy FA2 specification with `token_metadata_registry` entrypoint. Since this interface is legacy and not encouraged anymore, we should switch to the new one. We are doing it for the Lorentz version in https://github.com/tqtezos/baseDAO/pull/98, let's do the same for the LIGO version afterwards.

# Acceptance criteria

1. `token_metadata_registry` is removed.
2. There is something analogous to our `print-metadata` command that we have for the Lorentz version. It should print metadata that should be defined by us, not by the contract's users. It should include off-chain views described in TZIP-12 including the `token_metadata` one. We prefer this way of exposing token metadata because it lets us keep the storage of the main contract smaller and thus the contract itself slightly smaller (e. g. less instructions to access parts of the storage).

Martoon-00:
## Description

Add the changes to comply with TZIP-12 + TZIP-16.

- [x] Add off-chain views for LIGO
- [ ] Update the spec?

## Related issue(s)

Resolves #42
Resolves #107 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


gromakovsky:
>  Do we want to have `%token_metadata` to provide redundancy?

Why may we want to do it? AFAIU we need to have `%token_metadata` in the storage OR `token_metadata` off-chain view. I think the latter is better for us because it decreases storage size of the main contract. And I don't see a reason to do both. Btw, you say:
> token_metadata is not necessary when metadata is kept in storage

But AFAIU it's not kept in storage (if you mean `%token_metadata`). And I think it shouldn't be kept there.

> Do we want to provide total_supply?

AFAIU we SHOULD do it. So let's add it and then see how it will affect the contract size.

>  It is not clear how to provide info about FA2 permissions (TZIP-16 mentions some Permissions policies document without a link)

AFAIU we shouldn't provide anything because we follow the default one. The document is here: https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/permissions-policy.md

One more thing: we can (and should) remove `%token_metadata_registry` entrypoint now, it's not needed anymore.

Martoon-00:
> AFAIU we shouldn't provide anything because we follow the default one

Hmm, I now see that we indeed do. I was stuck for a while because morley-ledgers somehow does not export a proper default permission policy :thinking:

Martoon-00:
Okay, this is all done as far as I can judge.

Though I'm yet not sure about the following points:
1. what exactly is worth adding to the specification/README? 
2. there is one TODO about whether do we want to fill token metadata ourselves somehow or accept that from the user after all.
3. looks like now there is no need to have token metadata in Config even for the Lorentz version, remove it from there? Probably this makes sense since token metadata is really just part of metadata now, not part of the contract code.

gromakovsky:
> looks like now there is no need to have token metadata in Config even for the Lorentz version, remove it from there? Probably this makes sense since token metadata is really just part of metadata now, not part of the contract code.

It was supposed to be done in #72 long time ago. You can do it here if you want or leave it to someone else to do in #72.

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -30,17 +38,86 @@ knownBaseDAOMetadata = mconcat
   , views baseDAOViews
   ]
 
+------------------------------------------------------------------------
+-- Off-chain views
+------------------------------------------------------------------------
+
+type DaoView = View (ToT MetadataStorage)
+
 baseDAOViews :: [View $ ToT MetadataStorage]
```
gromakovsky: `DaoView` can be used here.

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -30,17 +38,86 @@ knownBaseDAOMetadata = mconcat
...
-    , vImplementations =
-        [ VIMichelsonStorageView $
-            mkSimpleMichelsonStorageView @MetadataStorage . WithoutParam $
-              stToField #sPermitsCounter
-        ]
-    }
+  [ getBalanceView
+  , allTokensView
+  , isOperatorView
+    -- total_supply is skipped since the necessary info is not tracked
```
gromakovsky, line 93: #126 TODO should be here (maybe it's added in a follow-up commit).

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -30,17 +38,86 @@ knownBaseDAOMetadata = mconcat
...
+getBalanceView = View
+  { vName = "get_balance"
+  , vDescription = Just
+      "Get balance of an address according to TZIP-12."
+  , vPure = Just True
+  , vImplementations =
+      [ VIMichelsonStorageView $
+          mkMichelsonStorageView @MetadataStorage @Natural Nothing [] $
+            WithParam @FA2.BalanceRequestItem $ do
+              forcedCoerce_
```
gromakovsky, line 114: What is coerced here and why can't it be done safer?

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
```diff
@@ -106,6 +108,39 @@ isOperatorView = View
...
+              dup
+              dip $ do
+                get
+                ifSome nop $ failCustom_ #fA2_TOKEN_UNDEFINED
+              pair
+      ]
+  }
+  where
+    -- TODO: what to do here?
+    -- Accept token metadatas in CLI?
```
gromakovsky: I think ideally `print-metadata` should have options like `--token-name` with some sane default values.

gromakovsky:
Nitpicking: probably more commits should be tagged with #107 instead of #42.

ligo/Makefile
```diff
@@ -10,7 +10,7 @@ BUILD = $(LIGO) compile-contract --syntax cameligo
 # Where to put build files
 OUT ?= out
 
-.PHONY: all clean test
+.PHONY: all clean build test
```
gromakovsky, line 13: You added `build` here, but there is no `build` target in this file :confused:

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
Martoon-00: Yep, noticed this later, though with the following commits these types are not so same anymore.

src/Lorentz/Contracts/BaseDAO/TZIP16Metadata.hs
Martoon-00: `FA2.BalanceRequestItem` to `LedgerKey`. Yeah, apparently I better write this via getters.

ligo/Makefile
Martoon-00: :ffuuu:

Martoon-00:
I'm done

src/BaseDAO/CLI.hs
```diff
@@ -121,21 +124,49 @@ tzip16MetadataParser = do
...
+  symbol <-
+    mkCLOptionParser (Just defSymbol) (#name .! (prefix <> "-token-symbol"))
+    (#help .! "Symbol of the token (according to TZIP-12)")
+  name <-
+    mkCLOptionParser (Just defName) (#name .! (prefix <> "-token-name"))
+    (#help .! "Name of the token (according to TZIP-12)")
+  decimals <-
+    mkCLOptionParser (Just defDecimals) (#name .! (prefix <> "-token-decimals"))
+    (#help .! "Decimals field of the token (according to TZIP-12)")
+  return $ FA2.mkTokenMetadata name symbol decimals
```
gromakovsky, line 144: Meh, I am quite surprised that `mkTokenMetadata` accepts any `Text` as `decimals`, but it's our of scope of this issue. I think we should reject any `decimals` value that is not a number, but AFAIU currently we don't.

scripts/test.bats
```diff
@@ -47,6 +48,10 @@ contracts="GameDAO MetadataCarrier RegistryDAO TreasuryDAO TrivialDAO"
   "$exe" print-metadata > metadata.json
 }
 
+@test "Print LIGO metadata" {
+  "$ligo_exe" print-metadata > metadata.json
```
gromakovsky, line 52: Can you please update it now to demonstrate how to use new command line options?

gromakovsky (changes requested):

src/BaseDAO/CLI.hs
gromakovsky: So I propose to parse it as a number and then convert to `Text`.

gromakovsky:
Also please check whether `template` compiles (I am not sure it's checked by CI).

ligo/haskell/test/README.md
```diff
@@ -1,4 +1,4 @@
-# BaseDAO LIGO tests
+# BaseDAO LIGO meta
 
 BaseDAO LIGO tests are implemented in Haskell, using `cleveland` [library](https://gitlab.com/morley-framework/morley/-/tree/master/code/cleveland).
```
pasqu4le, line 3: This probably should have not be renamed and left as `BaseDAO LIGO tests` seeing as this README is specifically under `haskell/tests` and the "Testing" section in `ligo/README` points here.

pasqu4le (commented):
Only a minor nitpick as far as the LIGO version is concerned.

src/BaseDAO/CLI.hs
Martoon-00: Ah okay, I decided that there is a special reason why any `Text` is allowed. Let's really parse as numbers.

Martoon-00:
> Also please check whether `template` compiles (I am not sure it's checked by CI).

Yeah, just noticed it does not :unamused:

ligo/haskell/test/README.md
Martoon-00: Oh right, good catch.

gromakovsky (approved):

PR #99, opened 2021-01-08
[#95] Implement FA2 endpoints for LIGO version

### LINKED ISSUE #95: Implement FA2 endpoints for LIGO version

We need to implement the LIGO version of the contract.

Part of it are the endpoint required by FA2, which should not be affected by any of the differences between the LIGO and Lorentz version of the contract.

Note: we already have an LIGO implementation of FA2 in `stablecoin`, but that's in a different LIGO dialect.
It would, however, be simpler to convert it than to start from scratch.

Note: it depends on #57 but currently this seem to be almost done, so it should be fine to start development on its branch already.

# Acceptance criteria

FA2-related endpoint are implemented in LIGO.

rinn7e:
## Description
Problem: We need to implement LIGO version of BaseDAO contract. First of
all we should implement FA2 entrypoints and test it.

Solution: Implement FA2 entrypoints for LIGO version of BaseDAO and add
its tests.

## Related issue(s)

Resolves #95 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/src/token/fa2.mligo
+++ b/ligo/src/token/fa2.mligo
@@ -1,0 +1,144 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+// Corresponds to Token/FA2.hs module
+
+#include "../types.mligo"
+#include "../management.mligo"
+
+
+let unfrozen_token_id: nat = 0n
+let frozen_token_id: nat = 1n
+
+// -----------------------------------------------------------------
+// Helper
+// -----------------------------------------------------------------
+
+[@inline]
+let check_sender (from_ , store : address * storage): address =
+  if (sender = store.admin) then from_
+  else if (sender = from_) then from_
+  else
+    let key: operator = { owner = from_; operator = sender} in
+    if Big_map.mem key store.operators then
+      from_
+    else
+      (failwith("FA2_NOT_OPERATOR"): address)
+
+let debit_from (amt, from_, token_id, store: nat * address * token_id * storage): storage =
+  match Big_map.find_opt (from_, token_id) store.ledger with
+    Some bal ->
+      let store =
+        match Michelson.is_nat (bal - amt) with
+          Some new_bal ->
+            { store with
+              ledger = Big_map.update (from_, token_id) (Some new_bal) store.ledger
+            }
+        | None ->
+            (failwith ("FA2_INSUFFICIENT_BALANCE") : storage)
+      in store
+  | None ->
+      if (amt = 0n) then store // We allow 0 transfer
+      else (failwith ("FA2_INSUFFICIENT_BALANCE") : storage)
+
+let credit_to (amt, to_, token_id, store : nat * address * nat * storage): storage =
+  let new_bal =
+    match Big_map.find_opt (to_, token_id) store.ledger with
+      Some bal -> bal + amt
+    | None -> amt
+  in  { store with
+        ledger = Big_map.update (to_, token_id) (Some new_bal) store.ledger
+      }
+
+
+// -----------------------------------------------------------------
+// Transfer
+// -----------------------------------------------------------------
+
+let transfer_item (store, ti : storage * transfer_item): storage =
+  let transfer_one (store, tx : storage * transfer_destination): storage =
+    let valid_from_ = check_sender (ti.from_, store) in
+    if tx.amount = 0n then store
+    else (
+      let valid_token_id =
+        if tx.token_id = unfrozen_token_id then tx.token_id
+        else if tx.token_id = frozen_token_id then (
+          if sender = store.admin then tx.token_id
+          else (failwith ("FROZEN_TOKEN_NOT_TRANSFERABLE"): token_id)
+        ) else (failwith ("FA2_TOKEN_UNDEFINED") : token_id)
+      in credit_to
+        ( tx.amount, tx.to_, valid_token_id
+        , debit_from (tx.amount, valid_from_, valid_token_id, store)
+        )
+    )
+  in List.fold transfer_one ti.txs store
+
+let transfer (params, store : transfer_params * storage): return =
+  let store = ensure_not_migrated(store) in
+  let store = List.fold transfer_item params store in
+  (([] : operation list), store)
+
+
+// -----------------------------------------------------------------
+// Balance of
+// -----------------------------------------------------------------
+
+[@inline]
+let validate_token_type (token_id : token_id): token_id =
+  if (token_id = unfrozen_token_id || token_id = frozen_token_id) then
+    token_id
+  else (failwith ("FA2_TOKEN_UNDEFINED"): token_id)
+
+let balance_of (params, store : balance_request_params * storage): return =
+  let check_one (req : balance_request_item): balance_response_item =
+    let valid_token_id = validate_token_type(req.token_id) in
+    match Big_map.find_opt (req.owner, valid_token_id) store.ledger with
+      Some bal -> { request = req; balance = bal}
+    | None -> { request = req; balance = 0n}
+  in
+  let result = List.map check_one params.requests in
+  let transfer_operation = Tezos.transaction result 0mutez params.callback
+  in (([transfer_operation] : operation list), store)
+
+// -----------------------------------------------------------------
+// Token metadata registry entrypoint
+// -----------------------------------------------------------------
+
+let token_metadata_registry (params, store : token_metadata_registry_param * storage):return =
+  let store = ensure_not_migrated (store) in
+  let transfer_operation = Tezos.transaction store.token_address 0mutez params
+  in (([transfer_operation] : operation list), store)
+
+
+// -----------------------------------------------------------------
+// Update operators entrypoint
+// -----------------------------------------------------------------
+[@inline]
+let validate_operator_token (token_id : token_id): token_id =
+  if (token_id = unfrozen_token_id) then
+    token_id
+  else if (token_id = frozen_token_id) then
+    (failwith ("OPERATION_PROHIBITED"): token_id)
+  else
+    (failwith ("FA2_TOKEN_UNDEFINED"): token_id)
+
+let update_one (store, param: storage * update_operator): storage =
+  let (operator_update, operator_param) =
+    match param with
+      Add_operator p -> (Some unit, p)
+    | Remove_operator p -> ((None : unit option), p)
+  in
+  let valid_token_id = validate_operator_token (operator_param.token_id) in
+  if (sender = operator_param.owner) then
+    let key: operator = { owner = operator_param.owner; operator = operator_param.operator} in
+    let updated_operators = Big_map.update key operator_update store.operators
+    in  { store with
+          operators = updated_operators
+        }
+  else (failwith("NOT_OWNER"): storage)
+
+let update_operators (params, store : update_operators_param * storage):return =
+  let store = ensure_not_migrated (store) in
+  let store = List.fold update_one params store in
+  (([] : operation list), store)
+
--- a/src/BaseDAO/ShareTest/Common.hs
+++ b/src/BaseDAO/ShareTest/Common.hs
@@ -3,8 +3,15 @@
 
 {-# LANGUAGE NumericUnderscores #-}
 
-module Test.Common
-  ( checkTokenBalance
+-- | Contain common functions and types that are used
+-- in the shared tests.
+module BaseDAO.ShareTest.Common
+  ( OriginateFn
+  , IsLorentz
+  , expectFailed
+
+  , mkFA2View
+  , checkTokenBalance
   , withOriginated
   , originateTrivialDao
   , originateBaseDaoWithConfig
@@ -14,13 +21,14 @@ module Test.Common
   , addDataToSign
   , permitProtect
   , sendXtz
+  , expectMigrated
   ) where
 
 import Universum
 
 import qualified Data.Map as M
 import qualified Data.Set as S
-import Lorentz
+import Lorentz hiding ((>>))
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Lorentz.Test (contractConsumer)
 import Morley.Nettest
@@ -31,6 +39,19 @@ import qualified Lorentz.Contracts.BaseDAO as DAO
 import Lorentz.Contracts.BaseDAO.Types
 import qualified Lorentz.Contracts.BaseDAO.Types as DAO
 import qualified Lorentz.Contracts.TrivialDAO as DAO
+
+type OriginateFn param m = m ((Address, Address), (Address, Address), TAddress param, Address)
+
+-- TODO: Since Ligo contract has different error messages (string type)
+-- we need this to be able to catch error message properly.
+type IsLorentz = Bool
+
+-- | Used for check error from ligo contract
+expectFailed
+ :: forall caps base m
+  . (MonadNettest caps base m)
+  => Address -> MText -> m () -> m ()
+expectFailed addr val = flip expectFailure (NettestFailedWith addr val)
 
 -- | Function that originates the contract and also make a bunch of
 -- address (the `addrCount` arg determines the count) for use within
@@ -191,3 +212,8 @@ sendXtz addr = do
         , tdParameter = ()
         }
   transfer transferData
+
+expectMigrated
+  :: (MonadNettest caps base m)
+  => Address -> m a -> m ()
+expectMigrated addr = expectCustomError #mIGRATED addr
--- a/baseDAO.cabal
+++ b/baseDAO.cabal
@@ -23,6 +23,8 @@ source-repository head
 library
   exposed-modules:
       BaseDAO.CLI
+      BaseDAO.ShareTest.Common
+      BaseDAO.ShareTest.FA2
       Lorentz.Contracts.BaseDAO
       Lorentz.Contracts.BaseDAO.Doc
       Lorentz.Contracts.BaseDAO.Management
@@ -51,6 +53,7 @@ library
   build-depends:
       aeson-pretty
     , base >=4.7 && <5
+    , cleveland
     , containers
     , lorentz
     , morley
--- a/ligo/src/management.mligo
+++ b/ligo/src/management.mligo
@@ -5,6 +5,23 @@
 
 #include "types.mligo"
 
+
+// -----------------------------------------------------------------
+// Shared Function
+// -----------------------------------------------------------------
+#if !MANAGEMENT_H
+#define MANAGEMENT_H
+
+[@inline]
+let ensure_not_migrated (storage : storage): storage =
+  match storage.migration_status with
+      Not_in_migration -> storage
+    | MigratingTo (p) -> storage
+    | MigratedTo (p) -> (failwith ("MIGRATED") : storage)
+
+#endif  // MANAGEMENT_H included
+
+
 let transfer_ownership
     (param, store : transfer_ownership_param * storage) : return =
   not_implemented("transfer_ownership")
--- a/ligo/src/token.mligo
+++ b/ligo/src/token.mligo
@@ -4,9 +4,14 @@
 // Corresponds to Token.hs module
 
 #include "types.mligo"
+#include "token/fa2.mligo"
 
 let call_fa2(param, store : fa2_parameter * storage) : return =
-  not_implemented("call_fa2")
+  match param with
+    Transfer (p) -> transfer (p, store)
+  | Balance_of (p) -> balance_of(p, store)
+  | Token_metadata_registry (p) -> token_metadata_registry(p, store)
+  | Update_operators (p) -> update_operators(p, store)
 
 let burn(param, store : burn_param * storage) : return =
   not_implemented("burn")
--- a/ligo/test/Ligo/BaseDAO/Helper.hs
+++ b/ligo/test/Ligo/BaseDAO/Helper.hs
@@ -1,0 +1,75 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{-# LANGUAGE RebindableSyntax #-}
+{-# OPTIONS_GHC -Wno-unused-do-bind #-}
+
+module Ligo.BaseDAO.Helper
+  ( mkFullStorage
+  ) where
+
+import Universum ((*))
+import Lorentz
+import Ligo.BaseDAO.Types
+
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+
+defaultConfig :: Config
+defaultConfig  = Config
+  { cUnfrozenTokenMetadata = FA2.TokenMetadata
+      { FA2.tmTokenId = unfrozenTokenId
+      , FA2.tmSymbol = [mt|unfrozen_token|]
+      , FA2.tmName = [mt|Unfrozen Token|]
+      , FA2.tmDecimals = 8
+      , FA2.tmExtras = mempty
+      }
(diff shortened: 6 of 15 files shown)


rinn7e:
@pasqu4le I have added these 4 entrypoints:
  - transfer

pasqu4le:
> Do we plan to re-use the tests from Lorentz BaseDAO?

If possible yes, as it would introduce less duplication and be more robust.

However we might come into trouble in which case we'd have to re-implement, particularly if the mirrored types in `Ligo.BaseDAO.Types` diverge so much that makes this impossible.

I think you have figured out a way to try to solve the problem above, if possible let's go with it.

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+            )
+        in (from_, updated_store)
+      else if tx.token_id = frozen_token_id then
+        (failwith ("FROZEN_TOKEN_NOT_TRANSFERABLE"): address * storage)
+      else
+        (failwith ("FA2_TOKEN_UNDEFINED") : address * storage)
+    )
+
+let transfer_item (store, ti : storage * transfer_item): storage =
+  let result = List.fold transfer_one ti.txs (ti.from_, store)
```
pasqu4le, line 86: Folding over the pair with the `from_` value doesn't make much sense here in terms of cost, because it's a value that never changes.

We should instead make it a parameter of `transfer_one`, or (considering my previous comment) make `transfer_one` a closure here.

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+  in  { store with
+        ledger = Big_map.update (to_, token_id) (Some new_bal) store.ledger
+      }
+
+
+// -----------------------------------------------------------------
+// Transfer
+// -----------------------------------------------------------------
+
+// TODO: how to add to 'where'
```
pasqu4le, line 65: I don't understand this comment: can you explain?

pasqu4le (changes requested):

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+  match bal_o with
+    Some bal ->
+      let result: balance_response_item = { request = req; balance = bal}
+      in result :: res_list
+  | None ->
+      let zero_bal: balance_response_item = { request = req; balance = 0n}
+      in zero_bal :: res_list
+
+let balance_of (params, store : balance_request_params * storage): return =
+  let result = List.fold (check_one (store)) params.requests ([]: balance_response_item list) in
```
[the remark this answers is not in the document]
rinn7e: I see, I didn't know this closures (I was checking CameLigo only). Let me try to use it to see if we can avoid the various currying.

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+
+// -----------------------------------------------------------------
+// Helper
+// -----------------------------------------------------------------
+
+let ensure_not_migrated (storage : storage): bool =
+  match storage.migration_status with
+      Not_in_migration -> true
+    | MigratingTo (p) -> true
+    | MigratedTo (p) -> (failwith ("MIGRATED") : bool)
```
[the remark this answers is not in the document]
rinn7e: Yeah, I was planning to use `unit` type as well (It supposed to be `unit` but I forgot to update, and it applies to all various failwith functions).

I did it this way since it supposed to resembles a guard check, for example in stablecoin pascaligo (different function but serve similar purpose):

```
function ensure_not_paused
  ( const store : storage
  ) : unit is
  fail_on
    ( store.paused
    , "CONTRACT_PAUSED"
    )

function transfer
  ( const param: transfer_param
  ; const store: storage
  ; const full_param: closed_parameter
  ) : entrypoint is block
{ ensure_not_paused(store)
....
}
```

In this usage, we don't really care about the return type.

I'm trying to achieve something similar in cameligo, but seems not possible, so that's why I use various `cond_` which means we don't care the about the return value.

So what do you think about this, Pasquale? Should I update as you suggested earlier or keep it as unit.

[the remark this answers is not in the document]
pasqu4le: I still think we should use the `storage` version.

First of all (not that important) because it seems more in `cameLIGO`'s style to only return useful types, given that it does not have a clean way to ignore the result of calling the function.

Secondly, and more importantly, because it's more efficient: with the `bool` or `unit` version there is a `PUSH`/`UNIT` instruction that does not get optimized away and so increases the cost (however little that may be).
With the `storage` version it doesn't happen because the return is already just below in the stack.
Copied from the resulting code, it's:
```
IF_LEFT { DROP ; UNIT } { IF_LEFT { DROP ; UNIT } { DROP ; PUSH string "MIGRATED" ; FAILWITH } } } ;
```
vs
```
IF_LEFT { DROP } { IF_LEFT { DROP } { DROP 2 ; PUSH string "MIGRATED" ; FAILWITH } } } ;
```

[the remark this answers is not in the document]
pasqu4le: I have two more requests about this function:

1. define it as `[@inline]`. Just as the `ensure_not_paused` your mentioned is, this will most likely be more efficient.
2. move it to `management.mligo` and encase it between this:
```
#if !MANAGEMENT_H
#define MANAGEMENT_H

<definition here>

#endif  // MANAGEMENT_H included
```
there. For an explanation why see the note about shared functions here: #100

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
pasqu4le: @rinn7e have you tried re-implementing this?

pasqu4le:
Please remember to rebase now that #81 has been merged :)

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
rinn7e: Not yet, I was focusing on make the tests work. I'll make this changes as well as other optimization tomorrow.

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+      let updated_store =
+        match Michelson.is_nat (new_bal_int) with
+          Some new_bal ->
+            { store with
+              ledger = Big_map.update (from_, token_id) (Some new_bal) store.ledger
+            }
+        | None ->
+            // TODO: How to fail with some value
+            // (failwith ("FA2_INSUFFICIENT_BALANCE", (amt, bal)) : storage)
+            (failwith ("FA2_INSUFFICIENT_BALANCE") : storage)
```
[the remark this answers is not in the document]
rinn7e: Seems like we can `failwith` something else beside a string by using something similar to this: https://github.com/tqtezos/stablecoin/blob/master/ligo/stablecoin/actions.ligo#L717-L726

Do you think it is worth doing or just leave it as `string` for now?

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
rinn7e: I've made all the changes according to your review now, you can check this.

Next, I'm going to work on my other assigned issues.

rinn7e:
@gromakovsky @Martoon-00 Hope to get your reviews as well.

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+// Balance of
+// -----------------------------------------------------------------
+
+let validate_token_type (token_id : token_id): bool =
+  if (token_id = unfrozen_token_id) then
+    true
+  else if (token_id = frozen_token_id) then
+    true
+  else
+    (failwith ("FA2_TOKEN_UNDEFINED"): bool)
```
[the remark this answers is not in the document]
pasqu4le: I see you have changed the return type as suggested, but not the refactoring.

Is there a reason to prefer the second `if` to the `||`?

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,176 @@
...
+let check_one (store : storage) (res_list, req : balance_response_item list * balance_request_item): balance_response_item list =
+  let cond_ = validate_token_type(req.token_id) in
+  let bal_o = Big_map.find_opt (req.owner, req.token_id) store.ledger in
+  match bal_o with
+    Some bal ->
+      let result: balance_response_item = { request = req; balance = bal}
+      in result :: res_list
+  | None ->
+      let zero_bal: balance_response_item = { request = req; balance = 0n}
+      in zero_bal :: res_list
```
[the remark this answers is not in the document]
pasqu4le: @rinn7e I think you missed this one (in `balance_of`)

ligo/src/token/fa2.mligo
pasqu4le: @rinn7e still wondering what this comment is about :slightly_smiling_face:

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,161 @@
...
+
+// Corresponds to Token/FA2.hs module
+
+#include "../types.mligo"
+#include "../management.mligo"
+
+
+let unfrozen_token_id: nat = 0n
+let frozen_token_id: nat = 1n
+// let (|>) x f = f x
```
pasqu4le, line 12: I see you changed the `*_token_id` values, but you left behind this comment, let's remove it.

ligo/src/token/fa2.mligo
rinn7e: I was wondering how to add a `where` clause like in haskell. I guess we can just put the function in `let` as well.

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
pasqu4le: I didn't think about that. I think it is worth doing, but let's not do this here.

I have specified in the specification that the errors are returned as `string` only too, so that needs to be updated as well.

I'll open a separate issue to add this different `failwith` and probably work on it myself in parallel.

ligo/src/token/fa2.mligo
pasqu4le: I see, can you please remove this comment in this case?

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,155 @@
...
+              )
+          in updated_store
+        else if tx.token_id = frozen_token_id then
+          (failwith ("FROZEN_TOKEN_NOT_TRANSFERABLE"): storage)
+        else
+          (failwith ("FA2_TOKEN_UNDEFINED") : storage)
+      )
+  in
+  let result = List.fold transfer_one ti.txs store
+  in result
```
pasqu4le, line 78: A couple of points here: first the minor ones are that `updated_store` and `result` don't need to be declared and that the top-most `else` is not correctly indented.

More importantly there is an opportunity for optimization: by using the first big condition in the inner `if` you can write its block code once, however this comes with the cost of having to re-check the `tx.token_id = frozen_token_id` later on.
We can refactor this with the use of a `valid_token_id` and save that too, I suggest this (including the minor points above):
```suggestion
    if tx.amount = 0n then acc
    else (
      let valid_token_id =
        if tx.token_id = unfrozen_token_id then tx.token_id
        else if tx.token_id = frozen_token_id then (
          if sender = store.admin then tx.token_id
          else (failwith ("FROZEN_TOKEN_NOT_TRANSFERABLE"): token_id)
        ) else (failwith ("FA2_TOKEN_UNDEFINED") : token_id)
      in credit_to
        ( tx.amount, tx.to_, valid_token_id
        , debit_from (tx.amount, valid_from_, valid_token_id, acc)
        )
    )
  in List.fold transfer_one ti.txs store
```

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,155 @@
...
+let balance_of (params, store : balance_request_params * storage): return =
+  let check_one (res_list, req : balance_response_item list * balance_request_item): balance_response_item list =
+    let valid_token_id = validate_token_type(req.token_id) in
+    match Big_map.find_opt (req.owner, valid_token_id) store.ledger with
+      Some bal ->
+        { request = req; balance = bal} :: res_list
+    | None ->
+        { request = req; balance = 0n} :: res_list
+  in
+  let result = List.fold check_one params.requests ([]: balance_response_item list) in
```
pasqu4le, line 105: No need to rebuild the list manually, we can keep the logic using `List.map` instead of `List.fold`:
```suggestion
  let check_one (req : balance_request_item): balance_response_item =
    let valid_token_id = validate_token_type(req.token_id) in
    match Big_map.find_opt (req.owner, valid_token_id) store.ledger with
      Some bal -> { request = req; balance = bal}
    | None -> { request = req; balance = 0n}
  in
  let result = List.map check_one params.requests in
```
the resulting Michelson code seems to be a little better too.

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,155 @@
...
+
+// -----------------------------------------------------------------
+// Token metadata registry entrypoint
+// -----------------------------------------------------------------
+
+let token_metadata_registry (params, store : token_metadata_registry_param * storage):return =
+  let cond_ = ensure_not_migrated (store) in
+  let transfer_operation = Tezos.transaction store.token_address 0mutez params
+  in
+    (([transfer_operation] : operation list), store)
```
pasqu4le, line 117:
```suggestion
  let valid_store = ensure_not_migrated (store) in
  let transfer_operation = Tezos.transaction valid_store.token_address 0mutez params
  in (([transfer_operation] : operation list), valid_store)
```

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,155 @@
...
+    let key: operator = { owner = operator_param.owner; operator = operator_param.operator} in
+    let updated_operators = Big_map.update key operator_update store.operators
+    in  { store with
+          operators = updated_operators
+        }
+  else (failwith("NOT_OWNER"): storage)
+
+let update_operators (params, store : update_operators_param * storage):return =
+  let cond_ = ensure_not_migrated (store) in
+  let updated_store = List.fold update_one params store in
```
pasqu4le, line 149:
```suggestion
  let valid_store = ensure_not_migrated (store) in
  let updated_store = List.fold update_one params valid_store in
```

pasqu4le (commented):
A couple more comments in the ligo side of things

ligo/src/token/fa2.mligo
rinn7e: That makes sense, I've made this change as well as other changes as well.

src/BaseDAO/ShareTest/Common.hs
```diff
@@ -0,0 +1,215 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{-# LANGUAGE NumericUnderscores #-}
+
+module BaseDAO.ShareTest.Common
```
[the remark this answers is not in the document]
rinn7e: I've added the docs as well as making other changes.

pasqu4le (approved):
LGTM now

gromakovsky:
From quick glance it looks ok. I noticed that you implemented legacy FA2 with `token_metadata_registry`. Probably it's done this way because that's how it's done in Lorentz (and there it's done this way for historical reasons). I think it's fine to have legacy implementation in the first version since we are in hurry, but let's update it when we have time. I've created https://github.com/tqtezos/baseDAO/issues/107 for that.

ligo/src/token/fa2.mligo
Martoon-00: Btw, this practice of calling store variables different each time does not seem good to me.
1. I have to invent a new name for storage variable each time I call a function.
2. Now I have to be **extra** careful when updating the code, because it is too easy to use an outdated variable, and LIGO won't warn me if the most recent storage variable remained unused.

In proposals logic I have methods which perform up to 3 checks and then 2-3 updates, and with regard to this practice those issues seem quite severe.

Maybe we could just call the variable as `store`  everywhere? LIGO seems not to care about a variable with the same name being defined multiple times, and particularly for storage variable this seems to be much safer. Though I don't know whether will it work as expected.

At the very least maybe use `store1`, `store2` e.t.c.? This way (1) is not a problem anymore, and issues with (2) can be relatively easy detected.

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
Martoon-00: Currently only one function from the entire module is enclosed into `MANAGEMENT_H`, probably that's not what you meant? Like, it looks like it's worth deduplicating either the entire module, or nothing.

ligo/src/token/fa2.mligo
Martoon-00: I agree that making verification return `valid_store` and let further functions work with validated storage is generally a good practice, but to me in makes sense only in a language where `store` and `valid_store` would have different types, while here the compiler does not prevent us from abusing this pattern accidentially. After all, we only gain many different ways to implement the same thing (pick/ignore validated store, how to call the returned variables, how to chain several checks, e.t.c) and more points for mutual nitpicking :smile:

pasqu4le: @Martoon-00 you make good points, I though about those too yesterday and didn't come up with a way to improve this substantially.

Regarding checks the good think is that these can often be made all at once and have for example something like:
```haskell
  let valid_store = check1 (check2 (check3 store)) in
```

I also tried to make a function composition, but since this isn't actually `OCaml` it seems to be only doable with currying which impacts on produced Michelson, otherwise it would have looked like this:
```haskell
  let updated_store =
    store
    |> ensure_not_migrated
    |> List.fold update_one params
```

I think it's fine to call the variable `store` anywhere since replacement seems to be working fine, but we should be extra careful this does not produce wrong code (tests should catch that tho).

@rinn7e if you agree as well could you make these changes?

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
pasqu4le: Yeah that's what I meant because this should be left to #100 to solve (and there are no problems now that the endpoint functions are not defined), there I more clearly specified that we need to end up with a complete solution:
> We can either use the existing source file and put all definitions in the block or make a separate one just for the 2 checking functions instead. Your pick.

ligo/src/token/fa2.mligo
Martoon-00: Yeah, pretty sad that LIGO is limited here, without all the conveniences we used to in Haskell it is actually difficult to write on a language with immutable variables.

rinn7e: Yeah, I'm good with just calling the it `store` everywhere as well.

rinn7e: I've pushed the changes now, let me know if this is good enough.

Martoon-00: Thanks!

ligo/src/token/fa2.mligo
```diff
@@ -74,9 +74,9 @@ let transfer_item (store, ti : storage * transfer_item): storage =
   in List.fold transfer_one ti.txs store
```
pasqu4le, line 74: In light of using always the `store` name, I think maybe we should consider renaming the `acc` in `transfer_item` too?

This seems to work just as fine: 
```ocaml
let transfer_item (store, ti : storage * transfer_item): storage =
    let valid_from_ = check_sender (ti.from_, store) in
    if tx.amount = 0n then store
    else (
      let valid_token_id =
        if tx.token_id = unfrozen_token_id then tx.token_id
        else if tx.token_id = frozen_token_id then (
          if sender = store.admin then tx.token_id
          else (failwith ("FROZEN_TOKEN_NOT_TRANSFERABLE"): token_id)
        ) else (failwith ("FA2_TOKEN_UNDEFINED") : token_id)
      in credit_to
        ( tx.amount, tx.to_, valid_token_id
        , debit_from (tx.amount, valid_from_, valid_token_id, store)
        )
    )
  in List.fold transfer_one ti.txs store
```

here we loose the ability to access the outer-most `store`, but in this case it makes no difference (logically).

Checking the resulting Michelson code there is indeed a slight change for the better (because ti doesn't have to fetch `store.admin` from the further away), so I suggest that we either:
- replace `acc` with `store` as in the block above
- replace `store.admin` with `acc.admin`

ligo/src/token/fa2.mligo
```diff
@@ -28,15 +28,15 @@ let check_sender (from_ , store : address * storage): address =
...
+      let store =
         match Michelson.is_nat (bal - amt) with
           Some new_bal ->
             { store with
               ledger = Big_map.update (from_, token_id) (Some new_bal) store.ledger
             }
         | None ->
             (failwith ("FA2_INSUFFICIENT_BALANCE") : storage)
-      in updated_store
+      in store
```
pasqu4le: Just wanted to mention that as @sras found out this can also be written like this:
```ocaml
    Some bal -> (
      match Michelson.is_nat (bal - amt) with
        Some new_bal ->
          { store with
            ledger = Big_map.update (from_, token_id) (Some new_bal) store.ledger
          }
      | None ->
          (failwith ("FA2_INSUFFICIENT_BALANCE") : storage)
    )
```
I don't think it's any better (or worst), I'm fine with it either way

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
Martoon-00: Btw, I also started writing checks so that they return `unit`, and I observe that performance aspect of this seems to be not so critical to me.

Note that we want to return `unit` only for checks, and checks are normally small which means that we always want to inline them (if we don't do this, then checks get extracted into a lambda annotated with storage/config type which is enormously huge, takes about 250 bytes of op size).

And for inlined functions returning `unit` does not add unnecessary `UNIT; DROP`.

Whether is returning `unit` comply with the CameLIGO style (in case if they have a reasonable style with respect to their featureset :thinking:) is also an insteresting question. For instance, if I return non-unit in a check, I have to annotate `failwith` with the returned type, so both approaches seem to be in some sort unpleasant.

Also there is a point about that distinguishing read-only and state-updating methods in their signatures is good, though I don't know whether will we encounter cases where this matters.

So probably I will just follow the style we came up with in here, but if points above make sense to you then next time we can keep in mind that both approaches make sense. Or decide that picking cameLIGO was a mistake :smile:

ligo/src/token/fa2.mligo
```diff
@@ -0,0 +1,144 @@
...
+// Update operators entrypoint
+// -----------------------------------------------------------------
+[@inline]
+let validate_operator_token (token_id : token_id): token_id =
+  if (token_id = unfrozen_token_id) then
+    token_id
+  else if (token_id = frozen_token_id) then
+    (failwith ("OPERATION_PROHIBITED"): token_id)
+  else
+    (failwith ("FA2_TOKEN_UNDEFINED"): token_id)
```
Martoon-00, line 123: Oh by the way, could you rename `assertEq0Or1` method in Lorentz version as well?

I think it would be nice to have as exact match between two versions of the contract as possible, and your name looks much better.

If we don't do this now, probably that name will remain there forever )

ligo/src/token/fa2.mligo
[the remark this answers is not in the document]
pasqu4le: I'm fairly sure I had checked the function as inlined too and it did have an additional `UNIT` instruction in it, but probably it was because it contained a `match` in cameLIGO and it wasn't able to optimize that away.

The improvement is very small either way, it only saves _maybe_ a `DROP` and some `UNIT`s.

I agree on the rest, after this is done we should consider weather we can make better/different choices next.

Martoon-00 (approved):
I reviewed superficially (since @pasqu4le made a very thorough review), the overall approach looks good to me.

ligo/src/token/fa2.mligo
rinn7e: Yeah, make sense. I've updated this now.

PR #110, opened 2021-01-13
[#57] Add `ligo-test` in CI

### LINKED ISSUE #57: Initialize LIGO version development

rinn7e:
## Description

This PR updates our CI to run tests for the LIGO contract.
Currently we run them only on emulator because the fail on network due to oversized operation. An issue about that will be created soon.

## Related issue(s)

Resolves #57

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -39,7 +39,14 @@ steps:
     commands:
     - nix run -f ci.nix stack2cabal pkgs.diffutils -c ./scripts/ci/validate-cabal-files.sh
 
-  - label: build
+  - label: build-ligo
+    if: *not_scheduled
+    commands:
+    - nix-build ci.nix -A build-ligo -o baseDAO.tz
+    artifact_paths:
+      - baseDAO.tz
+
+  - label: build-haskell
     if: *not_scheduled
     commands:
     - nix-build ci.nix -A all-components
@@ -61,11 +68,6 @@ steps:
     - nix run -f ci.nix tezos-client -c
       ./result/bin/baseDAO-test --nettest-run-network
       -A localhost -P 8735
-
-  - label: ligo
-    if: *not_scheduled
-    commands:
-    - nix-build ci.nix -A build-ligo
 
   - label: weeder
     if: *not_scheduled
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -79,13 +79,14 @@ steps:
       ./result/bin/baseDAO-test --nettest-no-run-network
       -A localhost -P 8735
 
-  - label: weeder
-    if: *not_scheduled
-    commands:
-    - nix-build ci.nix -A weeder-script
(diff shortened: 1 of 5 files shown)


rinn7e:
So I've spent quite sometimes on this, but couldn't make it work. I was able to upload and download the artifact, but `test-ligo` seems to not be able to detect the downloaded artifact (Probably wrong path).

@pasqu4le Hope to get your advice on this.

zhenyavinogradov:
I've fixed building the tests in CI, now there's just a problem with invocation, it says it needs different arguments https://buildkite.com/serokell/basedao/builds/845#38523cf0-e830-488a-9f58-456fe41e69a7

src/BaseDAO/ShareTest/FA2.hs
```diff
@@ -163,7 +163,7 @@ updatingOperatorScenario originateFn = do
   callFrom (AddressResolved owner2) dao (Call @"Transfer") transferParams
     & expectCustomError_ #fA2_NOT_OPERATOR
   callFrom (AddressResolved owner1) dao (Call @"Update_operators") ([FA2.AddOperator notOwnerParams])
-    & expectCustomError_ #nOT_OWNER
+    & expectCustomError_ #fA2_NOT_OPERATOR
```
pasqu4le, line 166: Not sure why this was changed or how it relates to the problems on CI, but it seems the tests are now working and this is the only thing failing (correctly, as it should indeed be `#nOT_OWNER` AFAIK).
```suggestion
    & expectCustomError_ #nOT_OWNER
```

pasqu4le (approved):

Issue #91, opened 2020-12-30
Revise and restore checks applying to most entrypoints


gromakovsky:
# Clarification and motivation

There are 2 checks we want to make in most of entrypoints: `ensureZeroTransfer` and `ensureNotMigrated`.
In #85 we temporarily disabled `ensureZeroTransfer` because it's tricky to do it for everything besides `CallCustom` without copy-pasting it everywhere. And we don't need this check in `CallCustom` because in some custom entrypoints we want to permit XTZ transfers.

Apart from that `ensureNotMigrated` seems to be copy-pasted in many places, but we can do it only once and it should be more efficient.

# Acceptance criteria

1. `ensureZeroTransfer` is restored and called for all entrypoints besides `CallCustom`. Probably we need to restructure the parameter type and then do just one `IF_LEFT {} {ensureZeroTransfer}`.
2. `ensureNotMigrated` appears only once (or maybe twice or so) in the contract code. Perhaps a similar restructuring should be made to achieve that.

Martoon-00:
Interesting, how will we make sure that the user _does_ call `ensureZeroTransfer` and `ensureNotMigrated` in his custom entrypoints unless they are not necessary. For instance, in TreasuryDAO's `%default` we probably want to call `ensureNotMigrated`, but it is too easy to forget about.

gromakovsky:
Currently I don't see a way other than saying "please be careful".

gromakovsky:
I think we can also expand the configuration to include things like `permitXtzTransfer :: Bool`. E. g. for TreasuryDAO it's probably harmless to permit non-zero amount in any entrypoint because they won't be locked.

Martoon-00:
:pogchamp: That would be the simplest way.

gromakovsky:
One more thing to consider: currently we want to prohibit non-zero XTZ transfers in all entrypoints, including `propose`, however, sometimes it can be useful to transfer non-zero amount of XTZ in `propose`, for example it can be useful for #67 where we want to have fees in Tez.

pasqu4le:
Update on this issue's status.

The current contract (LIGO-only version) no longer has the migration functionality (see #173), which invalidates/cover point 2.

Regarding the n.1 acceptance criteria the contract already makes the no-xtz check only once as the overall parameter is divided in entrypoints that do/don't accept them.

The only point left is to decide on what entrypoints we want to keep the no-xtz check and on which we don't, given some new discussions on the matter I opened #256 for it.

[...]

Issue #106, opened 2021-01-12
Extend the use of LIGO failwith to pairs


pasqu4le:
# Clarification and motivation

In the LIGO version of the contract we always use `failwith` with a single string, however we'd like to match the Lorentz counterpart in the use of `pair`s for some errors.

The problem is that the standard `failwith` from LIGO only supports a few input types, but not `pair`s, however we can define our own specialized version of `failwith` using embedded Michelson (see [here](https://github.com/tqtezos/stablecoin/blob/master/ligo/stablecoin/actions.ligo#L717-L726)).

We should likely add this to `types.mligo` because this will be used by many modules and the `types` is safely imported by all of them.
Note: It looks like we'll need to declare it for a specific pair type, so we might need to have more than one version of this.

Additionally we'll need to update:
- the `specification-ligo.md` section on `Errors` because it says it always fails with a `string`
  note: we can copy the line to change from the Lorentz specification
- the shared tests introduced in #99, which needed to make a differentiation with the `IsLorentz` because of the different errors.

# Acceptance criteria

- `failwith` equivalent function(s) to throw as many `pair` errors as in the Lorentz version are added to LIGO's code.
- tests and documentation are updated accordingly.

rinn7e:
> We should likely add this to `types.mligo` because this will be used by many modules and the types is safely imported by all of them.

With our use case, it not really possible to share the fail function, since we annotate `failwith` with different types. e.q.

```
(failwith ("FROZEN_TOKEN_NOT_TRANSFERABLE"): token_id)
(failwith("NOT_OWNER"): storage)
```

In this two fail cases, we will need this function:

```ocaml
  let fail_unit: (string * unit -> _) =
      [%Michelson ({| { FAILWITH } |} : string * unit -> _)] in
```

But since they have different types, we can't use the shared function, and have to use the michelson code directly where it is needed like this:

```
([%Michelson ({| { FAILWITH } |} : string * unit -> token_id)]
  ("FROZEN_TOKEN_NOT_TRANSFERABLE", ()) : token_id)

([%Michelson ({| { FAILWITH } |} : string * unit -> storage)]
  ("FROZEN_TOKEN_NOT_TRANSFERABLE", ()) : storage)
```

There is really no ideal way to make this sharable. @pasqu4le What's your view on this, Pasquale?

pasqu4le:
@rinn7e I didn't think about the need for a fixed return type.

[...]

PR #104, opened 2021-01-12
[#102] Implement Token-related endpoints for LIGO version

rinn7e:
## Description

Problem: Token-related entrypoints for LIGO is not implemented yet.

Solution: Implemented token-related entrypoints for LIGO and add their
tests.

## Related issue(s)

Resolves #102 

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

[//]: # (Add more docs here if you have them in the repository)
- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](../tree/master/README.md)
    - Haddock

#### Stylistic guide (mandatory)

- [x] My commits comply with [the following policy](https://www.notion.so/serokell/Commit-and-PR-policy-4cf98e1a910a415d86b5f2491d9af1af).
- [x] My code complies with the [style guide](../tree/master/docs/code-style.md).

Issue #102, opened 2021-01-11
Implement the token entrypoints for the LIGO version of the contract


pasqu4le:
# Clarification and motivation

We need to implement the "token" entrypoints for the LIGO version of the contract.

There is already a `token.mligo` source file with the necessary entrypoints (without any implementations).

Note that this does not need to implement the FA2 part of the endpoints as those will be handled in #95, so namely it's about: `burn`, `mint` and `transfer_contract_tokens`

These should likely not differ much from their Lorentz counterpart, but in case something seems odd see the differences in the `Configuration`  paragraph between `specification-ligo.md` and `specification.md`.

Last note: tests will need to be re-implemented in Haskell due to the differences between the LIGO and Lorentz version of the contract that are reflected in the types.

# Acceptance criteria

All non-FA2 entrypoints defined in `token.mligo` have been implemented and there are tests covering them.

PR #111, opened 2021-01-13
[#100] Management LIGO entrypoints

### LINKED ISSUE #100: Implement management entrypoints for the LIGO version of the contract

We need to implement the "management" entrypoints for the LIGO version of the contract.

There is already a `management.mligo` source file with the necessary entrypoints (without any implementations).

Most entrypoints of this group (except `call_custom`) should roughly match their Lorentz counterpart, see the `specification-ligo.md` for directions in what should be their behavior.

`call_custom` is the exception because it replaces the functionality of allowing for custom endpoints, which is not possible in LIGO as it is in Lorentz for the lack of compile-time options.

For more information see the differences in the `Configuration`  paragraph between `specification-ligo.md` and `specification.md`.
The LIGO specification contains an explanation of this custom entrypoint implementation under `Custom entrypoints`.
It basically needs to `unpack` a `lambda` from storage and the "custom entrypoint" parameter from the entrypoint input `bytes`.

A note about shared functions:
this module will need an `ensure_not_migrated` and `authorize_admin` to be defined and used as well as as exported.
Because of how LIGO imports work we'd better make sure these are not included in the resulting contract multiple times (see [this discussion](https://github.com/tqtezos/baseDAO/pull/81#discussion_r549880585)) by using the trick with the C preprocessor:
```cameLIGO
#if !MANAGEMENT_H
#define MANAGEMENT_H

<definitions here>

#endif  // MANAGEMENT_H included
```
to allow this module to be imported by multiple ones safely.
We can either use the existing source file and put all definitions in the block or make a separate one just for the 2 checking functions instead. Your pick.

One more thing: please check out [this discussion](https://github.com/tqtezos/baseDAO/pull/99/files?file-filters%5B%5D=.cabal&file-filters%5B%5D=.hs&file-filters%5B%5D=.json&file-filters%5B%5D=.license&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.mligo&file-filters%5B%5D=.yaml#r553943207) about the type of `ensure_not_migrated` which has also been implemented in an open PR (at the time of writing) so to avoid repeating talking about it (and also implementing it, as you can just copy it from there).

Last note: tests will need to be re-implemented in Haskell due to the differences between the LIGO and Lorentz version of the contract (namely the `call_custom` above).

# Acceptance criteria

All entrypoints defined in `management.ligo` have been implemented and there are tests covering them.

sras:
## Description

1. Implements Management entrypoints for the LIGO contract along with test (mostly copied from tests for the Haskell version)
2. Refactor parameter to avoid multiple calls to `ensure_admin` and `ensure_not_migrated` functions.
3. Update specifications.

## Related issue(s)

Resolves #100 and addresses #91

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/src/common.mligo
+++ b/ligo/src/common.mligo
@@ -1,0 +1,15 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !COMMON_H
+#define COMMON_H
+
+#include "types.mligo"
+
+let authorize_admin (store : storage): storage =
+  if sender = store.admin then
+    store
+  else
+    (failwith("NOT_ADMIN"): storage)
+
+#endif
--- a/ligo/src/management.mligo
+++ b/ligo/src/management.mligo
@@ -4,13 +4,7 @@
 // Corresponds to Management.hs module
 
 #include "types.mligo"
-
-
-// -----------------------------------------------------------------
-// Shared Function
-// -----------------------------------------------------------------
-#if !MANAGEMENT_H
-#define MANAGEMENT_H
+#include "common.mligo"
 
 [@inline]
 let ensure_not_migrated (storage : storage): storage =
@@ -19,21 +13,64 @@ let ensure_not_migrated (storage : storage): storage =
     | MigratingTo (p) -> storage
     | MigratedTo (p) -> (failwith ("MIGRATED") : storage)
 
-#endif  // MANAGEMENT_H included
-
-
+(*
+ * Auth checks for admin and store the address in parameter to the
+ * 'pending_owner' field in storage.
+ *)
 let transfer_ownership
     (param, store : transfer_ownership_param * storage) : return =
-  not_implemented("transfer_ownership")
+      let store = authorize_admin(store) in
+        (([] : operation list), { store with pending_owner = param ; })
 
+(*
+ * Auth check for pending admin and copies the value in 'pending_owner' field
+ * to 'admin' field. The 'pending_owner' field is left with the address of the
+ * new admin.
+ *)
 let accept_ownership(param, store : unit * storage) : return =
-  not_implemented("transfer_ownership")
+  if store.pending_owner = Tezos.sender
+    then (([] : operation list), { store with admin = store.pending_owner })
+    else (failwith("NOT_PENDING_ADMIN") : return)
 
+(*
+ * Auth check for admin and sets the migration status to 'MigratingTo' using
+ * the address from parameter, overwriting any address from any previous
+ * 'Migrate' calls.
+ *)
 let migrate(param, store : migrate_param * storage) : return =
-  not_implemented("migrate")
+  let store = authorize_admin(store) in
+    (([] : operation list), { store with migration_status = MigratingTo (param) })
 
+(*
+ * Auth check for new contract and sets the migration_status to 'Migrated', using
+ * the address from the earlier value in the migration_status field.
+ *)
 let confirm_migration(param, store : unit * storage) : return =
-  not_implemented("confirm_migration")
+  match store.migration_status with
+    Not_in_migration -> (failwith("NOT_MIGRATING") : return)
+  | MigratingTo (new_addr) -> if new_addr = Tezos.sender
+      then (([] : operation list), { store with migration_status = MigratedTo (new_addr) })
+      else (failwith("NOT_MIGRATION_TARGET") : return)
+  | MigratedTo (new_addr)  -> (failwith("MIGRATED") : return)
 
-let call_custom(param, store : custom_ep_param * storage) : return =
-  not_implemented("call_custom")
+(*
+ * Call a custom entrypoint. Looks up the packed code for entrypoint using the entrypoint
+ * name in the parameter. Then unpacks the code to a lambda of type 'bytes -> return'.
+ * Then just pass the bytes from parameter to this lambda return the resulting value.
+ *
+ * NO AUTH CHECKS ARE DONE.
+ *)
+let call_custom(param, full_storage : custom_ep_param * full_storage) : return_with_full_storage =
+  let ep_name = param.0 in
+  let config = full_storage.1 in
+  let storage = full_storage.0 in
+  let packed_param = param.1 in
+
+  match Map.find_opt ep_name config.custom_entrypoints with
+    Some (ep_code) ->
+      begin
+        match ((Bytes.unpack ep_code) : ((bytes * full_storage) -> return_with_full_storage) option) with
+          Some lambda -> lambda (packed_param, full_storage)
+        | None -> (failwith("UNPACKING_FAILED") : return_with_full_storage)
+      end
+  | None -> (failwith("ENTRYPOINT_NOT_FOUND") : return_with_full_storage)
--- a/ligo/test/Ligo/BaseDAO/Types.hs
+++ b/ligo/test/Ligo/BaseDAO/Types.hs
@@ -64,10 +64,14 @@ data ProposalL = ProposalL
 
 type CallCustomParam = (MText, ByteString)
 
-data Parameter
+-- NOTE: Constructors of the parameter types should remain sorted for the
+--'ligoLayout' custom derivation to work.
+--
+-- TODO: This above will be no longer the case once the following issue is fixed.
+-- https://gitlab.com/morley-framework/morley/-/issues/527
+data ForbidXTZParam
   = Accept_ownership ()
   | Burn BurnParam
-  | CallCustom CallCustomParam
   | Call_FA2 FA2.Parameter
   | Confirm_migration ()
   | Drop_proposal ProposalKeyL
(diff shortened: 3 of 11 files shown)


ligo/src/common.mligo
```diff
@@ -0,0 +1,19 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !COMMON_H
+#define COMMON_H
+
+#include "types.mligo"
+
+let authorize_admin (store: storage): bool =
+    store.admin = Tezos.sender
```
pasqu4le, line 10: While this should be here (as it will be used by `token` entrypoints too) I suggest we modify it to match the version one @rinn7e made for it there, in #104 :
```suggestion
let authorize_admin (store : storage): storage =
  if sender = store.admin then
    store
  else
    (failwith("NOT_ADMIN"): storage)
```
because:
1. we are trying to make this check on `storage` data all have it returned (see [discussion here]())
2. it doesn't make sense to only make the boolean check and write the `failwith` elsewhere, because the latter is always going to be the same.

Also, we should check if this should be `[@inline]`d as well, it would surely make sense in it's current implementation, but probably still will in the suggested one.

P.s. @rinn7e since I have already tagged you, please review this PR as well when you can.

ligo/src/management.mligo
```diff
@@ -4,36 +4,74 @@
...
+ * the address from the earlier value in the migration_status field.
+ *)
 let confirm_migration(param, store : unit * storage) : return =
-  not_implemented("confirm_migration")
+  match store.migration_status with
+    Not_in_migration -> (failwith("NOT_MIGRATING") : return)
+  | MigratingTo (new_addr) -> if new_addr = Tezos.sender
+      then (([] : operation list), { store with migration_status = MigratedTo (new_addr) })
+      else (failwith("NOT_MIGRATION_TARGET") : return)
+  | MigratedTo (new_addr)  -> (failwith("MIGRATED") : return)
```
pasqu4le, line 53: I guess this clause will never happen now that `ensure_not_migrated` is executed before this, but I suppose that's fine.

ligo/src/management.mligo
```diff
@@ -4,36 +4,74 @@
...
+ * Call a custom entrypoint. Looks up the packed code for entrypoint using the entrypoint
+ * name in the parameter. Then unpacks the code to a lambda of type 'bytes -> return'.
+ * Then just pass the bytes from parameter to this lambda return the resulting value.
+ *
+ * NO AUTH CHECKS ARE DONE.
+ *)
+let call_custom(param, full_storage : custom_ep_param * full_storage) : return =
+  let ep_name = param.0 in
+  let config = full_storage.1 in
+  let storage = full_storage.0 in
```
pasqu4le, line 65: At this implementation it seems that the `full_storage` parameter could be replaced by the two `config` and `store` one (since we already made the un-pairing in `base_DAO.mligo`), however I was actually thinking that the `lambda` could accept the `full_storage` instead of the `storage`, so this would be fine if we did that.

The reason for this change in the lambda is that since this LIGO version is being made with runtime-customization in mind (is more flexible, whereas the Lorentz one is more static but also more efficient) we could have more interesting things happen in these custom entrypoints.
One example that was being considered is to have an entrypoint that can only be run once because during that first run it removes itself from the `map` (to reduce the cost of origination).

What do you think? can we do that?
If so please also update the `Custom entrypoints` part of the specification as well to use `full_storage` instead of `storage`.

ligo/test/Test/Ligo/BaseDAO/Management.hs
```diff
@@ -0,0 +1,269 @@
...
+withOriginated addrCount storageFn tests = do
+  addresses <- mapM (\x -> newAddress $ "address" <> (show x)) [1 ..addrCount]
+  baseDao <- originateUntyped $ UntypedOriginateData
+    { uodFrom = nettestAddress
+    , uodName = "BaseDAO Test Contract"
+    , uodBalance = zeroMutez
+    , uodStorage = untypeValue $ toVal $ storageFn addresses
+    , uodContract = convertContract baseDAOContractLigo
+    }
+  tests addresses (TAddress baseDao)
```
rinn7e, line 46: I was wondering if it is possible to simply share the tests with lorentz basedao. I've done this with my tests, but since the tests are done differently here so I'm not sure. Example share tests: https://github.com/tqtezos/baseDAO/blob/master/src/BaseDAO/ShareTest/FA2.hs

The idea is to use the same tests but with different originate function. The errors for now are not the same (errors in ligo version are all string), but it will be the same after I finished this #106 .

rinn7e (changes requested):

ligo/test/Test/Ligo/BaseDAO/Management.hs
sras: @rinn7e Yea, it should be possible (if the errors are changed after #106), and I did try to do it a little, but when I looked for a constraint like `ParameterC`, I missed the one in `Lorentz.Contract.BaseDAO.Types`. I was also not sure if made sense to reuse it, as the issue asked to reimplement tests due to difference in Haskell and Ligo version. Anyway I didn't look too closely, I guess we can revisit it after #106. 

@pasqu4le Does it make sense to open an issue?

ligo/src/management.mligo
sras: I had the impression that we wanted to restrict the custom entrypoints from changing the configuration in the storage. I think the use case discussed in the channel makes sense. So I have changed the lambdas to be `((bytes * full_storage) -> (operation list * full_storage))`. 

One thing though, I had to refactor the top level `case` a bit to accomodate for the fact that the customCall now returns a full_storage, unlike the other calls that returns `storage` only.

ligo/src/common.mligo
```diff
@@ -0,0 +1,19 @@
...
+
+[@inline]
+let ensure_not_migrated (storage : storage): storage =
+  match storage.migration_status with
+      Not_in_migration -> storage
+    | MigratingTo (p) -> storage
+    | MigratedTo (p) -> (failwith ("MIGRATED") : storage)
```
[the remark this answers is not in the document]
sras: Changed

ligo/src/management.mligo
```diff
@@ -4,36 +4,74 @@
...
 
 #include "types.mligo"
+#include "common.mligo"
 
-
-// -----------------------------------------------------------------
-// Shared Function
-// -----------------------------------------------------------------
 #if !MANAGEMENT_H
 #define MANAGEMENT_H
```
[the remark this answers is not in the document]
sras: Removed.

ligo/src/common.mligo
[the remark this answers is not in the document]
pasqu4le: @sras You removed the import, but left behind the calls to `ensure_not_migrated` in the `token/fa2.mligo` module. 

I guess it isn't failing compilation due to the particular order of the imports in `base_DAO.mligo`, but we should still remove them as they are no longer necessary.

pasqu4le:
@sras I have merged the PR solving #102 in the meanwhile, you need to rebase, but the changes should be limited to the `token.mligo` module, in particular it should be enough to:
1. `#include "common.mligo"`
2. remove the `authorize_admin` copy of the function
3. remove all usages of `ensure_not_migrated`

ligo/test/Test/Ligo/BaseDAO/Management.hs
pasqu4le: Yeah that's my bad, when I wrote that issue it just happened that it was clear we couldn't simply convert between the LIGO and Lorentz version of the contract and reuse all the test groups that we already had.

However Moremi has since made the test re-use as many scenarios as possible successfully, even tho it had to introduce differences in handling the failures.

Since we already have the new tests (will check them now BTW) I think it makes sense to merge this MR already and revisit them in a follow-up issue after #106 is merged as well. I'll make sure to open it.

ligo/test/Ligo/BaseDAO/Types.hs
```diff
@@ -58,10 +64,12 @@ data ProposalL = ProposalL
 
 type CallCustomParam = (MText, ByteString)
 
-data Parameter
+-- NOTE: Constructors of the parameter types should remain sorted for the
+--'ligoLayout' custom derivation to work.
+--
```
pasqu4le, line 69: This is actually unnecessary, because `ligoLayout` is able of sorting the constructors as well (just as LIGO does).

These were in order only because they were written before we had `ligoLayout` and have stayed alphabetically sorted since.

ligo/src/common.mligo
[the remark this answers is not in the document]
sras: @pasqu4le Ah. Missed that. Have removed those now. Also have done the rebase and changes to `token.mligo`.

pasqu4le (approved):
Just a comment that I think can be removed, everything else LGTM.

Feel free to remove it and squash the fixup commits too.

ligo/test/Ligo/BaseDAO/Types.hs
sras: @pasqu4le I actually came across this when the build started failing for me during the derivation of 

```
data MigratableParam
  = XtzForbidden ForbidXTZParam
  | CallCustom CallCustomParam
  deriving stock (Show)
```

With the error `• Couldn't match type ‘ForbidXTZParam’ with ‘(MText, ByteString)’`. I am not sure if I am not doing anything else wrong and it still fails if the order of the constructors is as shown above.

This was fixed when I just tried re-ordering the constructors. I set out to create an issue in morley, but came across the PR for `ligolayout` that mentioned something about sorting, which gave me the impression that it was known. Re-reading it, I have no idea what gave me that impression.

[1] https://gitlab.com/morley-framework/morley/-/merge_requests/731

pasqu4le: @sras you are correct, it seems there is a bug in `ligoLayout` for sum-types, I have reduced it to a simple example and opened an issue in `morley` to fix it.

Please keep the comment up, but add a TODO to update it in the future mentioning [morley #527](https://gitlab.com/morley-framework/morley/-/issues/527)

sras: Will do.

rinn7e (approved):

PR #114, opened 2021-01-14
[#101] Proposals logic in LIGO

### LINKED ISSUE #101: Implement proposal entrypoints for the LIGO version of the contract

We need to implement the "proposal" entrypoints for the LIGO version of the contract.

There is already a `proposal.mligo` source file with the necessary entrypoints (without any implementations).

Most entrypoints of this group will likely not match their Lorentz counterpart very closely, see the differences in the `Configuration`  paragraph between `specification-ligo.md` and `specification.md`.

The difference that will create the discrepancies between the two versions is that in LIGO we'll need to extract the values from the `proposal_metadata`  `map` (based on their `string` name) and `unpack` them.

A note about shared functions:
this module will need an `ensure_not_migrated` function to be imported from the `management.mligo` source file.
This PR can assume that one exists there without import problems (as the ones [discussed here](https://github.com/tqtezos/baseDAO/pull/81#discussion_r549880585)).

One more thing: please check out [this discussion](https://github.com/tqtezos/baseDAO/pull/99/files?file-filters%5B%5D=.cabal&file-filters%5B%5D=.hs&file-filters%5B%5D=.json&file-filters%5B%5D=.license&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.mligo&file-filters%5B%5D=.yaml#r553943207) about the type of `ensure_not_migrated` which has been implemented in that open PR (at the time of writing) first and assume that's the type you are going to find.

Note: in parallel with the Lorentz version the `proposal.mligo` "module" will be the only one importing the `permit.mligo` one and as such it should either fill it with useful functions or delete it if deemed unnecessary.

Last note: tests will need to be re-implemented in Haskell due to the differences between the LIGO and Lorentz version of the contract reflecting in different types.

# Acceptance criteria

All entrypoints defined in `proposal.mligo` have been implemented and there are tests covering them.

Martoon-00:
## Description

Adding proposals and permissions logic in LIGO, adjusting tests to cover it.

## Related issue(s)

Resolves #101

- Tests
        silently reappearing again.

- Documentation
    - Haddock


ligo/src/proposal.mligo
```diff
@@ -4,21 +4,361 @@
...
+  let store = ensure_not_migrated store in
+  let store = authorize_admin store in
+  let checked =
+    if   config.max_quorum_threshold < new_threshold
+      || config.min_quorum_threshold > new_threshold
+      then failwith("OUT_OF_BOUND_QUORUM_THRESHOLD")
+      else ()
+    in
+  let store = { store with quorum_threshold = new_threshold } in
+  (([] : operation list), store)
```
pasqu4le, line 203:
```suggestion
  let store =
    if config.max_voting_period < new_period || config.min_voting_period > new_period
    then (failwith("OUT_OF_BOUND_VOTING_PERIOD") : storage)
    else { store with voting_period = new_period }
  in (([] : operation list), store)
```
I just think it's a bit easier to read

ligo/src/proposal.mligo
```diff
@@ -4,21 +4,361 @@
...
+
+// Used in "flash". See the Haskell version for explanation.
```
pasqu4le, line 205:
```suggestion
// Used in "flush". See the Haskell version for explanation.
```
:camera_flash:

ligo/src/proposal.mligo
```diff
@@ -4,21 +4,361 @@
...
+    then
+      if proposal.upvotes > proposal.downvotes
+      then
+        let store = unfreeze_proposer_token (config, true, proposal, proposal_key, store) in
+        let store = unfreeze_voter_token (proposal, proposal_key, store) in
+        let store = delete_proposal (proposal.start_date, proposal_key, store) in
+        (([] : operation list), store)
+      else (failwith("FAIL_DROP_PROPOSAL_NOT_ACCEPTED") : return)
+    else (failwith("FAIL_DROP_PROPOSAL_NOT_ACCEPTED") : return)
+  else (failwith("FAIL_DROP_PROPOSAL_NOT_OVER") : return)
```
pasqu4le, line 363:
```suggestion
  if   is_voting_period_over(proposal, store)
    && do_total_vote_meet_quorum_threshold(proposal, store)
    && proposal.upvotes > proposal.downvotes
  then
    let store = unfreeze_proposer_token (config, true, proposal, proposal_key, store) in
    let store = unfreeze_voter_token (proposal, proposal_key, store) in
    let store = delete_proposal (proposal.start_date, proposal_key, store) in
    (([] : operation list), store)
  else (failwith("FAIL_DROP_PROPOSAL_NOT_OVER") : return)
```

pasqu4le (commented):
I only reviewed the LIGO part, but it mostly looks good

ligo/src/proposal.mligo
Martoon-00: I was also thinking about doing a similar optimization, but to be on par with Lorentz version we will also have to update `flush` in Lorentz. Though we want to do this optimization there for some time already, let's do it now.

Martoon-00:
Thanks, all other comments are pretty handy, make sense to me. Gonna fix it right after I deal with tests.

ligo/src/proposal.mligo
pasqu4le: I think it's fine to make some optimization on the fly if we notice them, in some cases these can be easier to spot in LIGO (not necessarily this case) and vice-versa.

There's also the fact that parts of the contracts will need to be different and so their optimizations too, so in general I don't think there's a problem with them (although we want to remember this if we need/ to optimize the Lorentz version too).

src/BaseDAO/ShareTest/Proposal/Config.hs
```diff
@@ -0,0 +1,217 @@
...
+  , decisionLambdaConfig
+  , voteConfig
+  ) where
+
+import Lorentz
+import Universum (Constraint, (?:))
+
+import Lorentz.Contracts.BaseDAO.Token.FA2 (creditTo)
+import qualified Lorentz.Contracts.BaseDAO.Types as DAO
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
```
pasqu4le, line 31:
```suggestion
import qualified Lorentz.Contracts.BaseDAO.Types as DAO
```

ligo/src/base_DAO.mligo
```diff
@@ -17,15 +17,15 @@ let base_DAO_contract
...
+      | Drop_proposal (p) -> drop_proposal(p, config, store)
       | Transfer_ownership (p) -> transfer_ownership(p, store)
       | Accept_ownership (p) -> accept_ownership(p, store)
       | Migrate (p) -> migrate(p, store)
       | Confirm_migration (p) -> confirm_migration(p, store)
-      | Propose (p) -> propose(p, store)
-      | Vote (p) -> vote(p, store)
-      | Set_voting_period (p) -> set_voting_period(p, store)
-      | Set_quorum_threshold (p) -> set_quorum_threshold(p, store)
+      | Propose (p) -> propose(config, p, store)
```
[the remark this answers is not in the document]
Martoon-00: Yeah right, good catch :+1: 

Fixed in place.

ligo/src/proposal.mligo
```diff
@@ -4,21 +4,361 @@
...
+let check_proposal_limit_reached (config, propose_params, store : config * propose_params * storage): unit =
+  if config.max_proposals <= List.length store.proposal_key_list_sort_by_date
+    then failwith("MAX_PROPOSALS_REACHED")
+    else ()
+
+[@inline]
+let freeze (tokens, addr, store : nat * address * storage): storage =
+  let store = debit_from (tokens, addr, unfrozen_token_id, store) in
+  let store = credit_to (tokens, addr, frozen_token_id, store) in
+  store
```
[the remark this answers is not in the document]
Martoon-00: My version seems a bit better to me, because this way it is easier to visually compare those two lines (with `credit_to` and `debit_from`) and spot a bug if it is present.

ligo/src/proposal.mligo
```diff
@@ -4,21 +4,361 @@
...
+    then failwith("VOTING_INSUFFICIENT_BALANCE")
+    else ()
+
+let submit_vote (vote_param, author, store : vote_param * address * storage): storage =
+  let proposal_key = vote_param.proposal_key in
+  let proposal =
+    match Map.find_opt proposal_key store.proposals with
+      None -> (failwith("PROPOSAL_NOT_EXIST") : proposal)
+    | Some p -> p
+    in
```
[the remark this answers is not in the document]
Martoon-00: Good catch!

ligo/src/proposal.mligo
Martoon-00: Mm, I tried this and the condition does not fit into 80 chars now :confused: 

While in other places like function signatures this is not crucial, here it seems to be.

ligo/src/proposal.mligo
Martoon-00: Great :facepalm:

ligo/src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: Fair enough, you have a point there :+1:

ligo/src/proposal.mligo
pasqu4le: What do you mean it's crucial? For compilation?

If for reading it's ok, the suggestion was more about returning a `storage` for the `if` by updating it in the successful branch instead of afterwards.

ligo/test/LICENSE
```diff
@@ -0,0 +1,19 @@
+MIT License Copyright (c) 2020 TQ Tezos
```
pasqu4le, line 1: AFAIU this file is not necessary and can be removed.

`reuse lint` seems to report all things being ok without it as well.

ligo/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -0,0 +1,19 @@
...
+  ) where
+
+import Universum hiding (compare, drop, (>>))
+
+import Test.Tasty (TestTree)
+
+import BaseDAO.ShareTest.Proposal
+import Test.Ligo.BaseDAO.Common
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
```
pasqu4le, line 15:
```suggestion
import Test.Tasty (TestTree)

import BaseDAO.ShareTest.Proposal
import Test.Ligo.BaseDAO.Common
```
It seems both the duplication ignoring and the universum import are unnecessary here.

ligo/src/proposal.mligo
Martoon-00: Crucial for readability, e.g. when I have two panes opened then part of the condition does not fit into the screen.

Though the second part - yep, will do now.

ligo/test/LICENSE
Martoon-00: This is about complaints of cabal, we refer to a license file from there and it was not present.

ligo/test/baseDAO-ligo-tests.cabal
```diff
@@ -3,6 +3,8 @@ cabal-version: 2.2
 -- This file has been generated from package.yaml by hpack version 0.34.2.
 --
 -- see: https://github.com/sol/hpack
+--
+-- hash: 54f52765cc1d2f811797ae85a910c8dfc0716621547586a64685aea1176d8775
```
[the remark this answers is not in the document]
Martoon-00: Oh thanks :+1:

Martoon-00:
Ok, mostly everything is finished, I'll remove `: unit` things now and that should be it.

ligo/test/LICENSE
pasqu4le: Ah I see :ok_hand:

ligo/src/proposal.mligo
pasqu4le: Ok, I agree. For a moment there I thought you had found another oddity of the cameligo syntax :sweat_smile:

pasqu4le (approved):

PR #118, opened 2021-01-15
[#106] Update LIGO `failwith` to be the same as Lorentz version

### LINKED ISSUE #106: Extend the use of LIGO failwith to pairs

rinn7e:
## Description

Problem: Current LIGO only fails with type `string` which is different
from the Lorentz version. Because of this, we require a `isLorentz`
check in the tests to be able to share the tests properly. We can update
the `failwith` in LIGO to be the same as lorentz which would make the
check no longer necessary.

Solution: Update LIGO `failwith` to be the same as Lorentz version,
remove `isLorentz` check in share tests.

## Related issue(s)

Resolves #106  

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/src/base_DAO.mligo
+++ b/ligo/src/base_DAO.mligo
@@ -26,7 +26,9 @@ let base_DAO_contract
             | M_right (fbp) ->
                 let result =
                   if Tezos.amount > 0tez
-                    then (failwith "FORBIDDEN_XTZ" : return)
+                    then
+                      ([%Michelson ({| { FAILWITH } |} : string * unit -> return)]
+                        ("FORBIDDEN_XTZ", ()) : return)
                     else
                       begin
                         match fbp with
--- a/ligo/test/Test/Ligo/BaseDAO/Management.hs
+++ b/ligo/test/Test/Ligo/BaseDAO/Management.hs
@@ -115,7 +115,7 @@ test_BaseDAO_Management =
               callFrom (AddressResolved newAddress1) baseDao (Call @"Confirm_migration") ()
               callFrom (AddressResolved owner) baseDao (Call @"Transfer_ownership")
                 (#newOwner .! newOwner)
-                & expectMigrated
+                & expectMigrated newAddress1
     ]
   , testGroup "Accept Ownership"
       [ nettestScenarioCaps "authenticates the sender" $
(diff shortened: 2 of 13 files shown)


ligo/test/Test/Ligo/BaseDAO/Management.hs
```diff
@@ -243,29 +243,29 @@ test_BaseDAO_Management =
...
-  => m a -> m ()
-expectMigrated act = expectFailure act (NettestFailedWithError [mt|MIGRATED|])
+  => Address -> m a -> m ()
+expectMigrated addr = expectCustomError #mIGRATED addr
 
 expectForbiddenXTZ
   :: (MonadNettest caps base m)
   => m a -> m ()
-expectForbiddenXTZ act = expectFailure act (NettestFailedWithError [mt|FORBIDDEN_XTZ|])
+expectForbiddenXTZ = expectCustomError_ #fORBIDDEN_XTZ
```
pasqu4le, line 271: Nitpick: missing an end-of-file newline

ligo/src/base_DAO.mligo
```diff
@@ -26,7 +26,9 @@ let base_DAO_contract
             | M_right (fbp) ->
                 let result =
                   if Tezos.amount > 0tez
-                    then (failwith "FORBIDDEN_XTZ" : return)
+                    then
+                      ([%Michelson ({| { FAILWITH } |} : string * unit -> return)]
+                        ("FORBIDDEN_XTZ", ()) : return)
```
Martoon-00, line 31: Wow, this even compiles in what one would expect in the resulting Michelson, nice.

Martoon-00 (approved):

pasqu4le (commented):
Only a nitpick and one more thing is missing 
> the specification-ligo.md section on Errors because it says it always fails with a string

The first line under `Errors` should be changed to:
> In error scenarios the baseDAO contract fails with a string or a pair where the first item is a string.

like in `specification.md

pasqu4le (approved):

PR #120, opened 2021-01-15
[#108] Improve LIGO version documentation

### LINKED ISSUE #108: Improve LIGO version documentation

The LIGO version of the contract only has some basic documentation set up, we need to extend it.

In particular we need its README to contain info about:
- mentions of the fact that it's still in development
- brief explanation of the differences with Lorentz (compile-time template-based vs origination-time option-based)
- how to obtain the contract (`make`)
- how to generate its storage (this can be an empty TODO paragraph for now, as we don't have it set up)
- how use `tezos-client` to originare it.
- how to run tests (`make`)

The more detailed part of the explanation of how the tests are implemented (haskell `morley`/`cleveland`) should be handled in the `test/` `README` instead.

This should also modify the top-level README, because this mentions that the LIGO version is in development by pointing at a merged PR (it should point at the ligo README instead).

# Acceptance criteria

- `ligo /README.md` and `ligo/test/README.md` contain some explanations about the contract and how to interact with it.
- the top-level `README` points at it

rinn7e:
## Description

Problem: The LIGO version of the contract only has some basic
documentation set up, we need to extend it.

Solution: Update LIGO README.md, test README.md, and link LIGO doc to
the main README instead of a PR.

## Related issue(s)

Resolves #108 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/README.md
+++ b/ligo/README.md
@@ -1,14 +1,51 @@
-# BaseDAO LIGO implementation
+# BaseDAO LIGO
 
-## Prerequisites
+__Note: This is currently in a very raw stage and can be originated only using
+[this approach](https://gitlab.com/tezos/tezos/-/issues/1053#note_481537115).__
 
-You will need [ligo executable](https://ligolang.org/docs/intro/installation) installed.
+This is the LIGO implementation of BaseDAO, written in the cameLIGO dialect.
+
+The main differences with its [Lorentz counterpart](/README.md) are:
+- `BaseDAO Lorentz`:
+    - uses compile-time configuration to define various options of a DAO,
+    - as a consequence, each contract implementing over `baseDAO` will produce a different `Michelson` contract.
+    - the configuration shapes the code before its generated, so the result is less flexible but more efficient
+
+- `BaseDAO LIGO`:
+    - uses runtime configuration, which is part of the overall storage of the contract
+    - only one `Michelson` contract is used, the user simply defines the configuration as part of the storage at origination
+    - the configuration is used by the code when it's executed, so the result is more flexible but not as efficient
 
 ## Build instructions
 
-Use Makefile to build a contract, for instance
+You will need the [ligo executable](https://ligolang.org/docs/intro/installation) installed to generate the contract.
 
+You can use the [`Makefile`](./Makefile) to build the LIGO contract by simply running:
 ```sh
 make all
 ```
+which will save the compiled contract in `out/baseDAO.tz`.
 
+## Generating `Storage`
+
+Currently we don't provide any convenient way to generate the storage, so you'll
+need to do so manually.
+
+We will work on this in future versions.
+
+## Using the contract
+
+You will need [tezos-client](http://tezos.gitlab.io/introduction/howtoget.html) or similar software to originate and interact with the contract on the Tezos network.
+
+For example, if you have compiled the contract and obtained a storage (`<storage>`) as described above you can originate it with:
+`tezos-client originate contract myDAO transferring 1 from alice running out/baseDAO.tz --init <storage>`
+
+## Testing
+
+The tests for this contract are written in Haskell and you will need the `stack` tool to run them.
+
+You can use the `Makefile` to do so as well:
+```
+make test
+```
+For more detail about LIGO tests, check out its [README.md](/ligo/test/)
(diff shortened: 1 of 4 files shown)


ligo/README.md
```diff
@@ -1,14 +1,43 @@
...
 make all
 ```
 
+## Generating `Storage`
+TODO
+
+## Originating the contract
+To originate `baseDAO LIGO`, run:
+```
+TODO
```
rinn7e, line 32: @pasqu4le Do we have a sample storage for LIGO dao? Since to originate the contract, we need a storage.

Hope to get review on the rest of the change as well.

ligo/README.md
```diff
@@ -1,14 +1,43 @@
...
 
 ## Prerequisites
 
 You will need [ligo executable](https://ligolang.org/docs/intro/installation) installed.
 
 ## Build instructions
 
-Use Makefile to build a contract, for instance
-
+To build the LIGO contract, simply run:
```
pasqu4le, line 21:
```suggestion
You can use the [`Makefile`](./Makefile) to build the LIGO contract by simply running:
```

ligo/README.md
```diff
@@ -1,14 +1,43 @@
...
 ```sh
 make all
 ```
```
pasqu4le, line 25:
```suggestion
which will save the compiled contract in `out/baseDAO.tz`.
```

ligo/README.md
```diff
@@ -1,14 +1,43 @@
...
+
+- `BaseDAO LIGO`:
+    - uses runtime configuration
+    - only one `michelson` contract is used, the user simply defines the configuration as part of originated storage
 
 ## Prerequisites
 
 You will need [ligo executable](https://ligolang.org/docs/intro/installation) installed.
 
 ## Build instructions
```
pasqu4le, line 19: I suggest moving `ligo` to be a build dependency:
```suggestion
## Build instructions

You will need the [ligo executable](https://ligolang.org/docs/intro/installation) installed to generate the contract.
```

ligo/README.md
```diff
@@ -1,14 +1,43 @@
...
 ```
 
+## Generating `Storage`
+TODO
+
+## Originating the contract
+To originate `baseDAO LIGO`, run:
+```
+TODO
+```
```
pasqu4le, line 33: Here instead I suggest mentioning `tezos-client` as a possible choice only.

I don't think it's a good idea to write an actual storage in this README, let's only mention an example call.
```suggestion
## Using the contract

You will need [tezos-client](http://tezos.gitlab.io/introduction/howtoget.html) or similar software to originate and interact with the contract on the Tezos network.

For example, if you have compiled the contract and obtained a storage (`<storage>`) as described above you can originate it with:
`tezos-client originate contract myDAO transferring 1 from alice running out/baseDAO.tz --init <storage>`
```

ligo/README.md
```diff
@@ -1,14 +1,43 @@
...
+
+## Testing
+
+`baseDAO LIGO` shares the tests with `baseDAO Lorentz` which are implemented in Haskell.
+
+To run the test:
```
pasqu4le, line 39: I think it's not necessary to repeat this here, let's just leave the line below pointing at the test's README for more info:
```suggestion
The tests for this contract are written in Haskell and you will need the `stack` tool to run them.

You can use the `Makefile` to do so as well:
```

ligo/README.md
```diff
@@ -1,14 +1,43 @@
-# BaseDAO LIGO implementation
+# BaseDAO LIGO
+
+__Note: This is still current in development.__
+
+This is the LIGO implementation of BaseDAO.
```
pasqu4le, line 5:
```suggestion
This is the LIGO implementation of BaseDAO, written in the cameLIGO dialect.
```

pasqu4le (commented):
I have a few suggestions

ligo/README.md
pasqu4le: I'm not sure we'll end up providing one at all, hence why the storage section is a TODO for now.

I suggested below in a comment how I think we should handle this.

pasqu4le:
@rinn7e one more thing: I wrote a comment on the issue about an additional point that I would like was addressed here as well, do you mind adding another commit to do that?

rinn7e:
> @rinn7e one more thing: I wrote a comment on the issue about an additional point that I would like was addressed here as well, do you mind adding another commit to do that?

Ah, I forgot about that. I've update the code block to use `ocaml` instead, as well as updating all your other suggestion. You can check this.

gromakovsky (changes requested):
Also note that CI detected a trailing whitespace.

gromakovsky (approved):

Issue #123, opened 2021-01-16
Simplify LIGO version deployment


gromakovsky:
# Clarification and motivation

Currently we say that the only way to originate the LIGO contract is to use https://gitlab.com/tezos/tezos/-/issues/1053#note_481537115, but there is no tool implementing this approach and forcing all our users to do it themselves is not an option.
We can make some optimizations, but I highly doubt we will manage to fit the contract origination into 16384 bytes. If we don't manage, we have 2 options:
1. Use that approach and provide some tooling around it. https://gitlab.com/morley-framework/morley/-/issues/530 can be helpful.
2. Move code into storage using approach described in https://gitlab.com/morley-framework/indigo/-/issues/23.

If we manage to optimize the contract sufficiently, we should simply update the relevant part(s) of README document(s). (e. g. see https://github.com/tqtezos/baseDAO/blob/f7723283c8d1eb65988facf4ecf3f4183a684695/ligo/README.md#basedao-ligo)

# Acceptance criteria

We don't force users to implement https://gitlab.com/tezos/tezos/-/issues/1053#note_481537115 fully themselves, we provide all necessary building blocks to deploy our LIGO contract. Documentation/instructions are updated accordingly.

PR #125, opened 2021-01-16
[#116] LIGO RegistryDAO contract using BaseDAO

### LINKED ISSUE #116: Implement Registry DAO using LIGO version

Now that we have a LIGO version of BaseDAO we want to demonstrate its usage. Let's implement [RegistryDAO](https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md) as an example.

# Acceptance criteria

1. Lambdas that are used to parameterize BaseDAO are implemented (in LIGO) according to the RegistryDAO specification.
2. There are instructions how to originate LIGO version of RegistryDAO and these instructions are as simple as possible.

sras:
Problem: As an example, we would like to implement RegistryDAO contract
using ligo version of the BaseDAO contract.

Solution: Implement configuration lambdas for BaseDAO contract which
implements RegistryDAO behavior.

Originally, RegistryDAO contract requires a parameterized contract, but
since it is not clear how to do this in LIGO, the Key/Value types are
just included as a type alias in the source file.

Apart from this, the implementation of decision lambda is a bit hackish
where the two types of proposals are distinguished by looking at the
included keys in the proposal metadata, which is just a `(string, bytes)
map`.

A dummy entrypoint is left in the RegistryDAO source file to facilitate
creation of a Michelson storage expression using the LIGO
compile_storage command.

#### Origination:

During origination of BaseDAO, the configuration for RegistryDAO can be
loaded by making a storage with RegistryDAO lambdas included in
configuration and required map keys in storage. A Michelson expression
for storage can be built during origination using the following LIGO
Command.

```
ligo compile-storage ligo/src/registryDAO.mligo dummy
  'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
```

Where the arguents to `default_registry_DAO_full_storage` function are
admin address, token_address and values of configuration parameters 'a',
'b', 's_max', 'c' and 'd'.

## Related issue(s)

Addresses part of  #116 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/README.md
+++ b/ligo/README.md
@@ -28,8 +28,8 @@ which will save the compiled contract in `out/baseDAO.tz`.
 
 ## Generating `Storage`
 
-Currently we don't provide any convenient way to generate the storage, so you'll
-need to do so manually.
+Currently we don't provide any convenient way to generate `BaseDAO` storage,
+but you can use [RegistryDAO Storage generation](#registrydao) as example.
 
 We will work on this in future versions.
 
@@ -49,3 +49,22 @@ You can use the `Makefile` to do so as well:
 make test
 ```
 For more detail about LIGO tests, check out its [README.md](./haskell/test/)
+
+## Example contracts
+
+### RegistryDAO
+
+The storage, including configuration for the RegistryDAO
+[specification](https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md)
+is included in `src/registryDAO.mligo`. You can convert this into a
+Michelson expression during the BaseDAO origination using the `ligo
+compile-storage` command as follows.
+
+```
+
+The `default_registry_DAO_full_storage` is a LIGO function defined in
+`ligo/src/RegistryDAO.mligo`, which returns a LIGO expression for storage,
+which is converted into Michelson using the `compile-storage` command. The
+arguments to this function are the admin address, the token address and the
+configuration parameters described in the Registry spec, which are 'a', 'b',
+'s_max', 'c' and 'd'.
--- a/ligo/src/proposal.mligo
+++ b/ligo/src/proposal.mligo
@@ -348,4 +348,3 @@ let drop_proposal (proposal_key, config, store : proposal_key * config * storage
   else
     ([%Michelson ({| { FAILWITH } |} : string * unit -> return)]
       ("FAIL_DROP_PROPOSAL_NOT_OVER", ()) : return)
-
--- a/ligo/src/registryDAO.mligo
+++ b/ligo/src/registryDAO.mligo
@@ -1,0 +1,184 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#include "defaults.mligo"
+#include "types.mligo"
+#include "proposal.mligo"
+#include "registryDAO/types.mligo"
+
+// The following include is required so that we have a function with an
+// entrypoint type to use with `compile-storage` command to make Michelson
+// storage expression.
+#include "base_DAO.mligo"
+
+let require_extra_value (key_name, ce : string * contract_extra) : bytes =
+    match Map.find_opt key_name ce with
+      Some (packed_b) -> packed_b
+    | None -> (failwith "CONFIG_VALUE_NOT_FOUND" : bytes)
+
+let require_unpacked_nat_opt (packed_b: bytes) : nat option =
+  match ((Bytes.unpack packed_b) : ((nat option) option)) with
+    Some (v) -> v
+  | None -> (failwith "UNPACKING_FAILED" : nat option)
+
+let require_unpacked_registry (packed_b: bytes) : registry =
+  match ((Bytes.unpack packed_b) : (registry option)) with
+    Some (v) -> v
+  | None -> (failwith "UNPACKING_FAILED" : registry)
+
+let require_unpacked_registry_affected (packed_b: bytes) : registry_affected =
+  match ((Bytes.unpack packed_b) : (registry_affected option)) with
+    Some (v) -> v
+  | None -> (failwith "UNPACKING_FAILED" : registry_affected)
+
+let lookup_config (key_name, ce : string * contract_extra) : nat option =
+    require_unpacked_nat_opt(require_extra_value(key_name, ce))
+
+let require_config (key_name, ce : string * contract_extra) : nat =
+  match lookup_config (key_name, ce) with
+    Some (v) -> v
+  | None -> (failwith "REQUIRED_CONFIG_NOT_FOUND" : nat)
+
+let apply_diff_registry (diff, registry : registry_diff * registry) : registry =
+  let
+    foldFn (registry, update: registry * (registry_key * registry_value option)) : registry =
+      let registry_key = update.0 in
+      let registry_value = update.1 in
+      Map.update registry_key registry_value registry
+  in List.fold foldFn diff registry
+
+let apply_diff_registry_affected (proposal_key, diff, registry_affected : proposal_key
+        * registry_diff * registry_affected) : registry_affected =
+  let
+    foldFn (registry_affected, update: registry_affected * (registry_key *
+      registry_value option)) : registry_affected =
+      let registry_key = update.0 in
+      Map.update registry_key (Some (proposal_key)) registry_affected
+  in List.fold foldFn diff registry_affected
+
+let apply_diff (diff, ce : registry_diff * contract_extra) : contract_extra =
+  let registry = require_unpacked_registry(require_extra_value("registry", ce)) in
+  let new_registry : registry = apply_diff_registry (diff, registry) in
+  Map.update "registry" (Some (Bytes.pack new_registry)) ce
+
+let apply_diff_affected (proposal_key, diff, ce : proposal_key *
+    registry_diff * contract_extra) : contract_extra =
+  let registry_af = require_unpacked_registry_affected(require_extra_value("registry_affected", ce)) in
+  let new_registry_af : registry_affected = apply_diff_registry_affected (proposal_key, diff, registry_af)
+  in Map.update "registry_affected" (Some (Bytes.pack new_registry_af)) ce
+
+// Configuration Lambdas
+
+(*
+ * Proposal check lambda : returns true if following two checks are true.
+ * 1. if proposer has locked a * s + b where s = size(pack(proposalMetadata))
+ * and a and b are from storage configuration.
+ * 2. proposal size < s_max where s_max is from stored configuration.
+ *)
+let registry_DAO_proposal_check (params, store : propose_params * storage) : bool =
+  let proposal_size = Bytes.size(Bytes.pack(params.proposal_metadata)) in
+  let frozen_scale_value = require_config ("a", store.extra) in
+  let frozen_extra_value = require_config ("b", store.extra) in
+  let max_proposal_size = require_config ("s_max", store.extra) in
+  let requred_token_lock = frozen_scale_value * proposal_size + frozen_extra_value
+  in (params.frozen_token = requred_token_lock) && (proposal_size < max_proposal_size)
+
+(*
+ * Proposal rejection return lambda: returns `c * frozen / d`. where c and d
+ * are specified by the DAO creator in configuration and frozen is the amount
+ * that was frozen by the proposer.
+ *)
+let registry_DAO_rejected_proposal_return_value (params, store : proposal * storage) : nat =
+  let slash_scale_value = require_config ("c", store.extra) in
+  let slash_division_value = require_config ("d", store.extra)
+  in (slash_scale_value * params.proposer_frozen_token) / slash_division_value
+
+(*
+ * Uses the keys in the proposal metadata to differentiate between configuration
+ * proposal and normal proposals. If the map contains `agoraPostID` key, the proposal
+ * is considerd as 'normal'. Or else it is considered as carrying a 'configuration'
+ * proposal.
+ *
+ * For 'normal' proposals, the contract_extra is expected to contain the bigmap
+ * for registry and bigmap for key update tracking under keys 'registry' and
+ * 'registry_affected' keys.
+ *
+ * For 'configuration' proposal, the code expects all the configuration keys
+ * included in the proposal (or it throws an error). The expected keys are "a",
+ * "b", "s_max", "c" and "d".  'None' values are ignored and 'Some' values are
+ * used for updating corresponding configuraton.
+ *)
+let registry_DAO_decision_lambda (proposal, store : proposal * storage) : return =
+  let propose_param : propose_params = {
+    frozen_token = proposal.proposer_frozen_token;
+    proposal_metadata = proposal.metadata
+    } in
+  let proposal_key = to_proposal_key (propose_param, proposal.proposer) in
+  match Map.find_opt "agoraPostID" proposal.metadata with
+    Some (packed_b) ->
+      begin
+        match ((Bytes.unpack packed_b) : (nat option)) with
+          Some (agora_post_id) ->
+            begin
+              match Map.find_opt "updates" proposal.metadata with
+                Some (packed_b) ->
+                  begin
+                    match ((Bytes.unpack packed_b) : (registry_diff option)) with
+                      Some (diff) ->
+                        let ce_after_registry_update = apply_diff(diff, store.extra) in
+                        let ce_after_registry_affected_update =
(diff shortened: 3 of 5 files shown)


rinn7e:
> Origination:

> During origination of BaseDAO, the configuration for RegistryDAO can be
loaded by making a storage with RegistryDAO lambdas included in
configuration and required map keys in storage. A Michelson expression
for storage can be built during origination using the following LIGO
Command.

> ligo compile-storage ligo/src/registryDAO.mligo dummy 'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address),("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
  
> ...

Maybe this should be added to Basedao Ligo `README.md`.

And I'm not sure if this is in the scope of this MR, but we should have some tests involving ligo registryDAO as well.

sras:
>Maybe this should be added to Basedao Ligo README.md.

Do you have a suggestion where in Readme, we should include it? 

>And I'm not sure if this is in the scope of this MR, but we should have some tests involving ligo registryDAO as well.

Wasn't sure of that myself. I think it requires writing the Michelson expression for storage to a file and reading it In Haskell tests into a Typed/Untyped `Value` and originating the contract with it. So we should probably have to include it in CI as well to generate this storage expression file.

Or is there a better way?

rinn7e:
> Do you have a suggestion where in Readme, we should include it?

For me, I would add a section "Example Contracts", and add a subsection about registryDAO there. 

> Or is there a better way?

I'm not sure myself, maybe @pasqu4le could comment on this.

sras:
>For me, I would add a section "Example Contracts", and add a subsection about registryDAO there.

@rinn7e Have added this.

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,190 @@
...
+  match Map.find_opt "registry" ce with
+    Some (packed_registry) ->
+      begin
+        match (Bytes.unpack (packed_registry) : registry option) with
+          Some (registry) -> let
+            new_registry : registry = apply_diff_registry (diff, registry)
+            in Map.update "registry" (Some (Bytes.pack new_registry)) ce
+        | None -> (failwith "REGISTRY_NOT_FOUND" : contract_extra)
+      end
+  | None -> (failwith "REGISTRY_NOT_FOUND" : contract_extra)
```
pasqu4le, line 57: Why not replace this outer error with `require_config` as well?
I think the generic error is fine.

Sadly I cannot think of a way to de-duplicate the logic of `Bytes.unpack` on the spot because of the type signature requirement :thinking:

pasqu4le:
> Or is there a better way?

I cannot think of one, it seems like that's the way to go.

I would put the storage file generation in the Makefile as well BTW, if not for simply `make` at least in a way that can be used by `make test`.

Have you tried this BTW? It should be able to be originated using the complex method (see #123 ) and I hope the cost of the operations is small enough.

ligo/src/proposal.mligo
```diff
@@ -349,3 +349,5 @@ let drop_proposal (proposal_key, config, store : proposal_key * config * storage
     ([%Michelson ({| { FAILWITH } |} : string * unit -> return)]
       ("FAIL_DROP_PROPOSAL_NOT_OVER", ()) : return)
 
+let flush(param, config, store : nat * config * storage) : return =
+  not_implemented("flush")
```
rinn7e, line 353: This shouldn't be here, probably from rebase.

rinn7e (changes requested):

docs/specification-ligo.md
```diff
@@ -873,3 +873,25 @@ Motivation:
...
+```
+ligo compile-storage ligo/src/registryDAO.mligo dummy 'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
+```
+
+The `default_registry_DAO_full_storage` is a LIGO function defined in
+`ligo/src/RegistryDAO.mligo`, which returns a LIGO expression for storage,
+which is converted into Michelson using the `compile-storage` command. The
+`dummy` entrypoint is just included to make this command work. The arguments to
+this function are the admin address, the token address and the configuration
+parameters described in the Registry spec, which are 'a', 'b', 's_max', 'c' and 'd'.
```
rinn7e, line 897: I think this should be in `ligo/README.md` instead of `specification-ligo.md`.

rinn7e (changes requested):

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,190 @@
...
+  match Map.find_opt "agoraPostID" proposal.metadata with
+    Some (packed_b) ->
+      begin
+        match ((Bytes.unpack packed_b) : (nat option)) with
+          Some (agora_post_id) ->
+            begin
+              match Map.find_opt "updates" proposal.metadata with
+                Some (packed_b) ->
+                  begin
+                    match ((Bytes.unpack packed_b) : (((registry_key, registry_value option) map) option)) with
```
rinn7e, line 132: Is it necessary to use `map` here for `updates`:

```
((registry_key, registry_value option) map)
```

AFAIU, in the spec, it is meant to be a `list` instead: https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md#L25

docs/specification-ligo.md
sras: Ah, yes. Btw there is already some mention of storage generation in that file. Do you think if we should amend it, point to the RegistryDAO section maybe?

https://github.com/tqtezos/baseDAO/blob/master/ligo/README.md#generating-storage

ligo/src/proposal.mligo
```diff
@@ -22,3 +22,6 @@ let set_quorum_threshold(param, store : voting_period * storage) : return =
 
 let flush(param, config, store : nat * config * storage) : return =
   not_implemented("flush")
+
+let to_proposal_id (proposal_param, proposer : propose_params * address) : proposal_key =
+  Crypto.blake2b (Bytes.pack (proposal_param, proposer))
```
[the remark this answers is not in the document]
sras: Rebased and now changed the code to use `to_proposal_key` from master.

ligo/src/types.mligo
```diff
@@ -242,8 +242,55 @@ type config =
   ; custom_entrypoints : custom_entrypoints
   }
```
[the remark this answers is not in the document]
sras: Moved the default config/storage values to `default.ligo`.

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,190 @@
...
+      in (([] : operation list), { store with extra = new_ce })
+
+let default_registry_DAO_full_storage (admin, token_address, a, b, s_max, c, d
+    : (address * address * nat * nat * nat * nat * nat)) : full_storage = let
+  fs = default_full_storage (admin, token_address) in
+  let new_storage = { fs.0 with
+    extra = Map.literal [
+          ("registry" , Bytes.pack (Map.empty : registry));
+          ("registry_affected" , Bytes.pack (Map.empty : registry_affected));
+          ("a" , Bytes.pack a);
```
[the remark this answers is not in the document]
sras: Have added.

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,190 @@
...
+#include "proposal.mligo"
+#include "registryDAO/types.mligo"
+
+(*
+ * This is a dummy entrypoint so that this file can be called with LIGO's `compile-storage` command as shown below, providing admin and token addresses.
+ *
+ * ligo compile-storage ligo/src/registryDAO.mligo dummy 'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
+ *)
+let dummy (action, store : parameter * full_storage) : operation list * full_storage
+  = (([] : operation list), store)
```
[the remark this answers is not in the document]
sras: Yes, it works. Changed.

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,190 @@
...
+
+let require_config (key_name, ce : string * contract_extra) : nat =
+  match lookup_config (key_name, ce) with
+    Some (v) -> v
+  | None -> (failwith "REQUIRED_CONFIG_NOT_FOUND" : nat)
+
+let apply_diff_registry (diff, registry
+  : ((registry_key, registry_value option) map) * registry) : registry =
+  let
+    foldFn (registry, update: registry * (registry_key * (registry_value option))) : registry
```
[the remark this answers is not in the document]
sras: Done.

ligo/src/registryDAO.mligo
sras: Originally, the `require_config` only worked with `nat option`, but here we needed a map. But I have now refactored it with some helpers that this verbosity is a bit reduced.

ligo/src/proposal.mligo
sras: Removed.

ligo/src/registryDAO.mligo
sras: Good catch. Switched to use list.

docs/specification-ligo.md
rinn7e: Yeah, I'm good with that. Maybe we could update it to something like this or similar:

```
Currently we don't provide any convenient way to generate `BaseDAO` storage, but you can use [RegistryDAO Storage generation](link-to-section) as example.
```

rinn7e (approved):

docs/specification-ligo.md
sras: Done.

ligo/src/registryDAO.mligo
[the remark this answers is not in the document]
pasqu4le: I'm not sure I understand the problem with tests, wouldn't we be able to have different version of `defaultStorage` for different tests?

Or is the problem that we're going to need too many of them because the tests need a lot of different ones?

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,189 @@
...
+  match lookup_config (key_name, ce) with
+    Some (v) -> v
+  | None -> (failwith "REQUIRED_CONFIG_NOT_FOUND" : nat)
+
+let apply_diff_registry (diff, registry : registry_diff * registry) : registry =
+  let
+    foldFn (registry, update: registry * (registry_key * registry_value option)) : registry =
+    let registry_key = update.0 in
+    let registry_value = update.1 in
+    Map.update registry_key registry_value registry
```
pasqu4le, line 47:
```suggestion
      let registry_key = update.0 in
      let registry_value = update.1 in
      Map.update registry_key registry_value registry
```
just indentation

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,189 @@
...
+  in List.fold foldFn diff registry
+
+let apply_diff_registry_affected (proposal_key, diff, registry_affected : proposal_key
+        * registry_diff * registry_affected) : registry_affected =
+  let
+    foldFn (registry_affected, update: registry_affected * (registry_key *
+      registry_value option)) : registry_affected =
+      let registry_key = update.0 in
+        Map.update registry_key (Some (proposal_key)) registry_affected in
+    List.fold foldFn diff registry_affected
```
pasqu4le, line 57:
```suggestion
      Map.update registry_key (Some (proposal_key)) registry_affected
  in List.fold foldFn diff registry_affected
```
again only keeping with the indentation

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,189 @@
...
+
+let apply_diff (diff, ce : registry_diff * contract_extra)
+  : contract_extra =
+      let registry = require_unpacked_registry(require_extra_value("registry", ce)) in
+      let new_registry : registry = apply_diff_registry (diff, registry) in
+        Map.update "registry" (Some (Bytes.pack new_registry)) ce
```
pasqu4le, line 63:
```suggestion
let apply_diff (diff, ce : registry_diff * contract_extra) : contract_extra =
  let registry = require_unpacked_registry(require_extra_value("registry", ce)) in
  let new_registry : registry = apply_diff_registry (diff, registry) in
  Map.update "registry" (Some (Bytes.pack new_registry)) ce
```
the first line fits into 80 char limit as well.

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,189 @@
...
+let registry_DAO_proposal_check (params, store : propose_params * storage) : bool =
+  let proposal_size = Bytes.size(Bytes.pack(params.proposal_metadata)) in
+  let frozen_scale_value = require_config ("a", store.extra) in
+  let frozen_extra_value = require_config ("b", store.extra) in
+  let max_proposal_size = require_config ("s_max", store.extra) in
+  let requred_token_lock = frozen_scale_value * proposal_size + frozen_extra_value
+  in (params.frozen_token = requred_token_lock) && (proposal_size < max_proposal_size)
+
+(*
+ * Proposal rejection return lambda: returns `c * frozen / d`. wehre c and d
```
pasqu4le, line 88:
```suggestion
 * Proposal rejection return lambda: returns `c * frozen / d`. where c and d
```
Same as before :sweat_smile:

ligo/src/registryDAO.mligo
```diff
@@ -0,0 +1,189 @@
...
+ *)
+let registry_DAO_rejected_proposal_return_value (params, store : proposal * storage) : nat =
+  let slash_scale_value = require_config ("c", store.extra) in
+  let slash_division_value = require_config ("d", store.extra)
+  in (slash_scale_value * params.proposer_frozen_token) / slash_division_value
+
+(*
+ * Proposal rejection return lambda: returns `c * frozen / d`. where c and d
+ * are specified by the DAO creator in configuration and frozen is the amount
+ * that was frozen by the proposer.
```
pasqu4le, line 100: It seems this should be updated or just removed, probably just the result of copy-paste from the previous function.

The rest of the comment looks ok tho.

ligo/README.md
```diff
@@ -49,3 +49,24 @@ You can use the `Makefile` to do so as well:
...
+### RegistryDAO
+
+The storage, including configuration for the RegistryDAO
+[specification](https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md)
+is included in `ligo/src/registryDAO.mligo`. You can convert this into a
+Michelson expression during the BaseDAO origination using the `ligo
+compile-storage` command as follows.
+
+```
+ligo compile-storage ligo/src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
```
pasqu4le, line 64:
```suggestion
is included in `src/registryDAO.mligo`. You can convert this into a
Michelson expression during the BaseDAO origination using the `ligo
compile-storage` command as follows.

```
ligo compile-storage src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
```
I think we should remove `ligo` from the path(s) because this file is in `ligo/` already and the other commands (aka `make`) already assume we are in this directory.

pasqu4le (commented):
Only minor points, but LGTM.

Since we are going to merge this soon, my request to put the storage generation in the `Makefile` can be handled with the following PRs with the tests.

BTW on this point: please remove the `Resolves` from the issue description, we'll have that next PR close the issue.

ligo/src/registryDAO.mligo
sras: Done

pasqu4le (approved):

PR #128, opened 2021-01-18
[#119] Optimize compiled ligo contract

### LINKED ISSUE #119: Optimize LIGO contract using morley optimizer

We have an `optimize` command in `morley` that can make a Michelson contract a bit smaller. We'd like to use it because the smaller our contracts are, the better (as long as all the functionality is preserved).

# Acceptance criteria

LIGO version of BaseDAO that appears in build artifacts, releases and is passed to tests is optimized using `morley optimize`.

Note that we have an example in https://github.com/tqtezos/stablecoin.

rvem:
## Description
Problem: 'morley' has an option for contract optimization, which is
supposed to reduce contract size and its gas consumption.

Solution: Use 'morley optimize' for the compiled ligo contract, thus
we can test and provide optimized version of baseDAO contract.

## Related issue(s)

Resolves #119 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -42,7 +42,8 @@ steps:
   - label: build-ligo
     if: *not_scheduled
     commands:
-    - nix-build ci.nix -A build-ligo -o baseDAO.tz
+    - nix-build ci.nix -A build-ligo -o baseDAO-non-optimized.tz
+    - nix run -f ci.nix morley -c morley optimize --contract baseDAO-non-optimized.tz --output baseDAO.tz
     artifact_paths:
       - baseDAO.tz
 
--- a/ci.nix
+++ b/ci.nix
@@ -173,4 +173,22 @@ rec {
       };
     });
   });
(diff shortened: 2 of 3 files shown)


pasqu4le (commented):
LGTM, but I'd like the optimization to happen in the `Makefile` as well.
Even tho it's not in the "acceptance criteria" I think it should be done here.

Specifically I'd say that we should make sure that `make test` (in `ligo/`) uses the optimized contract.
I think it would be better for the default `make` to not optimize it (so one can see the differences more easily if he makes changes) or to compile the non-optimized and optimized version separately.

rvem:
That makes sense, will update the Makefile as well

rvem:
@pasqu4le, I've updated `ligo/Makefile`, please check

pasqu4le (approved):
:ok_hand: LGTM.

It seems it managed to remove about ~150 bytes in origination costs :grin:

PR #132, opened 2021-01-19
[#72] Cleanup Config

### LINKED ISSUE #72: Cleanup Config

rinn7e:
## Description

Problem: Currently, `daoName` and `daoDescription` are  used only
for documentation and are not part of `Storage`. We should make it more
clear about that.

Solution: Make it clearer that these values are used only for documentation,
and not part of `Storage`.

## Related issue(s)

Resolves #72 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/Lorentz/Contracts/BaseDAO/Types.hs
+++ b/src/Lorentz/Contracts/BaseDAO/Types.hs
@@ -106,7 +106,11 @@ instance HasAnnotation FA2.Parameter
 
 data Config contractExtra proposalMetadata otherParam = Config
   { cDaoName :: Text
+  -- ^ Name of the DAO.
+  -- Only used in documentation generator and does not get stored in @Storage@.
   , cDaoDescription :: Markdown
+  -- ^ Description of the DAO.
+  -- Only used in documentation generator and does not get stored in @Storage@.
 
   , cProposalCheck :: forall s store.
         StorageC store contractExtra proposalMetadata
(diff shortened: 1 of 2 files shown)


rinn7e:
@gromakovsky Since most of the changes are already made in #98, I'm not sure what else to add beside just adding the comments.

src/Lorentz/Contracts/BaseDAO/Types.hs
```diff
@@ -106,7 +106,11 @@ instance HasAnnotation FA2.Parameter
 
 data Config contractExtra proposalMetadata otherParam = Config
   { cDaoName :: Text
+  -- ^ Name of the DAO.
+  -- Only used in documentation generater and does not get stored in `Storage`.
   , cDaoDescription :: Markdown
+  -- ^ Description of the DAO.
+  -- Only used in documentation generater and does not get stored in `Storage`.
```
gromakovsky, line 113: 1. `s/generater/generator` everywhere.
2. In Haddock backticks should be replaced with single quotes or `@`s.

gromakovsky (approved):
Apparently there is indeed nothing to do besides comments.

Issue #117, opened 2021-01-14
Implement Treasury DAO using LIGO version


gromakovsky:
# Clarification and motivation

Now that we have a LIGO version of BaseDAO we want to demonstrate its usage. Let's implement [TreasuryDAO](https://github.com/tqtezos/baseDAO/blob/master/docs/treasury.md) as an example.

# Acceptance criteria

1. Lambdas that are used to parameterize BaseDAO are implemented (in LIGO) according to the TreasuryDAO specification.
2. There are instructions how to originate LIGO version of TreasuryDAO and these instructions are as simple as possible.

PR #133, opened 2021-01-20
[#117] Implement Treasury DAO using LIGO version 

rinn7e:
## Description
Problem: As an example, we would like to implement Treasury DAO contract
using ligo version of the BaseDAO contract.

Solution: Implement configuration lambdas for BaseDAO contract which
implements Treasury DAO behavior.

## Related issue(s)

Part of #117 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/README.md
+++ b/ligo/README.md
@@ -54,17 +54,41 @@ For more detail about LIGO tests, check out its [README.md](./haskell/test/)
 
 ### RegistryDAO
 
-The storage, including configuration for the RegistryDAO
-[specification](https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md)
-is included in `src/registryDAO.mligo`. You can convert this into a
+The storage, including configuration for the RegistryDAO [specification](https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md)
+is included in `ligo/src/registryDAO.mligo`. You can convert this into a
 Michelson expression during the BaseDAO origination using the `ligo
 compile-storage` command as follows.
 
+```bash
+ligo compile-storage ligo/src/registryDAO.mligo base_DAO_contract \
+    'default_registry_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), 0n, 0n, 0n, 0n, 0n)'
 ```
 
 The `default_registry_DAO_full_storage` is a LIGO function defined in
-`ligo/src/RegistryDAO.mligo`, which returns a LIGO expression for storage,
+`ligo/src/registryDAO.mligo`, which returns a LIGO expression for storage,
 which is converted into Michelson using the `compile-storage` command. The
 arguments to this function are the admin address, the token address and the
-configuration parameters described in the Registry spec, which are 'a', 'b',
-'s_max', 'c' and 'd'.
+configuration parameters described in the Registry spec, which are `a`, `b`,
+`s_max`, `c` and `d`.
+
+### TreasuryDAO
+
+The storage, including configuration for the TreasuryDAO [specification](https://github.com/tqtezos/baseDAO/blob/master/docs/treasury.md)
+is included in `ligo/src/treasuryDAO.mligo`. You can convert this into a
+Michelson expression during the BaseDAO origination using the `ligo
+compile-storage` command as follows.
+
+```bash
+ligo compile-storage ligo/src/treasuryDAO.mligo base_DAO_contract \
+    'default_treasury_DAO_full_storage(("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address), (0n, 0n, 0n, 0n, 0n, 0mutez, 100mutez))'
+```
+
+The `default_treasury_DAO_full_storage` is a LIGO function defined in
+`ligo/src/treasuryDAO.mligo`, which returns a LIGO expression for storage,
+which is converted into Michelson using the `compile-storage` command.
+
+The arguments to this function are:
+- the admin address
+- the token address
+- a tuple value consisting of configuration parameters described in the Treasury spec which are:
+    - `a`, `b`, `s_max`, `c`, `d`, `x`, and `z`.
--- a/ligo/src/helper/unpack.mligo
+++ b/ligo/src/helper/unpack.mligo
@@ -1,0 +1,30 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !HELPER_CONFIG_H
+#define HELPER_CONFIG_H
+
+#include "../types.mligo"
+
+// ------------------------------
+// Unpack Helpers
+// ------------------------------
+
+let unpack_key (key_name, p : string * ((string, bytes) map)) : bytes =
+    match Map.find_opt key_name p with
+      Some (b) -> b
+    | None -> (failwith "UNPACK_KEY_NOT_FOUND" : bytes)
+
+let unpack_nat (key_name, p : string * ((string, bytes) map)) : nat =
+  let b = unpack_key (key_name, p)
+  in match ((Bytes.unpack b) : (nat option)) with
+    Some (v) -> v
+  | None -> (failwith "UNPACKING_NOT_NAT_TYPE" : nat)
+
+let unpack_tez (key_name, p : string * ((string, bytes) map)) : tez =
+  let b = unpack_key (key_name, p)
+  in match ((Bytes.unpack b) : tez option) with
+    Some (v) -> v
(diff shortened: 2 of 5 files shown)


ligo/README.md
```diff
@@ -54,17 +54,41 @@ For more detail about LIGO tests, check out its [README.md](./haskell/test/)
...
+```
+
+The `default_treasury_DAO_full_storage` is a LIGO function defined in
+`ligo/src/treasuryDAO.mligo`, which returns a LIGO expression for storage,
+which is converted into Michelson using the `compile-storage` command.
+
+The arguments to this function are:
+- the admin address
+- the token address
+- a tuple value consists of configuration parameters described in the Treasury spec which are:
```
pasqu4le, line 93:
```suggestion
- a tuple value consisting of configuration parameters described in the Treasury spec which are:
```

ligo/src/treasuryDAO.mligo
```diff
@@ -0,0 +1,116 @@
...
+#include "treasuryDAO/types.mligo"
+
+// -------------------------------------
+// Helpers
+// -------------------------------------
+
+let unpack_transfer_type_list (b: bytes) : transfer_type list =
+  match ((Bytes.unpack b) : ((transfer_type list) option)) with
+    Some (v) -> v
+  | None -> (failwith "UNPACKING_NOT_NAT_OPTION_TYPE" : transfer_type list)
```
sras, line 19: `failwith` message probably needs changing

ligo/src/helper/unpack.mligo
```diff
@@ -0,0 +1,33 @@
...
+// Unpack Helpers
+// ------------------------------
+
+let unpack_key (key_name, p : string * ((string, bytes) map)) : bytes =
+    match Map.find_opt key_name p with
+      Some (b) -> b
+    | None -> (failwith "UNPACK_KEY_NOT_FOUND" : bytes)
+
+let unpack_nat (b: bytes) : nat =
+  match ((Bytes.unpack b) : (nat option)) with
```
pasqu4le, line 19: Small improvement, it seems that all the following `unpack` helpers (including this one and `unpack_transfer_type_list`) actually need `unpack_key` to be used, so why not modify them to include it instead and remove the repetition at the calling side, e.g.:
```suggestion
let unpack_nat (key_name, p : string * ((string, bytes) map)) : nat =
  let b = unpack_key(key_name, p) in
  match ((Bytes.unpack b) : (nat option)) with
```

ligo/src/helper/unpack.mligo
```diff
@@ -0,0 +1,33 @@
...
+    Some (v) -> v
+  | None -> (failwith "UNPACKING_NOT_NAT_TYPE" : nat)
+
+let unpack_nat_option (b: bytes) : nat option =
```
pasqu4le, line 23: This seems to be unused.

ligo/src/treasuryDAO.mligo
pasqu4le, line 19: I think the message was incorrectly copied from `unpack_nat_option` and is not correct.

Also this function could be moved to `helper/unpack.mligo` I suppose.

ligo/src/treasuryDAO.mligo
```diff
@@ -0,0 +1,116 @@
...
+            Some contract ->
+              let xtz_transfer_operation = Tezos.transaction unit xt.amount contract
+              in (is_valid, store, xtz_transfer_operation :: ops)
+          | None ->
+              (false, store, ops)
+          in result
+  in
+  let result = List.fold handle_transfer ts (true, store, ([] : operation list)) in
+  if result.0 then
+    (result.2, result.1)
```
pasqu4le, line 86:
```suggestion
  let (is_valid, store, ops) = List.fold handle_transfer ts (true, store, ([] : operation list)) in
  if is_valid then
    (ops, store)
```

ligo/src/treasuryDAO.mligo
```diff
@@ -0,0 +1,116 @@
...
+let treasury_DAO_proposal_check (params, store : propose_params * storage) : bool =
+  let proposal_size = Bytes.size(Bytes.pack(params.proposal_metadata)) in
+  let frozen_scale_value = unpack_nat(unpack_key("a", store.extra)) in
+  let frozen_extra_value = unpack_nat(unpack_key("b", store.extra)) in
+  let max_proposal_size = unpack_nat(unpack_key("s_max", store.extra)) in
+  let requred_token_lock = frozen_scale_value * proposal_size + frozen_extra_value in
+
+  if (params.frozen_token = requred_token_lock) && (proposal_size < max_proposal_size) then
+    let ts = unpack_transfer_type_list(unpack_key("transfers", params.proposal_metadata)) in
+    let is_all_transfers_valid (is_valid, transfer_type: bool * transfer_type) =
```
sras, line 34: Can't we short circuit if `is_valid` is false ?

src/Lorentz/Contracts/TreasuryDAO/Types.hs
```diff
@@ -27,7 +27,7 @@ data TreasuryDaoContractExtra = TreasuryDaoContractExtra
   , ceSlashDivisionValue :: Natural -- d
   , ceMinXtzAmount :: Mutez -- y
   , ceMaxXtzAmount :: Mutez -- z
-  , ceMaxProposalSize :: Natural --z
+  , ceMaxProposalSize :: Natural -- s_max
```
pasqu4le, line 30: good catch, it seems the same mistake was made in the comment inside the `Default` instance, would you mind fixing that as well?

ligo/src/treasuryDAO.mligo
```diff
@@ -0,0 +1,116 @@
...
+      match transfer_type with
+          Token_transfer_type tt -> is_valid
+        | Xtz_transfer_type xt ->
+            let min_xtz_amount = unpack_tez(unpack_key("x", store.extra)) in
+            let max_xtz_amount = unpack_tez(unpack_key("z", store.extra)) in
```
sras, line 39: Isn't it better to put these two lookup/unpack calls outside of the `is_all_transfers_valid`, since it apepar to remain constant and also appear to be a bit costly.

ligo/src/treasuryDAO.mligo
rinn7e: Yeah, I don't put this in the helper because it is only applicable to treasury dao, whereas `helper` is mean to commonly used function.

ligo/src/treasuryDAO.mligo
rinn7e: We can by using `failwith`, it is just that, I use `bool` so we can handle the failing gracefully. 

Failing gracefully for `proposalCheck` does not really matter, so I'm good with either way.

sras: I didn't get you, can't we just return false if `is_valid` is false, and thus skip the rest of the logic?

ligo/src/treasuryDAO.mligo
```diff
@@ -0,0 +1,121 @@
...
+// -------------------------------------
+// Configuration Lambdas
+// -------------------------------------
+
+let treasury_DAO_proposal_check (params, store : propose_params * storage) : bool =
+  let proposal_size = Bytes.size(Bytes.pack(params.proposal_metadata)) in
+  let frozen_scale_value = unpack_nat("a", store.extra) in
+  let frozen_extra_value = unpack_nat("b", store.extra) in
+  let max_proposal_size = unpack_nat("s_max", store.extra) in
+  let min_xtz_amount = unpack_tez("x", store.extra) in
```
sras, line 31: Shouldn't `min_xtz_amount` be unpacked from `y` key?

ligo/src/treasuryDAO.mligo
rinn7e: I get what mean now, I've pushed the change on this.

ligo/src/treasuryDAO.mligo
rinn7e: Fixed

ligo/src/treasuryDAO.mligo
pasqu4le: Ok, I see, it makes sense to leave it as-is in this case :+1:

ligo/src/treasuryDAO.mligo
```diff
@@ -0,0 +1,121 @@
...
+
+let treasury_DAO_decision_lambda (proposal, store : proposal * storage) : return =
+  let propose_param : propose_params = {
+    frozen_token = proposal.proposer_frozen_token;
+    proposal_metadata = proposal.metadata
+    } in
+  let proposal_key = to_proposal_key (propose_param, proposal.proposer) in
+  let ts = unpack_transfer_type_list("transfers", proposal.metadata) in
+  let handle_transfer (acc, transfer_type : (bool * storage * operation list) * transfer_type) =
+      let (is_valid, store, ops) = acc in
```
sras, line 68: Here also, does it make sense to do the processing if `is_valid` is false here?

rinn7e: I've updated this now.

pasqu4le (approved):

sras (approved):

PR #135, opened 2021-01-21
[#116] Add RegistryDao decision lambda test

### LINKED ISSUE #116: Implement Registry DAO using LIGO version

sras:
Problem: The tests for decision lambda are missing, because it is hard
to check for storage changes since it contains bigmaps.

Solution: Implement a test for decision lambda behavior where we update
s_max value with a proposal, and test it by creating another proposal
which should only work if the `s_max` value was updated.

## Related issue(s)

Resolves #116

- Tests
        silently reappearing again.

- Documentation
    - Haddock


ligo/haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -128,11 +129,62 @@ test_RegistryDAO =
 
               let spent = div (requiredFrozen * c) d
 
-              checkStorage (AddressResolved $ unTAddress consumer) (toVal [[FA2.BalanceResponseItem balanceRequestItem (100 - spent)]])
+              checkStorage (AddressResolved $ unTAddress consumer) (toVal [[FA2.BalanceResponseItem balanceRequestItem (defaultTokenBalance - spent)]])
+
+    , nettestScenarioCaps "checks it correctly calculates executes proposal that has won" $ do
```
pasqu4le, line 134:
```suggestion
    , nettestScenarioCaps "checks it correctly executes the proposal that has won" $ do
```

ligo/haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -128,11 +129,62 @@ test_RegistryDAO =
...
+                    ,([mt|b|], lPackValueRaw @(Maybe Natural) Nothing)
+                    ,([mt|c|], lPackValueRaw @(Maybe Natural) Nothing)
+                    ,([mt|d|], lPackValueRaw @(Maybe Natural) Nothing)
+                    ]
+              let sMaxUpdateproposalSize1 = metadataSize sMaxUpdateproposalMeta1
+              let requiredFrozenForUpdate = sMaxUpdateproposalSize1 * a + b
+
+              callFrom (AddressResolved wallet1) baseDao (Call @"Propose") (ProposeParams requiredFrozenForUpdate sMaxUpdateproposalMeta1)
+
+              -- Then we sent 2 upvotes for the proposal (as max quorum is 2)
```
pasqu4le, line 173:
```suggestion
              -- Then we send 2 upvotes for the proposal (as min quorum is 2)
```

pasqu4le (commented):
Nitpick, but in the commit (and PR) description it says:
> Problem: The tests for decision lambda are missing, because it is hard
to check the storage changes since it contains bigmaps.

I think `to check the storage changes` should be replaced with `to check for storage changes` or `to check that the storage changes`.
Maybe it's fine either way, but I think either of these is easier to read.

pasqu4le (approved):
LGTM, if possible I'd like @rinn7e to review this as well

rinn7e (approved):
Seems good for me!

PR #136, opened 2021-01-22
[#117] Add LIGO Treasury DAO tests

### LINKED ISSUE #117: Implement Treasury DAO using LIGO version

rinn7e:
## Description

Problem: We have implemented Treasury DAO in LIGO, but we have not added
the tests for it yet.

Solution: Add tests for LIGO Treasury DAO.

## Related issue(s)

Resolves #117  

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/Makefile
+++ b/ligo/Makefile
@@ -22,7 +22,7 @@ OUT ?= out
 
 .PHONY: all clean test
 
-all: $(OUT)/baseDAO.tz $(OUT)/registryDAO_storage.tz
+all: $(OUT)/baseDAO.tz $(OUT)/registryDAO_storage.tz $(OUT)/treasuryDAO_storage.tz
 
 # Compile LIGO contract into its michelson representation.
 $(OUT)/baseDAO.tz: src/**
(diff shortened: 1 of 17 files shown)


rinn7e:
The tests of Ligo Treasury DAO pretty much resemble the tests of the Lorentz version.

I've tried to make the shared tests, but since due to conflicting proposal metadata which disallow sharing the tests. Probably martoon knows how to make this work, but he's only holiday currently.

Hope to get some reviews on this.

ligo/Makefile
```diff
@@ -22,7 +22,7 @@ OUT ?= out
 
 .PHONY: all clean test
 
-all: $(OUT)/baseDAO.tz $(OUT)/registryDAO_storage.tz
+all: $(OUT)/baseDAO.tz $(OUT)/registryDAO_storage.tz $(OUT)/treasuryDAO_storage.tz
```
pasqu4le, line 25: As we did for registryDAO, now that we have a `make` way to create this, I think it should be used in `ligo/README.md` instead of `compile-storage` directly.

ligo/haskell/test/Test/Ligo/TreasuryDAO.hs
```diff
@@ -0,0 +1,222 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+--
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+-- For all the incomplete list pattern matches in the calls to with
+-- withOriginated func
```
pasqu4le, line 6: Nitpick.
```suggestion
-- For all the incomplete list pattern matches in the calls to the
-- 'withOriginated' function
```

ligo/src/treasuryDAO.mligo
```diff
@@ -97,6 +97,9 @@ let treasury_DAO_decision_lambda (proposal, store : proposal * storage) : return
     ([%Michelson ({| { FAILWITH } |} : string * unit -> return)]
         ("FAIL_DECISION_LAMBDA", ()) : return)
 
+let default_entrypoint (params, full_storage : bytes * full_storage): return_with_full_storage =
```
pasqu4le, line 100: AFAIU this exists only to be able to accept tez in transactions, am I correct?

If that's the case I suggest:
1. adding a comment to this function
2. renaming it to something more clear (e.g. `receive_xtz_entrypoint`), after all it now has little to do with `default`
3. change the name that it's saved with accordingly, instead of `default` (e.g. `receive_xtz`)

pasqu4le (commented):
Minor comments, but overall LGTM.

I don't know how to avoid the duplication and share the tests, but probably it's not worth spending more time on it since we'll likely not extend the Lorentz version much anymore, so they'll very possibly diverge anyway.

ligo/Makefile
rinn7e: I've updated this now.

ligo/haskell/test/Test/Ligo/TreasuryDAO.hs
rinn7e: This is actually not needed, so I removed this.

ligo/src/treasuryDAO.mligo
rinn7e: Yeah that's right. In Lorentz, this is done via the default entrypoint, since we can't do that in LIGO, it makes sense to rename this properly. I've updated this change now, you can check it.

ligo/haskell/test/Test/Ligo/TreasuryDAO.hs
pasqu4le: Ah, even better :ok_hand:

ligo/src/treasuryDAO.mligo
pasqu4le: Ok cool, still please add a comment to it.

The name is now clear, but I'd still like a sentence to say why we need it in the first place, aka to receive `xtz` because this is not allowed in most standard entrypoints.

pasqu4le (approved):
I left a comment thread open, but I'll approve as it's a minor thing.

Feel free to squash and merge with that.

Issue #129, opened 2021-01-18
Make management entrypoints' tests shared between Ligo and Lorentz


pasqu4le:
# Clarification and motivation

We have some shared tests for the Ligo and Lorentz version for most entrypoints, but not for the "management" ones implemented for #100 (see [this discussion](https://github.com/tqtezos/baseDAO/pull/111#discussion_r557406874)).

The difficult part was mostly due to the different errors between the Ligo and Lorentz version, now that #106 has been completed however, we should try to move as many tests to be shared to reduce duplication.

# Acceptance criteria

As many tests as possible for management entrypoints are `ShareTest` and used in both the Lorentz and Ligo version of the contract.

PR #143, opened 2021-01-25
[#131] Remove `FAILWITH` `Pair "string" Unit` from LIGO 

### LINKED ISSUE #131: Stop adding Units into errors

In many places we fail with `Pair "string" Unit`. In case of the Lorentz version, it's standard behavior which I think is not justified and should be changed in Lorentz. See https://gitlab.com/morley-framework/morley/-/issues/383. In the LIGO version, we do it this way to be consistent with Lorentz.

These Units don't add any useful information, but increase contract size and gas costs. We can remove them from both versions at once (perhaps doing https://gitlab.com/morley-framework/morley/-/issues/383 first) or remove it from the LIGO one first and then wait for #383. In the latter case, we will have to temporarily accept both `Pair "string" Unit` and just `"string"` in tests.

# Acceptance criteria

We never fail with `Pair string unit` type, all such occurrences (in all versions of contracts) are replaced with just `string`.

rinn7e:
## Description

Problem: In many places we fail with `Pair "string" Unit`.
In case of the Lorentz version, it's standard behavior which
is not really justified and should be changed in Lorentz.

Solution: Use a `failAsText` helper in Lorentz to no longer fail with
`Pair "string" Unit`, refactor LIGO code to use standard `failwith`
where it can, make adjustment to all the tests.

## Related issue(s)

Resolves #131 
Depends on #129 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/BaseDAO/ShareTest/Common.hs
+++ b/src/BaseDAO/ShareTest/Common.hs
@@ -48,10 +48,30 @@ type IsLorentz = Bool
 
 -- | Used for check error from ligo contract
 expectFailed
- :: forall caps base m
-  . (MonadNettest caps base m)
-  => Address -> MText -> m () -> m ()
-expectFailed addr val = flip expectFailure (NettestFailedWith addr val)
+ :: forall caps base m a
+  . (MonadNettest caps base m, HasCallStack)
+  => Address -> MText -> m a -> m ()
+expectFailed addr val = withFrozenCallStack $
+  flip expectFailure (NettestFailedWith addr val)
+
+-- -- | Used for check error from ligo contract
+-- expectFailed
+--  :: forall caps base m
+--   . (MonadNettest caps base m, HasCallStack)
+--   => Address -> MText -> m () -> m ()
+-- expectFailed addr val ma =
+--   expectFailed "FA2_TOKEN_UNDEFINED ma
+--     `catch`
+--       (\_ -> expectFailedText addr val ma)
+  -- let mb = expectFailed "FA2_TOKEN_UNDEFINED ma
+  -- in withFrozenCallStack $
+  --     try @_ @SomeException mb >>= \case
+  --       Right _ -> mb
+  --       Left _ -> expectFailedText addr val ma
+
+-- expectStringError :: _
+-- expectStringError val act =
+
 
 -- | Helper functions which originate BaseDAO with a predefined owners, operators and initial storage.
 -- used in FA2 Proposals tests.
--- a/src/Lorentz/Helper.hs
+++ b/src/Lorentz/Helper.hs
@@ -1,0 +1,13 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Simplest DAOs on top of BaseDAO.
+module Lorentz.Helper
+  ( failAsText
+  ) where
+
+import Lorentz
+
+failAsText :: Label name -> s :-> t
+failAsText label =
+  failUsing (errorTagToMText label)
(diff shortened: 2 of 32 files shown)


rinn7e:
I've remove the `Pair "string" Unit` error in LIGO now. Due to not being able to catch error in cleveland nettest: https://serokell.slack.com/archives/C019PCGKTK2/p1611312823001400, I made the failwith in Lorentz version to fail with just string as well. This is a lot easier since when Lorentz gets updated to fail with just a string by default, we simply replace the function name with the new one. 

@pasqu4le Hope to get your review on this as well.

src/BaseDAO/ShareTest/Common.hs
```diff
@@ -48,10 +48,30 @@ type IsLorentz = Bool
...
+--     `catch`
+--       (\_ -> expectFailedText addr val ma)
+  -- let mb = expectFailed "FA2_TOKEN_UNDEFINED ma
+  -- in withFrozenCallStack $
+  --     try @_ @SomeException mb >>= \case
+  --       Right _ -> mb
+  --       Left _ -> expectFailedText addr val ma
+
+-- expectStringError :: _
+-- expectStringError val act =
```
pasqu4le, line 73: These look like remains of an abandoned approach, I suppose they can be removed.

src/Lorentz/Helper.hs
```diff
@@ -0,0 +1,13 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Simplest DAOs on top of BaseDAO.
```
pasqu4le, line 4: This needs an update

src/BaseDAO/ShareTest/Common.hs
```diff
@@ -48,10 +48,30 @@ type IsLorentz = Bool
...
 -- | Used for check error from ligo contract
 expectFailed
- :: forall caps base m
-  . (MonadNettest caps base m)
-  => Address -> MText -> m () -> m ()
-expectFailed addr val = flip expectFailure (NettestFailedWith addr val)
+ :: forall caps base m a
+  . (MonadNettest caps base m, HasCallStack)
+  => Address -> MText -> m a -> m ()
+expectFailed addr val = withFrozenCallStack $
```
pasqu4le, line 54: I may be wrong, but it seems to me that everywhere this is called the first argument is always made with `toAddress`, since this is an helper function to reduce duplication, why not put `toAddress` in it an remove it from the caller?

pasqu4le (commented):
I think the title and the commit message should reflect that the also Lorentz has been stripped from the `unit`s in its errors.

However aside from that I see that you went with the solution of updating Lorentz as well (due to the inability to accept both type of errors), but why didn't you try to solve the linked [morley issue](https://gitlab.com/morley-framework/morley/-/issues/383) as well?

I ask mainly because at its current state the solution seems half-away there but not very robust as it introduces the approach based on helpers that `morley`/`lorentz` know nothing about.
I think the morley issue could be revisited (I will comment there now), but it should be done before updating this PR. What do you think?

rinn7e:
> However aside from that I see that you went with the solution of updating Lorentz as well (due to the inability to accept both type of errors), but why didn't you try to solve the linked [morley issue](https://gitlab.com/morley-framework/morley/-/issues/383) as well?

From what I understand from reading the discussion on the link issues, it seems like catching the error will be possible after [!722](https://gitlab.com/morley-framework/morley/-/merge_requests/722) and [!755](https://gitlab.com/morley-framework/morley/-/merge_requests/755). 

Should I put this MR on hold for now, and wait until those MRs get merged?

pasqu4le:
Yes, I think we should put this on hold for a moment and address the morley issue first (I commented on it, please check it out).

rinn7e:
Close in favor of https://github.com/tqtezos/baseDAO/pull/147

Issue #124, opened 2021-01-16
Enable network tests for LIGO on CI


gromakovsky:
# Clarification and motivation

We pass `--nettest-no-run-network` in `ligo-test` on CI. That's because our LIGO contract is too big and can't be originated easily. See also #123, we likely need to do it first.

# Acceptance criteria

CI runs tests for the LIGO contract with `--nettest-run-network` enabled.

rvem:
I've checked ligo network tests in #148. The origination works now, but most of the tests fail with insufficient xtz error. Also, tests take ~30 to finish on a single local-chain. So, apparently, we should consider adding batching before adding such long-running tests to CI

pasqu4le:
Additional note: running locally after #148 it seems that the additional funds assigned to `admin` in `Test.Ligo.BaseDAO.Common`'s `defaultStartupContract` should be increased to check for a minimum of `4 XTZ`.

pasqu4le:
Addition, with #163 we use `morley-large-originator` to both Lorentz and Ligo versions, so this now extends to reactivating both's nettests.

Note that with those changes the Ligo origination/startup time has gone down a lot, as there's now the need for ~4 operations compared to the previous ~20.
However this still hold unacceptable times and will likely need upstream tooling improvements.

PR #144, opened 2021-01-25
[#123] Reduce origination size and provide a simple guide for it

### LINKED ISSUE #123: Simplify LIGO version deployment

pasqu4le:
## Description

The first two commits of this PR are an attempt to reduce the cost without being to aggressive.
The refactoring that resulted from them (mostly the fist) has been quite successful, but it's still not nearly enough to fit into the limits.

The commit that follows (still a WIP because of missing tests and docs update) solves the origination problem by moving the entrypoint code from the `code` section to lambdas stored in the `storage` and provides a way to generate the parameter needed to load these lambdas into an originated contract

## Related issue(s)

Resolves #123

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification-ligo.md
+++ b/docs/specification-ligo.md
@@ -53,10 +53,6 @@ For example a "treasury" style DAO would have a `proposal_metadata` containing:
 ```
 and an empty `contract_extra`.
 
-Lastly, there is one more `(string, bytes) map` type synonym: `custom_entrypoints`,
-used to execute arbitrary logic on the contract, see [its section](#custom-entrypoints)
-for more information on its content and usage.
-
 DAO configuration value parameters are captured by the `config` type:
 
 ```ocaml
@@ -84,18 +80,52 @@ type config =
   // ^ Determine the maximum value of voting period that is allowed to be set.
   ; min_voting_period : nat
   // ^ Determine the minimum value of voting period that is allowed to be set.
-
-  ; custom_entrypoints : custom_entrypoints
-  // ^ Packed arbitrary lambdas associated to a name for custom execution.
   }
 ```
+
+## Startup Configuration
+
+As a measure to fit into the origination operation limits, the contract `code`
+does not contain the entrypoints logic, instead these (except for one) are
+stored as lambdas in the storage.
+
+To support loading the entrypoints in the contract as well as other initial
+operations, the contract starts in a startup mode and contains a `startup`
+entrypoint (which is the one exception from above) to manage the transition.
+
+Here it's only possible to:
+1. use `startup` to add or remove entrypoints from the storage.
+2. call any endpoint, but only if the `sender` is `admin`.
+3. use `startup` to exit the startup mode, which is irreversible.
+
+To track this functionality we have `startup_storage` defined as:
+```ocaml
+type storable_entrypoint = (bytes * (storage * config)) -> (operation list * storage)
+
+type stored_entrypoints = (string, storable_entrypoint) big_map
+
+type startup_storage =
+  { starting_up : bool
+  // ^ flag, when 'true' the contract is in startup mode
+  ; stored_entrypoints : stored_entrypoints
+  // ^ big_map containing all the entrypoints with their parameter 'pack'ed
+  }
+```
+which constitutes the final piece of our overall storage.
+
+Note that the `stored_entrypoints` `big_map` includes both the standard entrypoint
+as well as any custom ones that one might use, see [its section](#custom-entrypoints)
+for more information on the usage of the latter.
 
 Note:
 - the `token_metadata` type matches the one defined in FA2.
 - the `proposal` type is defined below.
-- `storage` is the storage type of the contract without the configuration.
-- `full_storage` is instead the full storage of the contract, including its configuration,
-which is to say: `type full_storage = storage * config`.
+- `storage` is the storage type of the contract without the `config` and `startup_storage`.
+- stored entrypoints have access to both `storage` and `config` but can only
+modify the `storage`.
+- `full_storage` is instead the full storage of the contract, including its
+configuration and startup values, which is to say:
+`type full_storage = startup_storage * (storage * config)`.
 
 
 ```ocaml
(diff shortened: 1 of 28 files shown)


ligo/README.md
```diff
@@ -28,18 +25,60 @@ which will save the compiled contract in `out/baseDAO.tz`.
...
+
+All the standard entrypoint lambdas are provided in `ligo/src/startup.mligo`
+(with some more information) and can be complied into parameters to push them to
+the contract with the `ligo compile-parameter` command.
+
+This is also included in the `Makefile` and using `make all` will provide each
+parameter in the `out/entrypoints` directory.
+
+For example, to load (or override) an entrypoint into the storage, you can use:
+`tezos-client transfer 0 from alice to myDAO --arg <ep_parameter>`
```
rinn7e, line 69: I think we should provide a full example, since it is a bit hard to understand what needs to be done here:


Example: To load the `burn` entrypoint, run this command: 
(the byte can be found in `out/entrypoints/burn.tz`)
```
tezos-client transfer 0 from alice to myDAO --arg (Left (Some (Pair "burn"
                  (Some 0x05020000023f03210316050d0765046e000000062566726f6d5f076504620000000925746f6b656e5f696404620000000725616d6f756e74072f02000000210320074303680100000014454e545259504f494e545f4e4f545f464f554e44032702000001e0034c0317031603210316031603160316034803190325072c0200000000020000001e03200743036c030b0743036801000000094e4f545f41444d494e034203270321031603160317031605700002032105710003031703160342057000020321057100030316057000030317031703420342032103170317034c032105710002031703160570000203210571000303160317057000030316031605700003032105710004057000030321057100040570000303210571000403420329072f020000005b034c0320034c0320074303620000034c03210571000203190325072c020000000203200200000033034c0320074303620000034c03420743036801000000184641325f494e53554646494349454e545f42414c414e4345034203270200000077034c032105710002034c032105710002034b0356072f020000003b057000020320057000020320057000020320034c03420743036801000000184641325f494e53554646494349454e545f42414c414e434503420327020000001c034c0320034c032005700003034c0346057000030570000303420350034c0321057100020317057000020321057100030316031705700003032105710004031603160317031705700003034205700003031603160316034203420342053d036d0342))))
```

rinn7e (changes requested):

ligo/README.md
```diff
@@ -28,18 +25,60 @@ which will save the compiled contract in `out/baseDAO.tz`.
...
+which can also be used to load custom entrypoints.
+
+As long as "startup" mode is going on only the `admin` can call a stored
+entrypoint.
+
+To remove a previously loaded entrypoint you can instead use
+`tezos-client transfer 0 from alice to myDAO --arg '(Left (Some (Pair "<ep_name>", None)))'`
```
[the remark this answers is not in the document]
pasqu4le: Yes that sounds like a good idea, I think we can also provide the parameters to load the entrypoints this way too.

ligo/README.md
pasqu4le: I think that if we provide a full example we should use a very small custom entrypoint, because having a very long sequence of bytes isn't very useful for documentation.

Additionally I think I should add an example of calling this without fake `<ep_parameter>`, but using bash to `cat` an actual entrypoint. How does that sound?

pasqu4le:
@rinn7e I completed the test implementation and addressed your comments in a new commit.

I'll try to run all the tests in nettest and close #124 as well, but otherwise no more changes are expected for this PR.

@sras I'd also like you to review it as well.

ligo/README.md
rinn7e: Yeah, I'm good either ways with using a small entrypoint or `cat`.

rinn7e (approved):
Seems good for me!

docs/specification-ligo.md
```diff
@@ -821,19 +860,46 @@ Parameter (in Michelson):
 
 - For `vote` entrypoint with permit, returns the current suitable counter for constructing permit signature.
 
-## Custom entrypoints
+## Startup mode
 
-BaseDAO allows DAOs to define their own additional entrypoints.
+The contract has a startup mode that can be operated managed and exited at the
```
sras, line 865: >that can be operated managed and exited

Something feels weird here, may be a comma is missing, or one of operated/managed is redundant.

docs/specification-ligo.md
```diff
@@ -821,19 +860,46 @@ Parameter (in Michelson):
...
+beginning of its life-cycle, see [Startup Configuration](#startup-configuration).
+
+To fo this the contract `parameter` has:
```
sras, line 868: >To fo 

Possible typo

ligo/src/proposal.mligo
```diff
@@ -158,9 +155,13 @@ let check_vote_limit_reached
           ("MAX_VOTES_REACHED", ()) : vote_param)
     else vote_param
 
-let vote(votes, config, store : vote_param_permited list * config * storage): return =
+// Note: we require this because we want to put the entrypoint in a lamba and so
+// we need the contract address to be provided (can't use 'SELF_ADDRESS').
```
sras, line 159: > we need the contract address to be provided (can't use 'SELF_ADDRESS').

But why does the Edo document say it is allowed in Lambda?

>Instruction SELF_ADDRESS equivalent to SELF; ADDRESS but allowed in lambdas

https://tezos.gitlab.io/protocols/008_edo.html#michelson

sras (approved):
LGTM except for minor things.

ligo/src/proposal.mligo
pasqu4le: That's correct, but currently LIGO converts it to `SELF; ADDRESS` and so that won't work.

I think we should update this when LIGO is updated after `edo` activation, in the meanwhile I will update this comment.

PR #145, opened 2021-01-28
[#126] Implement `total_supply` off-chain view

### LINKED ISSUE #126: Implement `total_supply` off-chain view

TZIP-12 spec says that `total_supply` SHOULD be provided, so it may be useful to implement it. The tricky part is that we do not track total supply yet.

Note that it is likely that along with off-chain view, an on-chain view will have to be implemented as well because at the moment of writing we don't have a neat way to test an off-chain view against a contract (though such way may be added very soon, track morley-metadata), and also because there is a [suggestion](https://gitlab.com/tzip/tzip/-/issues/52) in `tzip` to use both types of views anyway
 
# Acceptance criteria

* `total_supply` off-chain view is implemented according to TZIP-12
* The new `totalSupply` field in the contract, if added, is covered with appropriate tests.

rinn7e:
## Description

Problem: TZIP-12 specs say that `total_supply` SHOULD be provided but we
do not track total supply yet.

Solution: Track total supply for all the tokens, add off-chain and
on-chain `total_supply` view, and add tests to cover all the cases.

## Related issue(s)

Resolves #126 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
+++ b/ligo/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
@@ -7,24 +7,36 @@ module Test.Ligo.BaseDAO.OffChainViews
 
 import Universum hiding (view)
 
+import qualified Data.Map as M
+import Lorentz (BigMap(..))
 import Named (defaults, (!))
 import Test.Tasty (TestTree)
 
 import Tezos.Address
 
+import BaseDAO.ShareTest.Common (totalSupplyFromLedger)
 import BaseDAO.ShareTest.OffChainViews
 import qualified Ligo.BaseDAO.Types as DAO
 import qualified Ligo.BaseDAO.TZIP16Metadata as DAO
 import qualified Lorentz.Contracts.BaseDAO.TZIP16Metadata as DAO
 
-defaultStorage :: DAO.StorageL
-defaultStorage =
-  DAO.mkStorageL
-  ! #admin (unsafeParseAddress "tz1M6dcor9QNTFr9Ri68cBYvpxrogZaMttuE")
+offChainViewStorage :: DAO.StorageL
+offChainViewStorage =
+  (DAO.mkStorageL
+  ! #admin addr
   ! #extra DAO.dynRecUnsafe
   ! #metadata mempty
   ! defaults
+  ) { DAO.sLedger = bal
+    , DAO.sTotalSupply = totalSupplyFromLedger bal
+    }
+  where
+    addr = unsafeParseAddress "tz1M6dcor9QNTFr9Ri68cBYvpxrogZaMttuE"
+    bal = BigMap $ M.fromList $
+      [ ((addr, DAO.unfrozenTokenId), 100)
+      , ((addr, DAO.frozenTokenId), 200)
+      ]
 
 test_FA2 :: TestTree
 test_FA2 =
-  mkFA2Tests defaultStorage (DAO.mkMetadataSettingsL DAO.defaultMetadataConfig)
+  mkFA2Tests offChainViewStorage (DAO.mkMetadataSettingsL DAO.defaultMetadataConfig)
--- a/ligo/src/permit.mligo
+++ b/ligo/src/permit.mligo
@@ -51,3 +51,13 @@ let get_vote_permit_counter (param, store : vote_permit_counter_param * storage)
     :: ([] : operation list)
   , store
   )
+
+let get_total_supply (param, store : get_total_supply_param * storage) : return =
+  let result = match Big_map.find_opt param.token_id store.total_supply with
+      None ->
+        (failwith("FA2_TOKEN_UNDEFINED") : nat)
+    | Some v -> v in
+  ( Tezos.transaction result 0mutez param.callback
+    :: ([] : operation list)
+  , store
+  )
--- a/ligo/src/proposal.mligo
+++ b/ligo/src/proposal.mligo
@@ -69,16 +69,16 @@ let check_proposal_limit_reached (config, propose_params, store : config * propo
           ("MAX_PROPOSALS_REACHED", ()) : storage)
     else store
 
-let freeze (tokens, addr, ledger : nat * address * ledger): ledger =
-  let ledger = debit_from (tokens, addr, unfrozen_token_id, ledger) in
-  let ledger = credit_to (tokens, addr, frozen_token_id, ledger) in
-  ledger
+let freeze (tokens, addr, ledger, total_supply : nat * address * ledger * total_supply): (ledger * total_supply) =
+  let (ledger, total_supply) = debit_from (tokens, addr, unfrozen_token_id, ledger, total_supply) in
+  let (ledger, total_supply) = credit_to (tokens, addr, frozen_token_id, ledger, total_supply) in
+  (ledger, total_supply)
 
 [@inline]
-let unfreeze (tokens, addr, ledger : nat * address * ledger): ledger =
-  let ledger = debit_from (tokens, addr, frozen_token_id, ledger) in
-  let ledger = credit_to (tokens, addr, unfrozen_token_id, ledger) in
-  ledger
+let unfreeze (tokens, addr, ledger, total_supply : nat * address * ledger * total_supply): (ledger * total_supply) =
+  let (ledger, total_supply) = debit_from (tokens, addr, frozen_token_id, ledger, total_supply) in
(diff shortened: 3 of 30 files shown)


rinn7e:
@pasqu4le I've added the `get_total_supply` off-chain view now, I have to add on-chain view as well to be able to test this. This on-chain view will be removed later when it is possible to test off-chain view against a contract.

The MR is pretty much ready for review, only a few `TODO` related to documentation are left. Hope to get your review on this.

pasqu4le:
I think we shouldn't necessarily remove the on-chain view for now.

The issue points at a [suggestion](https://gitlab.com/tzip/tzip/-/issues/52) that is still open, but aside from it if it doesn't create problems we can keep it for the extent of this issue and decide later if we want to remove it.

rinn7e:
> I think we shouldn't necessarily remove the on-chain view for now.
> 
> The issue points at a [suggestion](https://gitlab.com/tzip/tzip/-/issues/52) that is still open, but aside from it if it doesn't create problems we can keep it for the extent of this issue and decide later if we want to remove it.

I've removed the comment mentioning about removing the on-chain view. 

I also handle various TODO and add proper doc.

rinn7e:
I'll wait for #144 to be merged first before continue working on this to avoid merged conflict.

rinn7e:
@pasqu4le I've rebase with the latest, and also fixing some issues that were not working before. This should be ready for review.

ligo/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
```diff
@@ -7,23 +7,35 @@ module Test.Ligo.BaseDAO.OffChainViews
...
   ! defaults
+  ) { DAO.sLedger = bal
+    , DAO.sTotalSupply = getTotalSupplyFromLedger bal
+    }
+  where
+    addr = unsafeParseAddress "tz1M6dcor9QNTFr9Ri68cBYvpxrogZaMttuE"
+    bal = BigMap $ M.fromList $
+      [ ((addr, DAO.unfrozenTokenId), 100)
+      , ((addr, DAO.frozenTokenId), 200)
+      ]
```
pasqu4le, line 38: Why the addition of this address balance(s) here?

I'd imagine that in a `defaultStorage` the ledger should be empty (as it was before)

ligo/src/permit.mligo
```diff
@@ -51,3 +51,13 @@ let get_vote_permit_counter (param, store : vote_permit_counter_param * storage)
...
+
+let get_total_supply (param, store : get_total_supply_param * storage) : return =
+  let result = match Big_map.find_opt param.token_id store.total_supply with
+      None ->
+        (failwith("FA2_TOKEN_UNDEFINED") : nat)
+    | Some v -> v in
+  ( Tezos.transaction result 0mutez param.callback
+    :: ([] : operation list)
+  , store
+  )
```
pasqu4le, line 63:
```suggestion
  match Big_map.find_opt param.token_id store.total_supply with
    | None -> (failwith("FA2_TOKEN_UNDEFINED") : return)
    | Some v ->
      ( Tezos.transaction result 0mutez param.callback
        :: ([] : operation list)
      , store
      )
```
I'm surprised that the `:: ([] :operation list)` works BTW, shouldn't `Tezos.transaction` be wrapped in `[]`?

ligo/src/proposal.mligo
```diff
@@ -261,12 +264,11 @@ let unfreeze_proposer_and_voter_token
...
   let tokens = check_balance_less_then_frozen_value
               (tokens, proposal.proposer, proposal, proposal_key, store) in
-  let ledger = unfreeze(tokens, proposal.proposer, store.ledger) in
+  let (ledger, total_supply) = unfreeze(tokens, proposal.proposer, store.ledger, store.total_supply) in
   // unfreeze_voter_token
-  let do_unfreeze = fun (ledger, (addr, tokens) : ledger * (address * nat)) ->
-    unfreeze(tokens, addr, ledger)
-    in
-  {store with ledger = List.fold do_unfreeze proposal.voters ledger}
+  let do_unfreeze = fun ((ledger, total_supply), (addr, tokens) : (ledger * total_supply) * (address * nat)) -> unfreeze(tokens, addr, ledger, total_supply) in
```
pasqu4le, line 269: Let's split this in more lines

pasqu4le, line 269: Let's split this in more lines

pasqu4le, line 269: Let's split this in more lines

ligo/src/token/fa2.mligo
```diff
@@ -26,27 +22,46 @@ let check_sender (from_ , store : address * storage): address =
...
+                (match Michelson.is_nat (bal - amt) with
+                  Some new_bal ->
+                    let ledger = Big_map.update (from_, token_id) (Some new_bal) ledger in
+                    let total_supply = Map.update token_id (Some new_total_supply) total_supply in
+                    (ledger, total_supply)
+                | None ->
+                    ([%Michelson ({| { FAILWITH } |} : string * (nat * nat) -> (ledger * total_supply))] ("FA2_INSUFFICIENT_BALANCE", (amt, bal)) : (ledger * total_supply))
+                )
+            | None ->
+                (failwith("NEGATIVE_TOTAL_SUPPLY") : (ledger * total_supply))
```
pasqu4le, line 42: I think this should be splitted in steps, like:
- first `current_total_supply = ...` (put 0 in case of `None`)
- then `new_total_supply = ...` (fail on `None`)
- etc.

instead of indenting it so much.

src/BaseDAO/ShareTest/Common.hs
```diff
@@ -53,6 +54,14 @@ expectFailed
   => Address -> MText -> m () -> m ()
 expectFailed addr val = flip expectFailure (NettestFailedWith addr val)
 
+getTotalSupplyFromLedger :: Ledger -> TotalSupply
```
pasqu4le, line 57: Nitpick, but `get` seems to indicate some fetching, I suggest renaming this just to `totalSupplyFromLedger`

ligo/src/startup.mligo
```diff
@@ -100,6 +100,19 @@ let storable_get_vote_permit_counter (packed_param, config_store : bytes * confi
     end
   | None -> (failwith "ENTRYPOINT_NOT_FOUND" : return)
 
+let storable_get_total_supply (packed_param, config_store : bytes * configured_storage) : return =
```
pasqu4le, line 103: This should be mentioned as one of the entrypoints that doesn't directly traslate in the "module" comment on top.

pasqu4le (commented):
A few suggestions, also the specifications docs need to be updated as well

ligo/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
rinn7e: This `defaultStorage` is used to test FA2 off-chain views:

```
  , testGroup "get_total_supply" $
    let checkTotalSupply st param =
          runView @Natural (DAO.getTotalSupplyView mc) st (ViewParam param)
    in
    [ testCase "Get unfrozen token total supply" $
        checkTotalSupply storage DAO.unfrozenTokenId
          @?= Right 100
    , testCase "Get frozen token total supply" $
        checkTotalSupply storage DAO.frozenTokenId
          @?= Right 200
    ]
```

Maybe renaming this to something like offchainViewStorage would be better.

ligo/src/permit.mligo
rinn7e: Actually `::` here is a con, not type specifying. Could be quite confusing with the haskell syntax.

ligo/src/startup.mligo
rinn7e: I've updated this, as well as other changes, this should be good now.

ligo/src/permit.mligo
pasqu4le: Oh yeah :sweat_smile: syntax confusion

ligo/src/proposal.mligo
pasqu4le: I have no idea why this comment repeated 3 times, sorry about that

pasqu4le (commented):
LGTM now, except for the specification that still need to be updated.

rinn7e:
> LGTM now, except for the specification that still need to be updated.

I've updated both lorentz and ligo specification now. Please take a look, Pasquale.

pasqu4le (approved):

PR #147, opened 2021-02-08
[#131] Use `NoErrorArg` and remove `FAILWITH` `Pair "string" Unit` from LIGO

### LINKED ISSUE #131: Stop adding Units into errors

rinn7e:
## Description
Problem: In many places we fail with `Pair "string" Unit`, due to the
requirement in older Lorentz version. With the new Lorentz version, we
can fail with a normal string now.

Solution: Bump to latest Lorentz version to be able to use `NoErrorArg`, 
update LIGO code to fail with a normal string where possible. 

## Related issue(s)

Resolves #131 
Depends on #144 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/haskell/test/Test/Ligo/RegistryDAO.hs
+++ b/ligo/haskell/test/Test/Ligo/RegistryDAO.hs
@@ -243,4 +243,4 @@ test_RegistryDAO =
 expectFailProposalCheck
   :: (MonadNettest caps base m)
   => m a -> m ()
-expectFailProposalCheck = expectCustomError_ #fAIL_PROPOSAL_CHECK
+expectFailProposalCheck = expectCustomErrorNoArg #fAIL_PROPOSAL_CHECK
--- a/ligo/haskell/test/Test/Ligo/TreasuryDAO.hs
+++ b/ligo/haskell/test/Test/Ligo/TreasuryDAO.hs
@@ -65,7 +65,7 @@ validProposal = uncapsNettest $ withFrozenCallStack do
 
   withSender (AddressResolved owner1) $
     call dao (Call @"Propose") (ProposeParams (proposalSize + 1) proposalMeta)
-    & expectCustomError_ #fAIL_PROPOSAL_CHECK
+    & expectCustomErrorNoArg #fAIL_PROPOSAL_CHECK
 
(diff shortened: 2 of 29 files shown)


stack.yaml
```diff
@@ -7,5 +7,11 @@ packages:
   - .
   - template
   - ligo/haskell
+  - ../morley/code/morley
+  - ../morley/code/lorentz
+  - ../morley/code/cleveland
+  - ../morley/code/morley-client
+  - ../indigo
+  - ../morley-ledgers/code/morley-ledgers
```
rinn7e, line 15: This MR cannot be merged until latest version of `morley`, `indigo`, `morley-ledgers`, and MR #144 are merged.

rinn7e: To be specific, these are the MRs:

- lorentz: https://gitlab.com/morley-framework/morley/-/merge_requests/759
- indigo: https://gitlab.com/morley-framework/indigo/-/merge_requests/15
- morley-ledgers: https://gitlab.com/morley-framework/morley-ledgers/-/merge_requests/33

rinn7e:
To be specific, these are the MRs that it depends on:

- lorentz: https://gitlab.com/morley-framework/morley/-/merge_requests/759
- indigo: https://gitlab.com/morley-framework/indigo/-/merge_requests/15
- morley-ledgers: https://gitlab.com/morley-framework/morley-ledgers/-/merge_requests/33

pasqu4le (approved):

PR #149, opened 2021-02-15
[#137] RegistryDAO  proposal type to maintain proposal receiver contract addresses

### LINKED ISSUE #137: Add a new RegistryDAO proposal type that can push values to target contract(s)

We need to extend RegistryDAO with an additional proposal type.
> The proposals with addresses are diff’s to the list of parameter-receiver contract list
> If we want to pin down the diff-type, it can be `pair (list %remove address) (list %add address)` - it’s roughly `updateTransferlist` from TZIP-15

Currently there is no "parameter-receiver contract list", so apparently we should add one.

# Acceptance criteria

1. Both implementations (LIGO and Lorentz) are updated to have the new type of proposal. It should update some list of addresses. Apparently this list should be added to the contract first, it's not clarified yet.
2. Specification of RegistryDAO is updated.
3. Tests for new functionality are present.

sras:
## Description

1. Adds a new proposal types to add/remove from the list of contract addressed stored in the contractExtra field. 
2. Adds a new field to registryDAO's extra data.
3. Refactor some LIGO code that decides on the proposal type.
4. Partially add the TZIP-16 storage view to fetch the stored list of contract addresses.

I got in a bit of a mess, and went in circles for a bit, trying to find a way to include the storage view in the Lorentz version of RegistryDAO contract. The issue is that the contract metadata appear to be exposed to the external world during the printing metadata command of the CLI app only. But right now this only prints the main metadata, and does not print metadata for the derived contracts. 

So right now,  I have just left the metadata for registryDAO with the TZIP-16 view, in code [1], without exposing it. Would include it after the reviews/suggestions. A LIGO version of the view is left as well [2], just in case we decide on how to include it in the contract.

[1] https://github.com/tqtezos/baseDAO/compare/sras/%23137-reg-dao-pm-receiver?expand=1#diff-7a8fd3451f076274d8fe7e7f0cc176f645f6a21720cf188542d899bc744d7a47R232

[2] https://github.com/tqtezos/baseDAO/compare/sras/%23137-reg-dao-pm-receiver?expand=1#diff-cfa339bf11943633d70899871b33165e430650d4c23245b9a565df0624c7cfd3R236

## Related issue(s)

Resolves #137 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -14,6 +14,7 @@ Application-level types can be stored as part of the contract's metadata as spec
 Extra storage data:
 1. `big_map k v` is the actual key-value storage.
 2. `big_map k proposalId` for each `k` stores IDs of the proposal that affected `k` last time.
+3. `set address` is the set of successful-proposal receiver contract addresses.
 
 We store these two `big_map`s separately because we want to include `proposalId` for deleted keys as well.
 Alternatively, we can merge them and store `option v` values.
@@ -26,6 +27,10 @@ Every proposal includes (`proposalMetadata`):
 
 Another type of proposal is used to update parameters specified below,
 it takes 5 `option nat` values (called `s_max`, `a`, `b`, `c` and `d` below).
+
+Yet a third type of proposal is used to update the set of successful-proposal receiver contract
+addresses. This proposal accept a parameter that has two constructors. The proposal can add to,
+or remove from the set of addresses.
 
 Configuration lambdas are implemented as follows:
 
--- a/ligo/src/registryDAO.mligo
+++ b/ligo/src/registryDAO.mligo
@@ -11,30 +11,70 @@
 // storage expression.
 #include "base_DAO.mligo"
 
+(*
+ * Implements logic that differentiates between various proposals, and wraps
+ * the relavant data in appropriate constructors.
+ *)
+let extract_proposal (metadata : proposal_metadata) : proposal_type =
+  match Map.find_opt "agoraPostID" metadata with
+  | Some (packed_b) ->
+      begin
+        match ((Bytes.unpack packed_b) : (nat option)) with
+        | Some (agora_post_id) ->
+            begin
+              match Map.find_opt "updates" metadata with
+              | Some (packed_b) ->
+                  begin
+                    match ((Bytes.unpack packed_b) : (registry_diff option)) with
+                    | Some (diff) -> Normal_proposal diff
+                    | None ->
+                      (failwith "UPDATES_UNPACKING_FAILED" : proposal_type)
+                  end
+              | None -> (failwith "UPDATES_NOT_FOUND" : proposal_type)
+            end
+        | None -> (failwith "UNPACKING FAILED" : proposal_type)
+      end
+  | None ->
+      begin
+        match Map.find_opt "update_receivers" metadata with
+          | Some (packed_b) ->
+              begin
+                match ((Bytes.unpack packed_b) : update_receiver_param option) with
+                  | Some upr_param -> Update_receivers_proposal upr_param
+                  | None -> (failwith "UNPACKING FAILED" : proposal_type)
+              end
+          | None -> Configuration_proposal
+      end
+
 let require_extra_value (key_name, ce : string * contract_extra) : bytes =
     match Map.find_opt key_name ce with
-      Some (packed_b) -> packed_b
+    | Some (packed_b) -> packed_b
     | None -> (failwith "CONFIG_VALUE_NOT_FOUND" : bytes)
 
 let require_unpacked_nat(packed_b: bytes) : nat =
   match ((Bytes.unpack packed_b) : (nat option)) with
-    Some (v) -> v
+  | Some (v) -> v
   | None -> (failwith "UNPACKING_FAILED" : nat)
 
 let require_unpacked_nat_opt(packed_b: bytes) : nat option =
   match ((Bytes.unpack packed_b) : ((nat option) option)) with
-    Some (v) -> v
+  | Some (v) -> v
   | None -> (failwith "UNPACKING_FAILED" : nat option)
 
 let require_unpacked_registry (packed_b: bytes) : registry =
   match ((Bytes.unpack packed_b) : (registry option)) with
-    Some (v) -> v
+  | Some (v) -> v
   | None -> (failwith "UNPACKING_FAILED" : registry)
 
 let require_unpacked_registry_affected (packed_b: bytes) : registry_affected =
   match ((Bytes.unpack packed_b) : (registry_affected option)) with
-    Some (v) -> v
+  |  Some (v) -> v
   | None -> (failwith "UNPACKING_FAILED" : registry_affected)
+
+let require_unpacked_proposal_receivers (packed_b: bytes) : proposal_receivers =
+  match ((Bytes.unpack packed_b) : (proposal_receivers option)) with
+  | Some (v) -> v
+  | None -> (failwith "UNPACKING_FAILED" : proposal_receivers)
 
 let lookup_config (key_name, ce : string * contract_extra) : nat =
     require_unpacked_nat(require_extra_value(key_name, ce))
@@ -98,6 +138,13 @@ let registry_DAO_rejected_proposal_return_value (params, extras : proposal * con
   let slash_division_value = require_config ("d", extras)
   in (slash_scale_value * params.proposer_frozen_token) / slash_division_value
 
+type update_fn = (address * address set) -> proposal_receivers
+
+let update_receivers(current_set, updates, update_fn : proposal_receivers * address list * update_fn) : proposal_receivers =
+  let
+    foldFn (existing, item : proposal_receivers * address) : proposal_receivers = update_fn (item, existing)
+  in List.fold foldFn updates current_set
+
 (*
  * Uses the keys in the proposal metadata to differentiate between configuration
  * proposal and normal proposals. If the map contains `agoraPostID` key, the proposal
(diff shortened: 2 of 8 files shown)


pasqu4le:
> the contract metadata appear to be exposed to the external world during the printing metadata command of the CLI app only.

Yeah, that's correct and AFAIK this was sufficient up until now as there where no difference (aside from configurable ones, e.g. `name` and `description` left to be added manually) between the metadata of the various versions of the contract.

So since we now have "known" differences between the metadatas of `RegistryDAO` and the rest I think we should "expose" this through the CLI as well (especially considering they'll likely get more different as time goes on).

In other words I think that you should add alongside the existing `print-metadata` a new `print-metadata-registry` as well.

Please make a separate commit for this, so it can be easier to follow.

sras:
@pasqu4le I have added the new command to print registryDAO metadata. I think this is ready for another look now.

docs/registry.md
```diff
@@ -27,6 +28,10 @@ Every proposal includes (`proposalMetadata`):
 Another type of proposal is used to update parameters specified below,
 it takes 5 `option nat` values (called `s_max`, `a`, `b`, `c` and `d` below).
 
+Yet a thrid type of proposal is used to update the set of successful-proposal receiver contract
```
pasqu4le, line 31:
```suggestion
Yet a third type of proposal is used to update the set of successful-proposal receiver contract
```
typo

ligo/src/registryDAO/types.mligo
```diff
@@ -10,3 +10,13 @@ type registry_value = string
 type registry = (registry_key, registry_value) map
 type registry_affected = (registry_key, proposal_key) map
 type registry_diff = (registry_key * registry_value option) list
+type proposal_receivers = address set
+
+type update_receiver_param =
+  | Add_receivers of (address list)
+  | Remove_receivers of (address list)
```
pasqu4le, line 17: Looking at the comment in the issue it seems this might be incorrect?

The request (in case of a single proposal) is to have something like:
> One proposal type: receiver_diff : pair (list %remove address) (list %add address)

which sounds like it should be able to support adding and removing in the same proposal.

Although maybe that's not the case, given that if we went with two proposals only one thing at a time would be possible, To be sure I think this should be asked.

src/BaseDAO/CLI.hs
```diff
@@ -192,6 +200,13 @@ serveContractRegistry execName gitRev contracts version =
       PrintMetadata config ->
         putTextLn . decodeUtf8 . encodePretty $
           DAO.knownBaseDAOMetadata (DAO.mkMetadataSettings config)
+      PrintMetadataRegistry config -> let
+        metadataSettings = DAO.MetadataSettings
+          { msConfig = config
+          , mcOperatorsL = sOperatorsL
+          }
+        in putTextLn . decodeUtf8 . encodePretty $
+          (RegistryDAO.knownRegistryDAOMetadata @MText @MText metadataSettings)
```
pasqu4le, line 209: Why not use `DAO.mkMetadataSettings` here as well?
The result should be the same

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -163,6 +223,47 @@ config = DAO.defaultConfig
...
+-- without user's input.
+knownRegistryDAOMetadata
+  :: forall k v. IsoRegistryDaoProposalMetadata k v
+  => DAO.MetadataSettings (RegistryDaoStorage k v)
+  -> Metadata (ToT (RegistryDaoStorage k v))
+knownRegistryDAOMetadata settings = let
+  baseMetadata = DAO.knownBaseDAOMetadata settings
+  baseViews = case getViews baseMetadata of
+    Right v -> v
+    _ -> []
```
pasqu4le, line 238:
```suggestion
  baseViews = fromRight [] $ getViews baseMetadata
```

src/BaseDAO/CLI.hs
sras: I actually tried it, but couldn't get the types to work, as the `mkMetadataSettings` returns value of type `MetadataSettings (Storage () ())` but here we want it to be of type `MetadataSettings (Storage (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v))`

ligo/src/registryDAO/types.mligo
sras: >which sounds like it should be able to support adding and removing in the same proposal.

I am not sure, but It seems that allowing that will result in some ambiguity w.r.t to the order of operations when the both addition and removal lists contain the same address, because right now adding existing items, and removing non-existing ones are no-ops. 

So for every duplicate item in the lists, If we apply addition before removal, then existing copies will be removed. If we apply removal before addition, then the new instance will be added.

We can do some processign or validation on the input, but that will probably increase the size which AFAIU is not desireable.

src/Lorentz/Contracts/RegistryDAO.hs
sras: Updated.

docs/registry.md
sras: Updated

ligo/src/registryDAO/types.mligo
pasqu4le: That's a good point, the outcome would be more problematic :ok_hand:

src/BaseDAO/CLI.hs
pasqu4le: I see, it seems that most functions in `Lorentz.Contracts.BaseDAO.TZIP16Metadata` are polymorphic, but not `mkMetadataSettings`, possibly because it was expected to be the only one.

I think it's worth generalizing it here to be able to produce a `MetadataSettings` for any `store` that satisfies the constraints.

I think it shouldn't be too hard to build the `Lens` there, given that `MetadataSettings` already requires the store to have `"sOperators" := Operator ~> ()`

ligo/src/registryDAO.mligo
```diff
@@ -36,6 +71,11 @@ let require_unpacked_registry_affected (packed_b: bytes) : registry_affected =
     Some (v) -> v
   | None -> (failwith "UNPACKING_FAILED" : registry_affected)
 
+let require_unpacked_proposal_receivers (packed_b: bytes) : proposal_receivers =
+  match ((Bytes.unpack packed_b) : (proposal_receivers option)) with
+    Some (v) -> v
```
pasqu4le, line 76:
```suggestion
  | Some (v) -> v
```

ligo/src/registryDAO.mligo
```diff
@@ -120,29 +160,27 @@ let registry_DAO_decision_lambda (proposal, extras : proposal * contract_extra)
...
     } in
   let proposal_key = to_proposal_key (propose_param, proposal.proposer) in
-  match Map.find_opt "agoraPostID" proposal.metadata with
-  | Some (packed_b) ->
+  match extract_proposal(proposal.metadata) with
+  | Normal_proposal diff ->
+      let ce_after_registry_update = apply_diff(diff, extras) in
+      let ce_after_registry_affected_update =
+        apply_diff_affected(proposal_key, diff, ce_after_registry_update) in
+        (([] : operation list), ce_after_registry_affected_update)
```
pasqu4le, line 168: In most cases we reassing a new value to the same name (to "update" it), I think we should do that here as well:
```suggestion
      let extras = apply_diff(diff, extras) in
      let extras = apply_diff_affected(proposal_key, diff, extras) in
      (([] : operation list), extras)
```

ligo/src/registryDAO.mligo
```diff
@@ -120,29 +160,27 @@ let registry_DAO_decision_lambda (proposal, extras : proposal * contract_extra)
...
-        | None -> (failwith "UNPACKING FAILED" : operation list * contract_extra)
+        match urp with
+        | Add_receivers receivers ->
+            let
+              foldFn (existing, item : proposal_receivers * address) : proposal_receivers = Set.add item existing in
+            let new_set = List.fold foldFn receivers current_set in
+            (([] : operation list), Map.update "proposal_receivers" (Some (Bytes.pack new_set)) extras)
+        | Remove_receivers receivers ->
+            let foldFn (existing, item : address set * address) : address set = Set.remove item existing in
+            let new_set = List.fold foldFn receivers current_set in
```
pasqu4le, line 180: Just to follow up on the comment above, I think it's fine here to distinguish the `new_set` from the `current_set`, just because we don't end up replacing the `current_set` _directly_, but use it in a different way (packing it in this case).

ligo/src/registryDAO.mligo
```diff
@@ -194,3 +233,13 @@ let default_registry_DAO_full_storage (admin, token_address, a, b, s_max, c, d
     decision_lambda = registry_DAO_decision_lambda;
     } in
   (startup, (new_storage, new_config))
+
+let successful_proposal_receiver_view (full_storage : full_storage): proposal_receivers =
```
pasqu4le, line 237: Can you please add a brief comment here, just to point out that this isn't actually used (I'm not sure it will ever be, but for now let's keep it).

ligo/src/registryDAO.mligo
sras: Updated. Here and some other places where this was missing.

ligo/src/registryDAO.mligo
sras: Changed.

src/BaseDAO/CLI.hs
sras: I have included it. Hope it is what you actually wanted.

pasqu4le: Looks good to me, but it seems that the `config` shadows the existing binding and is causing CI to fail.

sras: Have fixed it. But it seems that the contract is oversized now. Trying to fix it.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -139,6 +145,61 @@ decisionLambda = do
...
+              toField #ceProposalReceivers
+
+              stackType @(Set Address : RegistryDaoContractExtra k v : [Address] ': store : _)
+              dip swap
+
+              stackType @(Set Address : [Address] : RegistryDaoContractExtra k v  ': store : _)
+              swap
+              stackType @([Address] : Set Address : RegistryDaoContractExtra k v  ': store : _)
+              iter $ do
+                dip $ push False
```
pasqu4le, line 195: @sras regarding the origination size, it seems that the contract is close to the limit, so probably not much is needed to make it fit.

The first optimization possible is here: this `push` is the only difference between the two branches.
So I suggest you just add a `Bool` and an `[Address]` on the stack in the two case clauses and move the majority of code outside of it. This will cut the code size and may be enough to fit into the limits.

P.s. The same should be done in the Ligo version BTW

sras: This reduced the size quite a lot, but appears that we still need to reduce ~40 bytes more. I ll check it.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -139,6 +145,45 @@ decisionLambda = do
...
+          update
+
+        -- Now we have the updated set at the top. We update the storage with it.
+        stackType @(Set Address ': Bool ': RegistryDaoContractExtra k v ': [Address] ': store : _)
+        dip drop
+        stackType @(Set Address ': RegistryDaoContractExtra k v ': [Address] ': store : _)
+        setField #ceProposalReceivers
+        stackType @(RegistryDaoContractExtra k v ': [Address] ': store : _)
+        dip drop
+        stSetField #sExtra
```
pasqu4le, line 185:
```suggestion
        dip (drop # stGetField #sExtra # getField #ceProposalReceivers)
        stackType @(UpdateReceivers ': Set Address ': RegistryDaoContractExtra k v ': store : _)

        -- Push a boolean value to indicate if we are adding addresses
        -- or removing them. True to add and False to receive.
        caseT
          ( #cAddReceivers /-> push True
          , #cRemoveReceivers /-> push False
          )
        swap
        stackType @([Address] ': Bool ': Set Address ': RegistryDaoContractExtra k v ': store : _)

        -- Now we have the input addresses at the top
        -- The boolean that indicate if we are adding or removing below it
        -- and the current set of addresse, under it.
        iter $ do
          -- We put a copy of the boolean operation indicator on top of the stack
          duupX @2
          dip update
        drop

        stackType @(Set Address ': RegistryDaoContractExtra k v ': store : _)
        setField #ceProposalReceivers
        stSetField #sExtra
```
I'm not sure how much of this would be optimized away by the optimizer anyway, but since you are so close this rearranging may be enough.

sras: Had to optimized a couple of more bytes, but it appear to be withing the limits now. Tests passed now.

pasqu4le (approved):
LGTM now.

Please check the PR checklist boxes, AFAICT they are all solved

PR #153, opened 2021-02-22
Remove 007 tests

rvem:
## Description
Problem: 007 protocol is now obsolete.

Solution: Remove 007-related tests from CI.

## Related issue(s)

https://issues.serokell.io/issue/TM-513

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -68,9 +68,9 @@ steps:
       - nix-build ci.nix -A packages.baseDAO.tests.baseDAO-test
       - ./result/bin/baseDAO-test --nettest-no-run-network
 
-  - label: test-local-chain-007
+  - label: test-local-chain-008
     env:
-      TASTY_NETTEST_NODE_ENDPOINT: "http://localhost:8735"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -151,18 +126,6 @@ steps:
...
-  #   agents:
-  #     queue: "scheduled"
-  #   commands: *ligo-nettest
-  #   retry:
-  #     automatic:
-  #       limit: 1
-
   # - label: scheduled edonet ligo test
   #   if: build.source == "schedule"
   #   env:
```
pasqu4le, line 131: (one line below, GH doesn't allow me to comment)

If I'm not mistaken I think the port should be changed here as well to `8733`.

Won't make a difference here, but could save some trouble when it will be uncommented.

pasqu4le (approved):

.buildkite/pipeline.yml
rvem: Why? Our edonet node has `8732` as an RPC port

pasqu4le: Oh right, sorry, got confused a second with local chains

PR #154, opened 2021-02-23
[#141] Typescript sdk

### LINKED ISSUE #141: Implement typescript SDK

We need to implement a typescript SDK `basedao.ts` to interact with the contract.
The requirements are as follows:
> - NPM install documentation, with a simple working example for every function
> - We need a library with all of the different ways to call the contract such as
> 1. Recognition function ('is this the DAO that I'm looking for?')
> 2. Type definitions for all the parameters
> 3. Individual functions to call each major function (`vote`, `wrap`, `unwrap`, etc.)

Note: it may be too complex to develop this all in the span of a single issue.
If possible we should try to solve this with a single PR and divide the commits in a way that's easy to follow, otherwise make multiple PRs.

# Acceptance criteria

A `basedao.ts` typescript SDK is implemented and documented.

sras:
## Description

Raising a Draft PR to get feedback on the approach used here. Basicaly we generate some typescript types
from the entrypoint info generated by morley. This PR includes 

1. A couple of Api methods that are implemented using these types 
2. Some sample contract calls to an dummy contract with BaseDAO param.

EDIT: To execute the contract call, cd to `typescript` directory and run `yarn install`, `yarn tsc` and `node dist/index.js`. Also please note that the source files in `typescript/src/generated` folder are autogenerated from entrypoint structure generated by Morley. So it might be missing some formatting.

## Related issue(s)

Resolves #141 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+import qualified Fmt
+import System.FilePath
+
+import Lorentz.Entrypoints
+import Michelson.Typed.Annotation
+import Michelson.Typed.Entrypoints (pnNotes)
+import Michelson.Untyped.Annotation
+
+-- | This is mostly same as the 'Notes t' type, but without the type
+-- parameter and without annotatations that we are not interested.
```
heitor-lassarote, line 20: Replace `annotatations` with `annotations`.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+-- wrap the annotation in a Just before returning it.
+extractAnn :: Annotation tag -> Maybe Text
+extractAnn ann_ =
+  case (ann_ == noAnn) || (isJust $ find (== ann_) ignoreEntypoints) of
+    True -> Nothing
+    _ -> Just $ unAnnotation ann_
+
+-- Some entrypoints we don't want to call directly.
+ignoreEntypoints :: [Annotation tag]
+ignoreEntypoints = [ann "running", ann "transfer_contract_tokens", ann "xtzForbidden", ann "call_FA2", ann "migratable"]
```
heitor-lassarote, line 108: Sorry, but why do we want to ignore those entrypoints?

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+  | TsCustom Typename
+  deriving stock Generic
+
+-- Implements the conversion of Type to its source code representation.
+instance Buildable TsType where
+  build = \case
+    TsNumber -> "number"
+    TsString -> "string"
+    TsBool -> "boolean"
+    TsUnit -> "void"
```
heitor-lassarote, line 138: I have mixed feelings about converting a `unit` to `void`. The reason is that we may have `unit` in a container such as `list` or `map` or even on the argument of a `lambda` and it will become an `undefined` in these situations. I think it's better if we had an empty type for this such as `{}`.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+  } deriving stock Generic
+
+instance Buildable TsModule where
+  build (TsModule name imports decls) = Fmt.unlinesF [build name, build imports, build decls]
+
+type IsOptional = Bool
+
+-- Represents the right side of a Type definition in Typescript.
+data TsTypeDef
+  = TsSum [Typename]
```
heitor-lassarote, line 157: I think `|` and `&` are called union and intersection types in TypeScript (rather than sum and product types like in Haskell)

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+  | TsInterface [(Fieldname, (Typename, IsOptional))]
+  | TsTuple [Typename]
+  | TsAlias TsType
+  deriving stock Generic
+
+-- Implements the conversion to source code.
+instance Buildable TsTypeDef where
+  build (TsSum xs) = "\n" +| (Fmt.indentF 2 $ (Fmt.blockListF' "|" build xs))
+  build (TsTuple fs) =  Fmt.unwordsF ["[", T.intercalate ", " fs, "]"]
```
heitor-lassarote, line 166: If you write something like `["foo", 42`], TypeScript infers it as `(string | number)[]` AFAIK, so I think you need `["foo", 42] as const` so that it's actually a tuple `[string, number]`.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+  writeFile (fp </> filename) (Fmt.pretty $ Fmt.unlinesF (importsLines <> declLines))
+  pure filename
+
+-- Given a typename, and a type, generated Typescript declarations for this
+-- type and all of its fields. Pairs are coverted into Tuples or Interfaces.
+-- If there is atleast one named field in the type, it is converted into
+-- interface.  Or else into a Typescript tuple. 'Or' structures are converted
+-- into Sum types after flattening their structure.  Rest of the types are
+-- converted into corresponding Typescript type after creating type aliased for
+-- their inner fields.
```
heitor-lassarote, line 228: I'll make a suggestion, as I think this is easier to read:

```haskell
-- Given a typename, and a type, generates Typescript declarations for this
-- type and all of its fields. If there is at least one named field in the type, it is
-- converted into an interface, otherwise into a TypeScript tuple. 'Or' structures are
-- converted into Union types after flattening their structure. The rest of the types are
-- converted into the corresponding TypeScript types after creating type aliases for
-- their inner fields.
```

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+  EpChainId -> [TsType typename $ TsAlias TsString]
+  EpMutez -> [TsType typename $ TsAlias TsNumber]
+  EpBool -> [TsType typename $ TsAlias TsBool]
+  EpKeyHash -> [TsType typename $ TsAlias TsString]
+  EpTimestamp -> [TsType typename $ TsAlias TsString]
+
+  where
+
+    hasAtleastOneNamedField :: [(Fieldname, EpType)] -> Bool
+    hasAtleastOneNamedField fs = isJust $ find ((> 0) . length . fst) fs
```
heitor-lassarote, line 323: `length` is O(n), so I'd avoid using it. Suggestion:
```haskell
hasAtLeastOneNamedField = isJust $ find (not . null . fst)
```
n.b.: Notice I changed `Atleast` with `AtLeast` as well.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,380 @@
...
+
+    decls = mkTypesForField <$> (indexEmptyFields $ fields1 <> fields2)
+    typeOptions = (\(_, (b, _), _) ->  b) <$> decls
+      -- Throws away information about if the field is optional. Need to check if it is relavant in 'Or' structures
+    thisType = TsType typename (TsSum typeOptions)
+    thisDecls = (\(_, _, c) -> c) <$> decls
+    in thisType : (concat thisDecls)
+
+  EpList a -> let
+    elementTypeName = typename <> "_" <> "item"
```
heitor-lassarote, line 262: Suggestion: TypeScript normally uses CamelCase for names, maybe it would be better if we had something like `typename <> "Item"` here and in other places?

ligo/haskell/src/Ligo/Typescript.hs
sras: These are not "real" entrypoints, and they have real entrypoints deeper down the tree. 

If you look at our parameter [here](https://github.com/tqtezos/baseDAO/blob/master/ligo/haskell/src/Ligo/BaseDAO/Types.hs#L116), you can see that it is a nested type, instead of a single, flat sum type. And this is so that we can remove duplication in terms of contract code for some checks that otherwise would have to be repeated for each separate entrypoints. Because of this nesting, we can do such checks once, on the top level branch. In the ligo contract, we have manually constructuring these stuctures so that these entrypoints wont show in the actual parameter.

That being said, there is indeed one bug here, which is, `transfer_contract_tokens` should not be ignored here, so I ll fix it.

```
parameter
  (or (option %startup (pair string (option bytes)))
      (or (or (pair %callCustom string bytes)
              (or (or (or (or (unit %accept_ownership)
                              (pair %burn (address %from_) (pair (nat %token_id) (nat %amount))))
                          (or (or %call_FA2
                                 (or (pair %balance_of
                                        (list %requests (pair (address %owner) (nat %token_id)))
                                        (contract %callback
                                           (list (pair (pair %request (address %owner) (nat %token_id)) (nat %balance)))))
                                     (list %transfer
                                        (pair (address %from_)
                                              (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))))
                                 (list %update_operators
                                    (or (pair %add_operator (address %owner) (pair (address %operator) (nat %token_id)))
                                        (pair %remove_operator (address %owner) (pair (address %operator) (nat %token_id))))))
                              (unit %confirm_migration)))
                      (or (or (bytes %drop_proposal) (nat %flush))
                          (or (pair %getVotePermitCounter (unit %param) (contract %callback nat))
                              (pair %get_total_supply (nat %token_id) (contract %callback nat)))))
                  (or (or (or (address %migrate) (pair %mint (address %to_) (pair (nat %token_id) (nat %amount))))
                          (or (pair %propose (nat %frozen_token) (map %proposal_metadata string bytes))
                              (nat %set_quorum_threshold)))
                      (or (or (nat %set_voting_period) (address %transfer_ownership))
                          (list %vote
                             (pair (pair %argument (bytes %proposal_key) (pair (bool %vote_type) (nat %vote_amount)))
                                   (option %permit (pair (key %key) (signature %signature)))))))))
          (pair %transfer_contract_tokens
             (address %contract_address)
             (list %params
                (pair (address %from_)
                      (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount))))))))) ;
```

heitor-lassarote: Ah, I see, thanks for clarifying!

ligo/haskell/src/Ligo/Typescript.hs
sras: I am not sure I get you. This line 

>Fmt.unwordsF ["[", T.intercalate ", " fs, "]"]

is supposed to appear on the right side of a type alias, like [here](https://github.com/tqtezos/baseDAO/blob/sras/%23141-typescript-sdk/typescript/src/generated/CallCustom.ts#L1). Can this be a problem?

heitor-lassarote: Ah, sorry, this is not a problem in this case!

ligo/haskell/src/Ligo/Typescript.hs
sras: I have fixed this and couple of other typos in comments.

ligo/haskell/src/Ligo/Typescript.hs
sras: Have removed the 'transfer_contract_tokens' entrypoint from list of ignored ones.

ligo/haskell/src/Ligo/Typescript.hs
sras: Have replaced it with '{}'.

ligo/haskell/src/Ligo/Typescript.hs
sras: Have replaced it with your suggestion which indeed reads better.

ligo/haskell/src/Ligo/Typescript.hs
sras: :facepalm: Fixed both.

ligo/haskell/src/Ligo/Typescript.hs
sras: Makes sense. Have replaced this.

ligo/haskell/src/Ligo/Typescript.hs
sras: Have renamed `TsSum` to `TsUnion`.

heitor-lassarote:
Thank you for addressing my comments!

typescript/baseDAO/src/api.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  private initContract(): Promise<ContractAbstraction<ContractProvider>> {
+    if (this.contract === undefined) {
+      const Tezos = new TezosToolkit(this.nodeAddr);
+      Tezos.setProvider({ signer: new InMemorySigner(this.senderSk) });
+      this.contract = Tezos.contract.at(this.contractAddr);
+      return this.contract;
+    } else return this.contract;
+  }
```
heitor-lassarote, line 70: Both branches of this method return `this.contract`. I think you may remove the `else` branch and have `return this.contract` just after the `if` statement.

sras: I thought TS is just not smart enough to figure that `this.contract` is not undefined just after the `if` statement, unless it is in a type refining branch of an if/else statement. Trying it anyway..

typescript/baseDAO/src/api.ts
```diff
@@ -0,0 +1,230 @@
...
+        })
+        //.catch((error) => println(`Error: ${JSON.stringify(error, null, 2)}`));
+        .catch((error) => println(`Error: ${error.toString()}`));
+  }
+
+  // entrypoint methods
+  balance_of(arg: Balance_of): Promise<string|void> {
+    return this.withContract((contract) => {
+      return contract.methods.balance_of(arg.requests, arg.callback);
+      });
```
heitor-lassarote, line 105: Nitpick: Just so this module is a bit smaller, when you have something like `(a) => { return b; }`, you may replace it with `a => b`.

typescript/baseDAO/api-extractor.json
```diff
@@ -0,0 +1,366 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
```
sras, line 1: @heitor-lassarote Please ignore this file. On second thought, I think we don't have to include it in the repo.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,440 @@
...
+          in (fn, (inner, False), withDiscriminator)
+
+-- While creating interfaces, if there are fields where no field
+-- name was found, put numeric strings for the fieldname.
+indexEmptyFields :: [(Fieldname, EpType)] -> [(Fieldname, EpType)]
+indexEmptyFields f = zipWith zipFn f ([0..] :: [Int])
+  where
+    zipFn (fn, t) idx = case fn == "" of
+      True -> (show idx, t)
+      _ -> (fn, t)
```
heitor-lassarote, line 408: Suggestion:
```haskell
zipFn (fn, t) idx = (bool fn (show idx) (fn == ""), t)
```

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,440 @@
...
+  | EpKeyHash
+  | EpTimestamp
+  | EpList EpType
+  | EpOption EpType
+  | EpPair AnnotatedField AnnotatedField
+  | EpOr AnnotatedField AnnotatedField
+  deriving stock (Show, Generic)
+
+type AnnotatedField
+ = (FieldAnn, EpType)
```
heitor-lassarote, line 48: I have mixed feelings about the usage of types like this because GHC often trips up and shows error messages like `Expected (FieldAnn, EpType) but got whatever` instead of using `AnnotatedField`. Apart from this, I think a `data` looks better than a `type` aliasing a tuple.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,440 @@
...
+  , tsmDecls :: [TsDecl]
+  , tsmExports :: [Typename]
+  } deriving stock Generic
+
+type IsOptional = Bool
+
+-- Represents the right side of a Type definition in Typescript.
+data TsTypeDef
+  = TsUnion [TsType]
+  | TsInterface [(Fieldname, (TsType, IsOptional))]
```
heitor-lassarote, line 165: I think `(Fieldname, (TsType, IsOptional))` looks a bit unintuitive, I propose to make this tuple as a new datatype.

typescript/baseDAO/src/api.ts
sras: Apparently it works, but I don't understand how. Because this works as well

```
  private initContract(): Promise<ContractAbstraction<ContractProvider>> {
      return this.contract;
  }
```

Why does it allow unification of `undefined | a` with `a` ?

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,440 @@
...
+  build (TsUnion xs) = Fmt.indentF 2 $ (Fmt.blockListF' "|" build xs)
+  build (TsTuple fs) =  Fmt.unwordsF ["[", T.intercalate ", " (Fmt.pretty <$> fs), "];"]
+  build (TsInterface fns) =  Fmt.unlinesF ["{", Fmt.indentF 2 $ Fmt.unlinesF $ mkPair <$> fns, "};"]
+    where
+      mkPair :: (Fieldname, (TsType, IsOptional)) -> Text
+      mkPair (fn, (tn, isOptional)) =
+        case isOptional of
+          True -> fn <> "?: " <> (Fmt.pretty tn) <> ";"
+          False -> fn <> ": " <> (Fmt.pretty tn) <> ";"
+  build (TsAlias t) = build t <> ";"
```
heitor-lassarote, line 181: What version of TypeScript are we targetting? AFAIK newest versions don't require semicolons because of ASI (Automatic Semicolon Insertion), but I don't mind them either.

typescript/baseDAO/src/api.ts
heitor-lassarote: I think it needs to be compiled with `strictNullCheck`, or just the strict mode, from what I've read. Apart from this, I'm not sure why such a thing is allowed...

heitor-lassarote: Yep, just tried it on https://www.typescriptlang.org/play
If you have this code:
```typescript
class Foo {
    private x: number | undefined
    public foo(): number {
        return this.x
    }
}
```

It will accuse an error, unless you disable `strictNullChecks`.

sras: Even after enabling the strict null checks, your suggestion works for me. So it turns out that the Typescript compiler is actually smarter than I imagined, and can actually infer that `this.contract` is not null after the non-null assignment in the `if` block. 

That is reeally impressive.

sras: Have included this change now.

typescript/baseDAO/src/api.ts
sras: Right. Have made this change where possible.

ligo/haskell/src/Ligo/Typescript.hs
sras: Changed.

ligo/haskell/src/Ligo/Typescript.hs
sras: Yes, changed.

ligo/haskell/src/Ligo/Typescript.hs
sras: Changed.

ligo/haskell/src/Ligo/Typescript.hs
sras: Yes, IIRC semicolons were optional in Typescript from the begining. Not sure though. I just included it here since including them appeared to be more common in Typescript world.

heitor-lassarote: Ok, it doesn't return leaving them. I think there are even some cases that leaving them out doesn't do what you expect, for example:
```typescript
return
{
    foo: bar
}
```
Will become:
``` typescript
return;
{
    foo: bar
};
```
Unless the brackets were in the same line as the `return`.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,448 @@
...
+import System.FilePath
+
+import Lorentz.Entrypoints
+import Michelson.Typed.Annotation
+import Michelson.Typed.Entrypoints (pnNotes)
+import Michelson.Untyped.Annotation
+
+-- | This is mostly same as the 'Notes t' type, but without the type
+-- parameter and without annotations that we are not interested.
+data EpType
```
pasqu4le, line 21: There seem to be other partial re-implementation of some `morley` types and functions here.

Have you tried re-using the existing ones instead? If so where there any particular issues?

I'm a bit concerned with maintaining this, especially considering that (very cool BTW) such approach allows for re-generation and it would be a bonus to have it auto-update for the most part.

pasqu4le (commented):
The general approach LGTM.

One thing that I would like to ask for is to document the re-generation of the files as well as adding the procedure to the ligo `Makefile`. 
A CI check would be the next step, but that sounds quite complex, so I'm also ok with adding that in the future.

ligo/haskell/src/Ligo/Typescript.hs
sras: >If so where there any particular issues?

I tried to use `mkEntrypointsMap` from morley that can convert an untyped `Type` to a map of entrypoint -> untyped types. Here we had a typed `Type` and wanted to make entrypoint names out of it. The issue here is that the typed `Type` cannot hold annotations, so they are dervived from the actual type via another typeclass, which appear to be what `parameterEntrypointsToNotes` function is doing. 

We would be able to use the `mkEntrypointsMap` function if we can make a untyped type from a `Notes t` stucture, without going through typed T first, because that would lose all annotations. 

I should be possible to get rid of `EpType` and some other types. ll try to do it.

>I'm a bit concerned with maintaining this

I am not sure if this can be generally implemented for all parameters, but if it is, may be it is worth extracting it to a separate package or include it in an existing one. What do you think?

sras: I have refactored the code to get rid of EpTypes and also ended re-using `mkEntrypointsMap` function from `morley`.

sras:
>One thing that I would like to ask for is to document the re-generation of the files as well as adding the procedure to the ligo Makefile.

I have added a section to `ligo/README.md` file. Also have added a new command to the CLI in `baseDAO-ligo-meta` and have included a make target to generate the Typescript files in`ligo/Makefile`.

ligo/Makefile
```diff
@@ -140,5 +140,10 @@ test: all
     REGISTRY_STORAGE_PATH=../$(OUT)/registryDAO_storage.tz \
     TREASURY_STORAGE_PATH=../$(OUT)/treasuryDAO_storage.tz
 
+typescript: all
```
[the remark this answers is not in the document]
pasqu4le: Also, perhaps this target would be better named `generete-typescript`, what do you think?

ligo/README.md
```diff
@@ -79,6 +79,11 @@ These configurable options are:
 - the `contract_name` is the alias that the originated contract will be associated with.
 - the `destination` is the directory where the steps files will be saved.
 
+## Generating Typescript interface types
```
pasqu4le, line 82: As far as "ligo" is concerned this may be sufficient, however we do need a `README` in the `typescript` directory as well.

It should mention:
1. a short description
2. dependencies: `yarn`, `node`, etc.
3. how to build/install (the info in the PR description seems to be outdated BTW)
4. a list of its features ("recognition" and entrypoints) and an example on how to use each (one entrypoint call is enough)

In short some basic documentation around this SDK

pasqu4le (commented):
I only have a few nitpicks, aside from the documentation request you can squash all the others with the rest of the commits.

ligo/haskell/src/Ligo/Typescript.hs
```diff
@@ -0,0 +1,428 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+module Ligo.Typescript (generateTs) where
```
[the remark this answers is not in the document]
sras: Done.

ligo/README.md
sras: Done.

typescript/baseDAO/README.md
```diff
@@ -0,0 +1,83 @@
...
+### Contract validation
+
+The target BaseDAO contract's storage and parameter can be validated against a known schema to make sure that
+we are talking to a compatible contract.
+
+You can obtain a parameter and storage schema with:
+
+```
+let contractStorageSchema = await BaseDAO.storageSchema();
+let contractParameterSchema = await BaseDAO.storageSchema();
```
pasqu4le, line 73:
```suggestion
let contractParameterSchema = await BaseDAO.parameterSchema();
```

pasqu4le (approved):
One last comment, otherwise LGTM

PR #157, opened 2021-03-03
[#152] Allow specifying a TZIP-16 metadata contract in the LIGO version

### LINKED ISSUE #152: Allow specifying a TZIP16 metadata contract in the LIGO version

The storage for `baseDAO` as well as `RegistryDAO` and `TreasuryDAO` of the LIGO version currently do not accept any metadata parameter.

While this could be provided manually, we should instead support the metadata contract as parameter to the `Makefile` storage generation functions and allow as such for a functionality that's similar to what `tzip16MetadataParser` does on the Lorentz side.

# Acceptance criteria

It's possible to provide a secondary contract metadata in the LIGO storage creation(s).

heitor-lassarote:
## Description

Problem: We currently can't specify any parameters for metadata when
creating a contract with LIGO.

Solution: Have the `default_{registry,treasury}_DAO_full_storage`
functions accept a `(string, bytes) big_map` corresponding to the
metadata. In the Makefile, add `metadata_map` fields which the user can
specify.

## Related issue(s)

Resolves #152

- Tests
  - [X] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/Makefile
+++ b/ligo/Makefile
@@ -64,10 +64,11 @@ endif
 
 $(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
 $(OUT)/trivialDAO_storage.tz : token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage(("$(admin_address)": address), ("$(token_address)": address))'
+	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage(("$(admin_address)": address), ("$(token_address)": address), $(metadata_map))'
 	# ================= Compilation successful ================= #
 	# See "$(OUT)/trivialDAO_storage.tz" for compilation result  #
 	#
@@ -79,10 +80,11 @@ $(OUT)/registryDAO_storage.tz : frozen_extra_value = 0n
 $(OUT)/registryDAO_storage.tz : max_proposal_size = 100n
 $(OUT)/registryDAO_storage.tz : slash_scale_value = 1n
 $(OUT)/registryDAO_storage.tz : slash_division_value = 0n
+$(OUT)/registryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/registryDAO_storage.tz: src/**
 	# ============== Compiling RegistryDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/registryDAO_storage.tz src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("$(admin_address)": address), ("$(token_address)": address), ${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value))'
+	$(BUILD_STORAGE) --output-file $(OUT)/registryDAO_storage.tz src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("$(admin_address)": address), ("$(token_address)": address), ${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value), $(metadata_map))'
 	# ================= Compilation successful ================= #
 	# See "$(OUT)/registryDAO_storage.tz" for compilation result #
 	#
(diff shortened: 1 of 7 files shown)


ligo/src/treasuryDAO.mligo
```diff
@@ -108,8 +101,8 @@ let receive_xtz_entrypoint (params, config_store : bytes * configured_storage)
 // Storage Generator
 // -------------------------------------
 
-let default_treasury_DAO_full_storage (admin, token_address, contract_extra
-    : (address * address * treasury_contract_extra)) : full_storage =
+let default_treasury_DAO_full_storage (admin, token_address, contract_extra, metadata
+    : (address * address * treasury_contract_extra * (string, bytes) big_map)) : full_storage =
```
pasqu4le, line 105: Let's add a `metadata_map` type synonym to `ligo/src/types.mligo` and use that instead.

Also, this change is missing from the `defaults.mligo` as well as the `trivialDAO_storage` in the `Makefile` that uses it.

P.s. remember to update the README as well.

ligo/haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -177,6 +180,21 @@ mkStorageL admin votingPeriod quorumThreshold extra metadata =
...
+  -> "extra" :! ContractExtraL
+  -> "metadataHostAddress" :! Address
+  -> "metadataHostChain" :? TZIP16.ExtChainId
+  -> "metadataKey" :! MText
+  -> StorageL
+mkStorageL' admin votingPeriod quorumThreshold extra hostAddress hostChain key =
+  mkStorageL admin votingPeriod quorumThreshold extra ! #metadata metadata
+  where
+    host = maybe TZIP16.contractHost TZIP16.foreignContractHost (argF #metadataHostChain hostChain) (arg #metadataHostAddress hostAddress)
+    metadata = TZIP16.metadataURI (TZIP16.tezosStorageUri host (arg #metadataKey key))
```
pasqu4le, line 196: Instead of adding more specific function I'd suggest to keep using the existing `metadata` field of `mkStorageL` and `mkFullStorageL`.

To provide the convenience of building it from the 3 parameters a new function that produces a `TZIP16.MetadataMap BigMap` with these two lines would be better suited IMO.

Also, a nitpick: please split the first one to fit into the line length limits.

ligo/src/treasuryDAO.mligo
```diff
@@ -36,16 +36,9 @@ let treasury_DAO_proposal_check (params, extras : propose_params * contract_extr
...
-          | Xtz_transfer_type xt ->
-              if (min_xtz_amount <= xt.amount && xt.amount <= max_xtz_amount) then
-                is_valid
-              else
-                false
-      else
-        false
+      match transfer_type with
+        | Token_transfer_type tt -> is_valid
+        | Xtz_transfer_type xt -> is_valid && min_xtz_amount <= xt.amount && xt.amount <= max_xtz_amount
```
pasqu4le, line 41: Well spotted :slightly_smiling_face:

ligo/Makefile
```diff
@@ -86,17 +86,25 @@ $(OUT)/trivialDAO_storage.tz: src/**
 	# See "$(OUT)/trivialDAO_storage.tz" for compilation result  #
 	#
 
+# Are we empty? If so produce an empty big_map, otherwise make the TZIP-16 metadata big_map.
+define make-metadata
+$(if $(filter "$(1)$(2)$(3)",""),(Big_map.empty : (string, bytes) big_map),Big_map.literal [("" , Bytes.pack "tezos-storage://$(1)$(if $(filter "$(2)",""),,.$(2))/$(3)")])
+endef
```
pasqu4le, line 92: :thinking: Ligo has `String.concat` and can check if they are empty.
Why not make a function there and use it instead?

We could just pass the 3 arguments to the storage generation functions as-is and deal with them there

ligo/src/treasuryDAO.mligo
heitor-lassarote: Changes done.

ligo/haskell/src/Ligo/BaseDAO/Types.hs
heitor-lassarote: Good suggestion! All done.

ligo/Makefile
heitor-lassarote: Done. Did it a bit differently than commented by using the `^` operator instead of `String.concat`, though, and checking with an explicit `if`.

ligo/src/defaults.mligo
```diff
@@ -41,5 +41,23 @@ let default_startup_storage : startup_storage = {
...
+  if address_null && chain_null && key_null then
+    (Big_map.empty : metadata_map)
+  else
+    let uri =
+      "tezos-storage://"
+      ^ host_address
+      ^ (if chain_null then "" else host_chain ^ ".")
+      ^ "/"
+      ^ key in
+    Big_map.literal ["", Bytes.pack uri]
```
pasqu4le, line 57: @heitor-lassarote I realized now that this actually doesn't work.

The conversion from string to bytes isn't the same one as `Bytes.pack` per [TZIP-16 specification](https://gitlab.com/tzip/tzip/-/blob/27e556c0ca7f13acbf0eacea354f426447f02f18/proposals/tzip-16/tzip-16.md#contract-storage):
> Unless otherwise-specified, the encoding of the values must be the direct stream
of bytes of the data being stored. For instance, an URI starting with http:
will start with the 5 bytes 0x687474703a (h is 0x68, t is 0x74,
etc.). There is no implicit conversion to Michelson's binary format (PACK) nor
quoting mechanism.

Ligo is unable to make this conversion AFAIK, so this cannot be supported directly in LIGO code (unless it gets added in the future).

So supporting this would only mean adding a dependency just for this (single) TZIP-16 metadata use.
IMO the only thing that we can do here is accept a metadata `big_map` value directly in the `Makefile`.

So change `default_full_storage` to have `metadata_map` as input instead of `metadata_host_address, metadata_host_chain, metadata_key` and in the `Makefile` let's just put:

```
$(OUT)/treasuryDAO_storage.tz : metadata_map = Big_map.empty

	<...>'default_treasury_DAO_full_storage(<...>, "$(metadata_map)")' <...>
```

pasqu4le (changes requested):
Found a problem, there's not much that ligo lets us to about it, please fix that, the rest LGTM

ligo/src/defaults.mligo
heitor-lassarote: Good observation! Made the changes.

ligo/README.md
```diff
@@ -153,3 +155,4 @@ The arguments to this function are:
     - `slash_division_value`,
     - `min_xtz_value`,
     - `max_xtz_value`.
+- a `big_map` containing metadata described per TZIP-16
```
pasqu4le, line 158:
```suggestion
- a `big_map` containing metadata as described in TZIP-16
```

ligo/README.md
```diff
@@ -125,7 +126,8 @@ which is converted into Michelson using the `compile-storage` command. The
 arguments to this function are the admin address, the token address and the
 configuration parameters described in the Registry spec, which are
 `frozen_scale_value`, `frozen-extra_value`, `max_proposal_size`,
-`slash_scale_value` and `slash_division_value`.
+`slash_scale_value` and `slash_division_value`. Additionally, a `metadata`
+parameter is accepted, accepting a `big_map` described per TZIP-16.
```
pasqu4le, line 130:
```suggestion
parameter is accepted, accepting a `big_map` as described in TZIP-16.
```

pasqu4le (commented):
just a couple more nitpicks

ligo/README.md
heitor-lassarote: Done the suggestion here...

ligo/README.md
heitor-lassarote: ... and here as well.

pasqu4le (approved):
LGTM now

PR #158, opened 2021-03-05
[#151] Store more info for RegistryDAO

### LINKED ISSUE #151: Store more info for RegistryDAO

There are some new requested changes for RegistryDAO, since they are relatively small and should be done soon I am grouping them all together:

1. ~~add a `name` and `description` field associated to each proposal~~
2. ~~add a description text field for transfers (for both kind) to be stored on the contract~~ Edit: no longer required
3. from the `voters` field of each stored `proposal` it should be possible to determine if the voter voted against or for it

# Acceptance criteria

The 3 requested points above are implemented for both the Lorentz and LIGO version of the contract.

heitor-lassarote:
## Description

Problem: We need to be able to figure out if some voter voted for or against a proposal.

Solution: For the relevant voters, change the `Natural` so that it's a right-combed `(Natural, VoteType)`, which allows us to determine more about the vote.

## Related issue(s)

Resolves #151

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/src/types.mligo
+++ b/ligo/src/types.mligo
@@ -87,6 +87,15 @@ type migration_status =
   | MigratingTo of address
   | MigratedTo of address
 
+// Represents whether a voter has voted against (false) or for (true) a given proposal.
+type vote_type = bool
+
+type voter =
+  { voter_address : address
+  ; vote_amount : nat
+  ; vote_type : vote_type
+  }
+
 type proposal_key = bytes
 type proposal_metadata = (string, bytes) map
 type proposal =
@@ -102,10 +111,8 @@ type proposal =
   // has actually paid
   ; proposer_fixed_fee_in_token : nat
 
-  ; voters : (address * nat) list
+  ; voters : voter list
   }
-
-type vote_type = bool
 
 type voting_period = nat
 type quorum_threshold = nat
--- a/ligo/haskell/src/Ligo/BaseDAO/Types.hs
+++ b/ligo/haskell/src/Ligo/BaseDAO/Types.hs
@@ -70,7 +70,7 @@ data ProposalL = ProposalL
   , plProposerFrozenToken :: Natural
   , plProposerFixedFeeInToken :: Natural
 
-  , plVoters              :: [(Address, Natural)]
+  , plVoters              :: [Voter]
   }
   deriving stock (Show)
 
--- a/ligo/src/proposal.mligo
+++ b/ligo/src/proposal.mligo
@@ -85,7 +85,7 @@ let add_proposal (propose_params, store : propose_params * storage): storage =
     ; proposer = Tezos.sender
     ; proposer_frozen_token = propose_params.frozen_token
     ; proposer_fixed_fee_in_token = store.fixed_proposal_fee_in_token
-    ; voters = ([] : (address * nat) list)
+    ; voters = ([] : voter list)
     } in
   { store with
     proposals =
@@ -132,9 +132,14 @@ let submit_vote (proposal, vote_param, author, store : proposal * vote_param * a
       then { proposal with upvotes = proposal.upvotes + vote_param.vote_amount }
       else { proposal with downvotes = proposal.downvotes + vote_param.vote_amount }
     in
+  let voter =
+    { voter_address = author
+    ; vote_amount = vote_param.vote_amount
+    ; vote_type = vote_param.vote_type
+    } in
   let proposal =
     { proposal with
-      voters = (author, vote_param.vote_amount) :: proposal.voters
+      voters = voter :: proposal.voters
     } in
   let (ledger, total_supply) = freeze (vote_param.vote_amount, author, store.ledger, store.total_supply) in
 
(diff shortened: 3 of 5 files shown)


sras (approved):
LGTM

ligo/haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -60,7 +60,7 @@ test_RegistryDAO =
             proposalMeta = DynamicRec mempty
             proposalSize = metadataSize proposalMeta
             in withSender (AddressResolved wallet1) $ call baseDao (Call @"Propose")
-              (ProposeParams proposalSize proposalMeta)
+              (ProposeParams [mt||] proposalSize [mt||] proposalMeta)
```
pasqu4le, line 63: `MText` inherits its `Monoid` instance from `Text`, so instead of using `[mt||]` I suggest using `mempty` as it's much more readable.

ligo/src/types.mligo
```diff
@@ -144,6 +146,8 @@ type custom_ep_param = (string * bytes)
 type propose_params =
   { frozen_token : nat
   ; proposal_metadata : proposal_metadata
+  ; name : string
+  ; description : string
```
pasqu4le, line 150: This doesn't look correct.

With the addition of this two values here (and in `proposal`s) every DAO is now forced to handle these two fields.

However, the issue was about adding this information to `RegistryDAO` only, which means that it should have been done with a modification of the `proposal_metadata` handling for that DAO only.

ligo/src/registryDAO.mligo
```diff
@@ -166,7 +166,9 @@ let registry_DAO_decision_lambda (proposal, extras : proposal * contract_extra)
     : operation list * contract_extra =
   let propose_param : propose_params = {
     frozen_token = proposal.proposer_frozen_token;
-    proposal_metadata = proposal.metadata
+    proposal_metadata = proposal.metadata;
+    name = proposal.name;
+    description = proposal.description;
     } in
   let proposal_key = to_proposal_key (propose_param, proposal.proposer) in
```
pasqu4le, line 173: Similarly this seems incorrect to me.

The discussion linked in the description indeed is about this point and the fact that `toProposalKey` uses the `propose_param` which has now the problem of including `name` and `description` (note: the issue remains if these are in the `proposal_metadata`).

I figured that this was bad because these are metadata and shouldn't really influence the proposal size calculation (which they do ATM), however it was argued that it wouldn't be that bad and that the parameters can be tweaked to account for those. What do you think?

pasqu4le (commented):
One more minor thing: In the commit message the first line says:
```
Problem: We need to add a `name` and `field` field to each proposal. We
```
but it should be fixed to:
```
Problem: We need to add a `name` and `description` field to each proposal. We
```

ligo/src/types.mligo
heitor-lassarote: I'll look into it. I assume the `voters` field should be handled only in `RegistryDAO` and must be changed as well?

ligo/src/registryDAO.mligo
heitor-lassarote: If for whatever reason the users think it should be accounted for, it's already there. If not, then they can tweak the parameters or leave these fields empty, so I think it makes sense to account for them. But I don't really have much of a strong opinion towards one option or the other, so perhaps the simpler one is best, I think.

ligo/src/types.mligo
pasqu4le: No, `voters` should be changed for every kind of DAO, so it's correct for the change to be in `baseDAO`.

You only need to move `name` and `description` to the `proposal_metadata` of RegistryDAO.

ligo/src/registryDAO.mligo
pasqu4le: Indeed, the parameters should be flexible enough to be tweaked or not for those, which leaves the choice to the user as well as make the default calculation much simpler.

So agreed, no modification to the `propose_params.frozen_token` check will need to be done here.

ligo/src/types.mligo
```diff
@@ -87,6 +87,8 @@ type migration_status =
   | MigratingTo of address
   | MigratedTo of address
 
+type vote_type = bool
```
pasqu4le, line 90: Can you add a comment to clarify which value means "voted for" and which "voted against"?

Just to have it cleared up somewhere for whoever reads this next

ligo/src/types.mligo
```diff
@@ -102,11 +104,9 @@ type proposal =
   // has actually paid
   ; proposer_fixed_fee_in_token : nat
 
-  ; voters : (address * nat) list
+  ; voters : (address * (nat * vote_type)) list
```
pasqu4le, line 107: one more thing: we should put this in its own type.

Firstly, because we have to use this signature in one place and that would be better.
Secondly, because we should annotate all the types in storage (especially this one whose purpose is to be read from).

So I suggest adding:
```ocaml
type voter =
  { voter_address : address
  ; vote_amount : nat
  ; vote_type : vote_type
  }
```
and just use a `voter list` here.

pasqu4le (commented):
Just two nitpicks, otherwise LGTM

pasqu4le (approved):

PR #159, opened 2021-03-08
[#113] RegistryDAO lookup-registry view

### LINKED ISSUE #113: Add an on-chain view to RegistryDAO to get an item

Sometimes it can be desirable to get something from RegistryDAO's big_map on-chain. Currently the most common (but imperfect) way to do it is using `view` entrypoints that look like `pair x (contract y)` (i. e. take a callback contract).

# Acceptance criteria

RegistryDAO has an additional entrypoint of type `pair k (contract (pair (k, option v)))` that takes a value of `k` key from the registry and passes it to the callback contract along with the original `k`. `k` may look redundant, but it looks safer this way.

We should also update https://github.com/tqtezos/baseDAO/blob/master/docs/registry.md accordingly.

sras:
## Description

1. Adds custom entrypoint to lookup registry values to both LIGO and Lorentz versions of RegistryDAO contract. 
2. Adds tests for both and updates specification.

More detailed description are included in commit messages.

## Related issue(s)

Resolves #113 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -47,3 +47,9 @@ One can set them to 1 and 1 by default to always unfreeze all tokens.
 
 * The decision lambda simply applies all updates from the accepted proposal one by one.
 They are applied in the same order as specified in the proposal, the head of the list is applied first.
+
+### Lookup registry on-chain view
+
+A custom entrypoint `lookup_registry` is included in the registryDAO contract
+to lookup values in the registry. The parameter of this entrypoint is `(pair k address)`
+where `address` is the address of a callback contract of type `(k, v option)`.
--- a/ligo/haskell/test/Test/Ligo/RegistryDAO.hs
+++ b/ligo/haskell/test/Test/Ligo/RegistryDAO.hs
@@ -188,6 +188,46 @@ test_RegistryDAO =
               -- Now we expect this to work
               withSender (AddressResolved wallet1) $
                 call baseDao (Call @"Propose") (ProposeParams requiredFrozen largeProposalMeta)
+
+    , nettestScenarioCaps "checks on-chain view correctly returns the registry value" $ do
+        -- The default values assigned from initialStorageWithExplictRegistryDAOConfig function
+        let frozen_scale_value = 1
+        let frozen_extra_value = 0
+        withOriginated 3
+          (\(admin: wallet1: voter1:_) -> setExtra @Natural [mt|max_proposal_size|] 200 $
+              initialStorageWithExplictRegistryDAOConfig admin [wallet1, voter1]) $
+
+          \(admin: wallet1: voter1 : _) _ baseDao -> do
+            let
+              proposalMeta = DynamicRec $ Map.fromList $
+                [ ([mt|updates|], lPackValueRaw [([mt|key|], Just [mt|testVal|])])
+                , ([mt|agoraPostID|], lPackValueRaw @Natural 10)
+                ]
+              proposalSize = metadataSize proposalMeta
+
+            runIO $ putTextLn $ show proposalSize
+            let requiredFrozen = proposalSize -- since frozen_scale_value and frozen_scale_value are 1 and 0.
+
+            -- We propose the addition of a new registry key
+            withSender (AddressResolved wallet1) $
+              call baseDao (Call @"Propose") (ProposeParams requiredFrozen proposalMeta)
+
+            -- Then we send 2 upvotes for the proposal (as min quorum is 2)
+            let proposalKey = makeProposalKey @ProposalMetadataL (ProposeParams requiredFrozen proposalMeta) wallet1
+            withSender (AddressResolved voter1) $
+              call baseDao (Call @"Vote") [PermitProtected (VoteParam proposalKey True 2) Nothing]
+
+            advanceTime (day 12) -- voting period is 11 days.
+            withSender (AddressResolved admin) $
+              call baseDao (Call @"Flush") (1 :: Natural)
+
+            consumer <- originateSimple "consumer" [] (contractConsumer @(MText, (Maybe MText)))
+
+            withSender (AddressResolved voter1) $
+              call baseDao (Call @"CallCustom") ([mt|lookup_registry|], lPackValueRaw ([mt|key|], consumer))
+
+            checkStorage (AddressResolved $ unTAddress consumer) (toVal [([mt|key|], Just [mt|testVal|])])
+
     ]
   ]
   where
--- a/ligo/haskell/test/Test/Ligo/RegistryDAO.hs
+++ b/ligo/haskell/test/Test/Ligo/RegistryDAO.hs
@@ -191,8 +191,6 @@ test_RegistryDAO =
 
     , nettestScenarioCaps "checks on-chain view correctly returns the registry value" $ do
         -- The default values assigned from initialStorageWithExplictRegistryDAOConfig function
-        let frozen_scale_value = 1
-        let frozen_extra_value = 0
         withOriginated 3
           (\(admin: wallet1: voter1:_) -> setExtra @Natural [mt|max_proposal_size|] 200 $
               initialStorageWithExplictRegistryDAOConfig admin [wallet1, voter1]) $
--- a/ligo/src/registryDAO.mligo
+++ b/ligo/src/registryDAO.mligo
@@ -251,7 +251,7 @@ let default_registry_DAO_full_storage (admin, token_address, frozen_scale_value,
     proposal_check = registry_DAO_proposal_check;
     rejected_proposal_return_value = registry_DAO_rejected_proposal_return_value;
     decision_lambda = registry_DAO_decision_lambda;
-    custom_entrypoints = Map.literal [("receive_xtz", Bytes.pack (lookup_registry))];
+    custom_entrypoints = Map.literal [("lookup_registry", Bytes.pack (lookup_registry))];
     } in
   (new_storage, new_config)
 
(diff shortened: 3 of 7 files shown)


ligo/haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -191,6 +191,51 @@ test_RegistryDAO =
...
+        let frozen_extra_value = 0
+        let slash_scale_value = 1
+        let slash_division_value = 1
+        withOriginated 3
+          (\(admin: wallet1: voter1:_) -> setExtra @Natural [mt|slash_division_value|] 2 $
+            setExtra @Natural [mt|frozen_scale_value|] frozen_scale_value $
+            setExtra @Natural [mt|frozen_extra_value|] frozen_extra_value $
+            setExtra @Natural [mt|slash_scale_value|] slash_scale_value $
+            setExtra @Natural [mt|max_proposal_size|] 200 $
+            setExtra @Natural [mt|slash_division_value|] slash_division_value $ initialStorageWithExplictRegistryDAOConfig admin [wallet1, voter1]) $
```
heitor-lassarote, line 206: `initialStorageWithExplictRegistryDAOConfig` already sets the `frozen_scale_value`, `frozen_extra_value`, `slash_scale_value` and `slash_division_value` to the values you set up above, so I think only setting `max_proposal_size` should be enough. Also, you're setting `slash_division_value` twice.

ligo/haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -191,6 +191,51 @@ test_RegistryDAO =
...
+          \(admin: wallet1: voter1 : _) _ baseDao -> let
+            proposalMeta = DynamicRec $ Map.fromList $
+              [ ([mt|updates|], lPackValueRaw [([mt|key|], Just [mt|testVal|])])
+              , ([mt|agoraPostID|], lPackValueRaw @Natural 10)
+              ]
+            proposalSize = metadataSize proposalMeta
+
+            in do
+              runIO $ putTextLn $ show proposalSize
+              let requiredFrozen = proposalSize * frozen_scale_value + frozen_extra_value
```
heitor-lassarote, line 217: Since `frozen_scale_value = 1` and `frozen_extra_value = 0`, you may as well skip these calculations.

Also, a nitpick: I think it would be a bit better if you started this `do` block at line 208, so that the test has one less level of indentation.

ligo/src/registryDAO.mligo
```diff
@@ -217,6 +217,21 @@ let registry_DAO_decision_lambda (proposal, extras : proposal * contract_extra)
         | None -> new_ce
       in (([] : operation list), new_ce)
 
+// A custom entrypoint to fetch values from Registry
+let lookup_registry (params, config_store : bytes * configured_storage) : operation list * storage =
```
heitor-lassarote, line 221: In `types.mligo`, your function corresponds to a `storable_entrypoint`, and your return type is `return`. I'm not sure if OCaml syntax allows you to use `storable_entrypoint` as a type here somehow, but I think it would be best to at least change the return type to `return`.

docs/registry.md
```diff
@@ -47,3 +47,14 @@ One can set them to 1 and 1 by default to always unfreeze all tokens.
...
+### Lookup registry on-chain view
+
+A custom entrypoint `lookup_registry` is included in the registryDAO contract to lookup values in the registry.
+The parameter of this entrypoint is
+
+```
+(pair k address)
+```
+
+Where `address` is the address of a callback contract of type `(k, v option)`.
```
heitor-lassarote, line 60: I think it would look better if you inlined this type, like so:

A custom entrypoint `lookup_registry` is included in the registryDAO contract to lookup values in the registry.
The parameter of this entrypoint is `(pair k address)`, where `address` is the address of a callback contract of type `(k, v option)`.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -172,11 +173,52 @@ decisionLambda = do
...
+  deriving stock (Generic, Show)
+  deriving anyclass (IsoValue, HasAnnotation)
+
+customEntrypointsHandler ::
+  forall s store k v.
+     ( IsoRegistryDaoProposalMetadata k v
+     , NiceParameter (k, v)
+     , DAO.StorageC store (RegistryDaoContractExtra k v) (RegistryDaoProposalMetadata k v)
+     ) => CustomParam k v : store : s
+    :-> List Operation : store : s
```
heitor-lassarote, line 188: Nitpick: please apply formatting to this function signature that is similar to the ones that already exist in this file. It's a bit inconsistent already as it seems there are two styles, so let's not add a third one.

test/Test/RegistryDAO.hs
```diff
@@ -36,11 +40,28 @@ test_RegistryDAO = testGroup "RegistryDAO Tests"
...
+getRegistryValue :: (Monad m) => NettestImpl m -> m ()
+getRegistryValue = uncapsNettest $ do
+  -- To check the view, we put a test value in registry during contract initialization
+  -- so that we don't have to create and execute a proposal to put the value there. Some
+  -- dummy values of proposal key and timestamp is used to create the corresponding registry
+  -- entry.
+  let dummyHash = HashUnsafe ""
+  let testKey = encodeUtf8 @Text "testKey"
+  let testVal = encodeUtf8 @Text "testVal"
+  let registryEntry = RegistryEntry (Just testVal) dummyHash  (Timestamp 0)
```
heitor-lassarote, line 58: I think you may simply use `"testKey"` and `"testVal"` literals here. But I think it would be best if you used `dummy` instead of `test`.
Also, another nitpick: There are two spaces after `dummyHash`.

sras: > I think you may simply use "testKey" and "testVal" literals here. 

Yes, I just wanted to make it explicit that the encoding being used is UTF8.

ligo/haskell/test/Test/Ligo/RegistryDAO.hs
sras: Makes sense. I think I wanted to make the values being used explicit. But since they just result in the proposal being rejected or not, I think it is not a big deal, so I have made the changes. Have left some comments though.

ligo/haskell/test/Test/Ligo/RegistryDAO.hs
sras: Changed.

ligo/src/registryDAO.mligo
sras: >if OCaml syntax allows you to use 

Not sure if I have missed something, but it appears that it not straightforward to do so. But I have changed to use  `return` here.

docs/registry.md
sras: Done.

src/Lorentz/Contracts/RegistryDAO.hs
sras: Have fixed the formatting.

test/Test/RegistryDAO.hs
sras: Have fixed it except for the explicit utf8 encoding.

ligo/src/registryDAO.mligo
heitor-lassarote: Yeah, I wasn't sure about it either. I just saw the existence of that type and assumed it could be used there.

heitor-lassarote (approved):
Oops, forgot to approve! It looks good to me now! :)

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -172,11 +173,51 @@ decisionLambda = do
     )
   nil
 
-config ::
+
+data CustomParam k v
```
pasqu4le, line 177: I think this should be in `Lorentz.Contracts.RegistryDAO.Types`, also to stay in line with the rest of the "unwritten convention" it should have `RegistryDAO` in its name.

src/Lorentz/Contracts/RegistryDAO.hs
```diff
@@ -192,6 +233,8 @@ config = DAO.defaultConfig
 
   , DAO.cMaxVotes = 1000
   , DAO.cMaxProposals = 500
+
+  , DAO.cCustomCall = customEntrypointsHandler @_ @_ @k @v
```
pasqu4le, line 237: Nitpick, but given that this is only used here, why not move the type arguments in the `forall` to have `k` and `v` first?

pasqu4le (commented):
Only minor points, otherwise LGTM

One more thing: in the Ligo commit the message says that the callback is implemented with an `address` because we cannot store `contract`s in the storage.

That's sort of incorrect, because the reason is that we cannot `unpack` contracts, and so we cannot store them in `bytes` (which are used for custom entrypoints' parameters).

sras:
>One more thing: in the Ligo commit the message says that the callback is implemented with an address because we cannot store contracts in the storage.

Right, I have fixed the commit message now.

src/Lorentz/Contracts/RegistryDAO.hs
sras: Yes, changed.

src/Lorentz/Contracts/RegistryDAO.hs
sras: Have renamed the type and moved it to `Lorentz.Contracts.RegistryDAO.Types` module.

pasqu4le (approved):
LGTM now

pasqu4le:
Rebased on `master`.

Note: the only real change that I had to make was adding a `ParameterHasEntrypoints` instance for the Lorentz version entrypoint.

PR #161, opened 2021-03-10
[TM-523] Bump morley and other deps

rvem:
## Description
Problem: We want to check whether this repo can be used with the latest
morley and new GHC version.

Solution: Bump dependencies and make required changes in the code.

Currently based on #160 

## Related issue(s)

https://issues.serokell.io/issue/TM-523

- Tests
        silently reappearing again.

- Documentation
    - Haddock


template/package.yaml
```diff
@@ -76,9 +76,14 @@ ghc-options:
...
   -Wno-monomorphism-restriction
   -Wno-implicit-prelude
+  -Wno-unused-packages
+  -Wno-prepositive-qualified-module
+  -Wno-missing-safe-haskell-mode
+
 
 dependencies:
-  - base >= 4.7 && < 5
+  - name: base
```
pasqu4le, line 85: Shouldn't this be `base_noprelude` now? :thinking:

pasqu4le (commented):
Only one question.

P.s. the only thing preventing the MR this is based on to be merged are the dependency updates... in this PR.

So, this can be merged into the branch it's currently targeting and then that one will be pushed to master.

template/package.yaml
rvem: indeed

pasqu4le (approved):

PR #162, opened 2021-03-12
[#138] Move all inlined constants to storage

### LINKED ISSUE #138: Move all inlined constants to storage

We should move all the constants that are currently in-lined in the `code` to the `storage`.

Currently the LIGO version of the contract has very little of them (due to the `config` being part of the overall contract `storage`)., but there may be some.
All the ones I can think of are the `unfrozen_token_id` and `frozen_token_id` `nat` values, but we should check carefully.

Some clarifications are needed:
1. Does "constants" include the lambda used in the configuration as well?
2. Should this affect the Lorentz version of the contract too, or just the LIGO one?
3. (a combination of the previous two) If this is an update intended for the Lorentz version, should only the `nat`/`Natural`s of its configuration be moved to `storage` or the code in it be moved to stored `lambda`s too?

# Acceptance criteria

Constants that are part of the `storage` and not in-lined into the `code`.

heitor-lassarote:
## Description

This PR moves constants inlined in the code to the storage. Particularly, `unfrozen_token_id` and `frozen_token_id` were moved to the storage in both the Ligo and Lorentz versions. In the Lorentz version, the `Config` type has its fields inlined in various parts at the `Proposal` module, and it also has been moved to the storage and its usages properly updated.

## Related issue(s)

Resolves #138

- Tests
  - [X] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/haskell/src/Ligo/BaseDAO/Types.hs
+++ b/ligo/haskell/src/Ligo/BaseDAO/Types.hs
@@ -121,6 +121,7 @@ data ParameterL
 data StorageL = StorageL
   { sAdmin :: Address
   , sExtra :: ContractExtraL
+  , sFrozenTokenId :: FA2.TokenId
   , sLedger :: Ledger
   , sMetadata :: TZIP16.MetadataMap BigMap
   , sMigrationStatus :: MigrationStatus
@@ -133,6 +134,7 @@ data StorageL = StorageL
   , sTokenAddress :: Address
   , sVotingPeriod :: VotingPeriod
   , sTotalSupply :: TotalSupply
+  , sUnfrozenTokenId :: FA2.TokenId
   , sFixedProposalFeeInToken :: Natural
   }
   deriving stock (Show)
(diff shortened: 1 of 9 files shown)


heitor-lassarote:
@pasqu4le, could I ask for your help with something on this PR?

There is one specific test that refuses to compile with `Duplicated entrypoint name` and I'm not sure what I've been doing wrong.

For context, [this is the error](https://buildkite.com/serokell/basedao/builds/1309#77de3fd8-cba6-40e3-826c-070af10fae53). Other tests compile and run fine with `baseDaoContract` (if I comment this test all other tests will pass), so I'm not sure why this one test does this. This is currently the only thing pending for this PR to be ready.

BTW, I had this error many times before, when I was giving a `Config ce pm op` as the argument to `baseDaoContract` (as it was its argument previously), but it was easily solved by changing it with the new arguments (`Text` and `Markdown`). This one test is the exception.

pasqu4le (approved):
LGTM :ok_hand:

PR #163, opened 2021-03-12
Use morley-large-originator to handle the large contract size

pasqu4le:
## Description

Problem: in #123 we had to deal with the excessive size of the Ligo
version of the contract, while the solution is working it's not
convenient to use, it also has excessive running times and may not
be enough in the near future.
Moreover, the Lorentz version is currently just under the limit and
will need a solution as well.

Solution: we have a new tool, `morley-large-originator` that can
facilitate the origination of large contracts.
Moreover this integrated with our testing framework, so from this
moment we'll can make use of it.

Note: this reverts most of the changes made for #123, to not overlap
the two approaches.
On top of that, nettests have been temporarily deactivated on the
Lorentz version as well, due to the excessive running time.

## Related issue(s)

Related to #123 and #124

- Tests
        silently reappearing again.

- Documentation
    - Haddock


ligo/src/base_DAO.mligo
```diff
@@ -3,77 +3,45 @@
...
 
 (*
- * Returns name and packed parameter of a stored entrypoint that requires the
- * contract to not be migrated.
- * This checks for such migration condition.
+ * Entrypoints that require the contract to be migrated.
+ * Note: This checks for such migration condition.
  *)
-let requiring_no_migration (param, store : migratable_parameter * storage) : (string * bytes) =
+let requiring_no_migration (param, store, config : migratable_parameter * storage * config)
```
heitor-lassarote, line 43: Perhaps we could use `full_storage` here and in `requiring_no_xtz`.

ligo/src/management.mligo
```diff
@@ -50,3 +50,24 @@ let confirm_migration(param, store : unit * storage) : return =
   | MigratedTo (new_addr)  ->
       ([%Michelson ({| { FAILWITH } |} : string * address -> return)]
         ("MIGRATED", new_addr) : return)
+
+(*
+ * Call a custom entrypoint. Looks up the packed code for entrypoint using the entrypoint
+ * name in the parameter. Then unpacks the code to a lambda of type
+ * '(bytes * full_storage) -> return'.
+ * Then just pass the bytes from parameter to this lambda return the resulting value.
```
heitor-lassarote, line 58: This phrase needs rewording as it's a bit difficult to understand.

ligo/src/defaults.mligo
```diff
@@ -14,6 +14,7 @@ let default_config : config = {
     min_quorum_threshold = 1n;
     max_voting_period = 2592000n; // 60 * 60 * 24 * 30
     min_voting_period = 1n;
+    custom_entrypoints = (Map.empty : (string, bytes) map);
```
heitor-lassarote, line 17: Let's use the `custom_entrypoints` type defined in `types.mligo`.

ligo/haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -189,17 +178,20 @@ data ConfigL = ConfigL
...
   , cMaxVotingPeriod :: Natural
   , cMinVotingPeriod :: Natural
+
+  , cCustomEntrypoints :: CustomEntrypointsL
   }
   deriving stock (Show)
 
-mkConfigL :: ConfigL
-mkConfigL = ConfigL
+mkConfigL :: [(MText, ByteString)] -> ConfigL
```
heitor-lassarote, line 186: I think it would be best to use a type synonym here or a named parameter like in `mkStorageL`.

heitor-lassarote (changes requested):

ligo/src/base_DAO.mligo
pasqu4le: It doesn't really make a difference in terms of compiled result, however not carrying `full_storage` allows us to avoid splitting it in both function (which would be necessary) and only do that in `base_DAO_contract`:
```ocaml
  let store : storage = full_store.0 in
  let config : config = full_store.1 in
```
which removes some boilerplate.

ligo/Makefile
```diff
@@ -118,10 +102,31 @@ $(OUT)/treasuryDAO_storage.tz: src/**
 	# See "$(OUT)/treasuryDAO_storage.tz" for compilation result #
 	#
 
+originate : storage = $(OUT)/trivialDAO_storage.tz
+originate : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+originate : contract_name = baseDAO
+originate: $(OUT)/baseDAO.tz
+	# ============== Originating DAO with $(storage) ============== #
+	@$(ORIGINATE) --contract $(OUT)/baseDAO.tz --from "$(admin_address)" \
+								--initial-storage '$(shell cat $(storage))' \
```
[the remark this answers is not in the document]
pasqu4le: Good catch, should be fixed now

ligo/src/base_DAO.mligo
heitor-lassarote: Understandable, then let's leave it as is.

heitor-lassarote (approved):
LGTM

PR #165, opened 2021-03-15
[#67] Add a configurable fixed fee for a proposal

### LINKED ISSUE #67: Fixed Fee for All Proposals

Have the possible requirement of having a modifiable fixed fee in either the native DAO asset or tez order to submit proposals to prevent spam. This fee being something that replaces the "proportion of the ask" style spam-measure.

kkirka:
## Description

It is desirable to prevent spam by setting a configurable fixed fee for submitting a proposal. We need to support both fees in Tezos and in the native asset.

PR checklist:
- [x] Support fees in native asset
- [x] Support fees Tez

## Related issue(s)

Resolves #67 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification-ligo.md
+++ b/docs/specification-ligo.md
@@ -107,6 +107,9 @@ type proposal =
   ; metadata : proposal_metadata
   ; proposer : address
   ; proposer_frozen_token : nat
+  ; proposer_fixed_fee_in_token : nat
+  // ^ A fee paid for submitting the proposal.
+  // Needed to refund correctly if the proposal is successful.
   ; voters : (address * nat) list
   // ^ List of voter addresses associated with the vote amount
   // Needed for `flush` entrypoint.
@@ -121,6 +124,7 @@ These values are:
 1. `admin : address` is the address that can perform administrative actions.
 2. `voting_period : nat` specifies how long the voting period lasts.
 3. `quorum_threshold : nat` specifies how many total votes are required for a successful proposal.
+4. `fixed_proposal_fee_in_token : nat` specifies the fee for submitting a proposal (in native DAO token).
 
 # Contract logic
 
@@ -230,6 +234,7 @@ Full list:
 * [`transfer_ownership`](#transfer_ownership)
 * [`accept_ownership`](#accept_ownership)
 * [`propose`](#propose)
+* [`set_fixed_fee_in_token`](#set_fixed_fee_in_token)
 * [`set_voting_period`](#set_voting_period)
 * [`set_quorum_threshold`](#set_quorum_threshold)
 * [`vote`](#vote)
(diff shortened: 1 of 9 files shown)


ligo/src/startup.mligo
```diff
@@ -72,6 +72,11 @@ let storable_vote (packed_param, config_store : bytes * configured_storage) : re
     vote_with_self(vote_param, ctrt_address, config_store.1, config_store.0)
   | None -> (failwith "ENTRYPOINT_NOT_FOUND" : return)
 
+let storable_set_fixed_fee_in_token (packed_param, config_store : bytes * configured_storage) : return =
```
pasqu4le, line 75: Note: this won't be necessary after rebase (this file is no longer needed).

ligo/Makefile
```diff
@@ -41,6 +41,7 @@ all: \
 	$(OUT)/entrypoints/migrate.tz \
 	$(OUT)/entrypoints/mint.tz \
 	$(OUT)/entrypoints/propose.tz \
+	$(OUT)/entrypoints/set_fixed_fee_in_token.tz \
```
pasqu4le, line 44: Note: rebase will make this change unnecessary.

ligo/src/base_DAO.mligo
```diff
@@ -58,6 +58,7 @@ let requiring_no_xtz (param : forbid_xtz_params) : (string * bytes) =
     | Confirm_migration p -> ("confirm_migration", Bytes.pack p)
     | Propose p -> ("propose", Bytes.pack p)
     | Vote p -> ("vote", Bytes.pack (p, Tezos.self_address))
+    | Set_fixed_fee_in_token p -> ("set_fixed_fee_in_token", Bytes.pack p)
```
pasqu4le, line 61: Note: after rebase you can change this to:
```ocaml
    | Set_fixed_fee_in_token p -> set_fixed_fee_in_token(p, store)

```

ligo/src/types.mligo
```diff
@@ -96,6 +96,7 @@ type proposal =
   ; metadata : proposal_metadata
   ; proposer : address
   ; proposer_frozen_token : nat
+  ; proposer_fixed_fee_in_token : nat
```
pasqu4le, line 99: Ah I didn't get why this was needed at first, but I get it now.

Changing the fixed fee only applies to proposals created after it, so we need to track this.
Can you add a quick comment here to note this, so we don't forget?

ligo/src/proposal.mligo
```diff
@@ -80,7 +80,7 @@ let unfreeze (tokens, addr, ledger, total_supply : nat * address * ledger * tota
   let (ledger, total_supply) = credit_to (tokens, addr, unfrozen_token_id, ledger, total_supply) in
   (ledger, total_supply)
 
-let add_proposal (propose_params, store : propose_params * storage): storage =
+let add_proposal (propose_params, fixed_fee_in_token, store : propose_params * nat * storage): storage =
```
pasqu4le, line 83: `fixed_fee_in_token` seems unnecessary, as it can be extracted from `store` itself, am I right?

pasqu4le (commented):
Only looked at the implementation, not the tests, mostly LGTM.

Only a couple of nitpicks for now.
Note: before continuing it needs to be rebased.

ligo/src/proposal.mligo
kkirka: Nice catch, thx! Initially I put it to `config` instead of store, and then I remembered it had to be modifiable, so I made it a part of store but forgot to remove the extra parameter.

ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -7,10 +7,62 @@ module Test.Ligo.BaseDAO.Proposal
```
pasqu4le, line 7: It looks like your tests are now only failing due to `Natural` not being in scope, it suggests you to take it from `Lorentz`, but you can also just use `Universum`:
```suggestion

import Universum

```
and it should be fine.

pasqu4le (commented):
Token fee LGTM now, my only suggestion there would be to add an option in the `Makefile` (and an argument in the default storage creation functions) to set the initial value for it.

pasqu4le (commented):
Changes look good, will wait for a spec update and more tests.

docs/specification-ligo.md
```diff
@@ -600,14 +605,29 @@ Parameter (in Michelson):
...
+Set_fixed_fee_in_token of nat
+```
+
+Parameter (in Michelson):
+```
+(nat %set_fixed_fee_in_token)
+```
+
+- Update the fee that the proposers have to pay to submit a proposal (the fee is returned if the proposal is successful and burnt otherwise)
+- This affects only new proposals; the existing proposals store the information about the fee paid for their submission.
```
pasqu4le, line 629:
```suggestion
- This affects only new proposals; the existing proposals store the fee paid for their submission.
```
Nitpick, but that seems redundant.

ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -9,62 +9,167 @@ import Universum
...
+          withSender (AddressResolved someone) $
+            call dao (Call @"Set_fixed_fee_in_token") 1000
+              & expectCustomErrorNoArg #nOT_ADMIN
+
+    , nettestScenario "a proposer is returned a fee after the proposal succeeds" $
+        uncapsNettest $ do
+          ((proposer, _), (voter, _), dao, admin) <- originateLigoDao
+
+          -- Use 60s for voting period, since in real network by the time we call
+          -- vote entrypoint 30s is already passed.
```
pasqu4le, line 112:
```suggestion
          -- the vote entrypoint 30s have already passed.
```
Nitpick as well.

Leaving the comment was a good idea tho, there's also other tests that are time-dependent and we'll need to keep this in mind.

pasqu4le (commented):
LGTM, when squashing remember to add a proper message to the test commit.

ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
kkirka: I found the code with `advanceTime` in another module, and copied it with this comment (it's indeed helpful). I think I should fix the grammar there as well.

pasqu4le: Oh I didn't know that.

Yes, please fix that as well :ok_hand:

pasqu4le (approved):

PR #167, opened 2021-03-18
[#166] Replace nat views with void entrypoints

### LINKED ISSUE #166: Remove unnecessary nat views

We currently have two `nat` views that are not necessary: `get_total_supply` and `get_vote_permit_counter`.
Both these views main use is for testing, which could be replaced.

More importantly these pose a potential security risk in case a `DAO` contract has the administrator role in another `DAO` or it's administrator of itself, in which voting period and quorum threshold can be changed by anyone.

These could have been just removed if #31 was ready, but that's not the case.
So, to update our tests accordingly for the moment we can (from best to worst):
1. avoid to check those values directly altogether (but keep the test indirectly as we do for other ones).
2. move the tests that depend on these to the integrational engine only.
3. replace them with `void`s and use those instead.

# Acceptance criteria

1. Tests are updated depending on which one of the solutions above is possible
2. Both on-chain views,  `get_total_supply` and `get_vote_permit_counter` are removed from the contract

Note: this issue affects both the Lorentz and Ligo version of the contract.

kkirka:
## Description

**Problem:** We need to somehow get data from storage but using views is insecure: other contracts can trick DAO into sending operations to arbitrary contracts.

**Solution:** Replace views with `void` entrypoints.

## Related issue(s)

Resolves #166 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
+++ b/ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
@@ -160,14 +160,13 @@ test_BaseDAO_Proposal =
           checkTokenBalance unfrozenTokenId dao proposer 48
 
           -- Check total supply
-          let getTotalSupply tokenId consumer = call dao (Call @"Get_total_supply") (mkView tokenId consumer)
-          consumer <- originateSimple "consumer" [] contractConsumer
-          withSender (AddressResolved proposer) $ getTotalSupply unfrozenTokenId consumer
-          checkStorage (AddressResolved $ toAddress consumer) (toVal [148 :: Natural]) -- initial = 200
+          withSender (AddressResolved proposer) $
+            call dao (Call @"Get_total_supply") (mkVoid unfrozenTokenId)
+              & expectError (VoidResult (148 :: Natural)) -- initial = 200
 
-          consumer2 <- originateSimple "consumer" [] contractConsumer
-          withSender (AddressResolved proposer) $ getTotalSupply frozenTokenId consumer2
-          checkStorage (AddressResolved $ toAddress consumer2) (toVal [52 :: Natural]) -- initial = 0
+          withSender (AddressResolved proposer) $
+            call dao (Call @"Get_total_supply") (mkVoid frozenTokenId)
+              & expectError (VoidResult (52 :: Natural)) -- initial = 0
 
     , nettestScenario "cannot propose with insufficient tokens to pay the fee" $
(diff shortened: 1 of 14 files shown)


ligo/src/base_DAO.mligo
```diff
@@ -33,7 +33,7 @@ let requiring_no_xtz (param, store, config : forbid_xtz_params * storage * confi
     | Flush (p) -> flush (p, config, store)
     | Burn (p) -> burn(p, store)
     | Mint (p) -> mint(p, store)
-    | GetVotePermitCounter (p) -> get_vote_permit_counter(p, store)
+    | Get_vote_permit_counter (p) -> get_vote_permit_counter(p, store)
```
pasqu4le, line 36: Good move :+1:

ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -160,14 +160,13 @@ test_BaseDAO_Proposal =
           checkTokenBalance unfrozenTokenId dao proposer 48
 
           -- Check total supply
-          let getTotalSupply tokenId consumer = call dao (Call @"Get_total_supply") (mkView tokenId consumer)
-          consumer <- originateSimple "consumer" [] contractConsumer
-          withSender (AddressResolved proposer) $ getTotalSupply unfrozenTokenId consumer
-          checkStorage (AddressResolved $ toAddress consumer) (toVal [148 :: Natural]) -- initial = 200
+          withSender (AddressResolved proposer) $
+            call dao (Call @"Get_total_supply") (mkVoid unfrozenTokenId)
+              & expectError (VoidResult (148 :: Natural)) -- initial = 200
```
pasqu4le, line 165: I wonder why your removed the `-- initial = *` comments from everywhere but here :thinking:

pasqu4le (commented):
Only a question and a nitpick: in both the PR description and the commit message there's a `usign` typo.

Otherwise LGTM, but you'll also need to run `./scripts/generate-cabal-files.sh` to update the cabal files.

ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
kkirka: Because copypasting is hard! Reverted :)

pasqu4le (approved):
LGTM now :ok_hand: 

Don't forget to squash before merging

PR #172, opened 2021-03-22
[#68] Add a nettest stress test for 'propose' entrypoint

### LINKED ISSUE #68: Allow the option of unlimited number of proposals in a cycle

Straightforward - enable an unlimited number of proposals in a cycle as an option for DAOs created using this framework.

sras:
## Description

Adds a nettest test to create 500 proposals, and check that the contract is useable after that.

## Related issue(s)

Resolves #68

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
@@ -177,6 +177,8 @@ test_BaseDAO_Proposal =
 
     , nettestScenarioOnEmulatorCaps "the frozen tokens are correctly unstaked when address cast multiple votes" $
         unstakesTokensForMultipleVotes getFreezeHistoryEmulator
+    , nettestScenarioOnNetworkCaps "proposal endpoint stress test" $
+        proposalCreationUnderStorageLimit (originateLigoDaoWithConfigDesc dynRecUnsafe)
     ]
 
   , testGroup "QuorumThreshold Updates"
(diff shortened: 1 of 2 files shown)


ligo/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -1269,3 +1273,42 @@ voteOutdatedProposal _ originateFn = do
...
       & expectCustomErrorNoArg #vOTING_PERIOD_OVER
+
+proposalCreationUnderStorageLimit
+  :: forall caps base m.
+    ( MonadNettest caps base m
+    , HasCallStack
+    )
+  => IsLorentz -> (ConfigDesc ConfigL -> OriginateFn ParameterL m) -> m ()
+proposalCreationUnderStorageLimit  _ originateFn = do
+  ((owner1, _), _, dao, admin) <- originateFn testConfigWithLongVP
```
pasqu4le, line 1284: @sras I now re-considered this approach and I think we should put the proposals in the contract at origination instead.

So, what I think this tests should do is:
1. create a DAO that contains 1000 proposals from the very start
2. check that we can make a new proposal on that
3. check that we can migrate after that as well

My reasoning is that we want to check for a lot more (1000) to guarantee some margine of error and after that we want to check a big operation (`propose`) as well as a `migrate`, which in the worst case scenario should still always work. What do you think?

Note that this way we should also be able to run this test in a reasonable amount of time on the network; the even larger origination will take a long time, but less than making all of the `propose` calls as it is now.

pasqu4le:
Note: the time-saving approach currently does not work due to the `big_map`s containing too many elements to originate the `originator` contract itself from `morley-large-originator`. 

This limitation however should be solved by 9a morley issue](https://gitlab.com/morley-framework/morley/-/issues/575) soon, so we should just wait for it, since we already proved that the default limit provided of 500 is currently sane.

sras:
Closing this MR as corresponding issue is no longer relavent.

PR #176, opened 2021-03-26
Don't use dedicated runner for scheduled jobs

rvem:
## Description
Problem: We now have more CI agents and having a dedicated scheduled
agent is no longer needed.

Solution: Remove scheduled queue tag.

## Related issue(s)

https://issues.serokell.io/issue/OPS-1193

- Tests
        silently reappearing again.

- Documentation
    - Haddock


pasqu4le (approved):
Note: this pipeline definition also needs an update to define steps inter-dependencies (due to the same infra change), but that's already been done in #180

Issue #54, opened 2020-11-17
Make it possible to enable custom token_id


gromakovsky:
# Clarification and motivation

Right now our contract always rejects all operations involving `token_id` other than 0 or 1. However, custom DAOs may have special tokens for their DAO logic. E. g. in GameDAO there can be multiple types of coins and each type may use its own `token_id`. And we may want them to be transferrable using FA2 interface.

# Acceptance criteria

A DAO developer who wants to permit FA2 operations involving `token_id` other than 0 or 1 should have an easy way to do that.

kkirka:
1) Do I understand correctly that token ids other than 0 or 1 can only be used in transfers? e.g., I can't freeze a token of id 2.
2) Can the administrator transfer someone's tokens with id > 1?

pasqu4le:
Answering your question:
1. correct, these other `token_id`s cannot be frozen/unfrozen, they can only be `transfer`red as per FA2.
2. Good question: I think we shouldn't.
    After all these additional tokens do not impact the "main" contract logic in any way and there may be use-cases where being outside of the admin's direct control may be desirable.
    On the other hand those that do want the admin to have control can always implement it by making him an operator, so it should not be a strong limitation for them.

PR #181, opened 2021-04-02
[#170] Byo Extension

### LINKED ISSUE #170: BYO token BaseDAO extension

Requirement: allow holders of a pre-existing token contract to lock up (i.e. wrap) their tokens as DAO tokens inside of the BaseDAO contract.

Example: 

As a token issuer of an FA2 token called `DROP`, I want to create a BaseDAO in which `DROP` can be wrapped as `bDROP` to enable participation in the DAO.

# Acceptance criteria

## Changes to BaseDAO

The (un)freeze entrypoint will perform a FA2 transfer to/from a fixed existing, external, FA2 governance token.

In other words, wrapping `N` tokens will:
1) mint N wrapped (frozen) tokens on the DAO side and
2) transfer N governance tokens from the gov-FA2 to the DAO.

Unwrapping N tokens likewise transfers N governance tokens to the gov-FA2 from the DAO and burns N wrapped tokens.

These entrypoints will replace the explicit freeze/unfreeze entrypoints specified here: https://github.com/tqtezos/baseDAO/issues/140

The DAO contract will no longer internally account for an unfrozen governance token.

sras:
## Description

Implements #170 

1. Freeze entrypoint now calls the token contract to make a transfer to baseDAO address from the address of the caller (right now token id of `0` is used in this call) and mints the corresponding amount of Frozen tokens to the caller here. 
2. Unfreeze does the reverse, and burns the frozen tokens for the caller in this contract.

Tests are modified to include a dummy token FA2 contract that remembers the transfer calls it gets, and to change the expected balances after proposal operations.

Most of the tests should be reviewed more carefully. May be some more checks can be added to the state of the dummy FA2 contract. I shall do it before removing the draft status of this PR.

## Related issue(s)

Resolves #170 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,19 +65,21 @@ endif
 	#
 
 $(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
-$(OUT)/trivialDAO_storage.tz : token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/trivialDAO_storage.tz : governance_token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/trivialDAO_storage.tz : governance_token_id = 0n
 $(OUT)/trivialDAO_storage.tz : now_val = Tezos.now
 $(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage(("$(admin_address)": address), ("$(token_address)": address), $(now_val), $(metadata_map))'
+	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage(("$(admin_address)": address), {address=("$(governance_token_address)": address); token_id=$(governance_token_id)}, $(now_val), $(metadata_map))'
 	# ================= Compilation successful ================= #
 	# See "$(OUT)/trivialDAO_storage.tz" for compilation result  #
 	#
 
 $(OUT)/registryDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
-$(OUT)/registryDAO_storage.tz : token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/registryDAO_storage.tz : governance_token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/registryDAO_storage.tz : governance_token_id = 0n
 $(OUT)/registryDAO_storage.tz : frozen_scale_value = 1n
 $(OUT)/registryDAO_storage.tz : frozen_extra_value = 0n
 $(OUT)/registryDAO_storage.tz : max_proposal_size = 100n
@@ -90,13 +92,14 @@ $(OUT)/registryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/registryDAO_storage.tz: src/**
 	# ============== Compiling RegistryDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/registryDAO_storage.tz src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("$(admin_address)": address), ("$(token_address)": address), ${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value), $(min_xtz_amount), $(min_xtz_amount), $(now_val), $(metadata_map))'
+	$(BUILD_STORAGE) --output-file $(OUT)/registryDAO_storage.tz src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("$(admin_address)": address), {address=("$(governance_token_address)": address); token_id=$(governance_token_id)}, ${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value), $(min_xtz_amount), $(min_xtz_amount), $(now_val), $(metadata_map))'
 	# ================= Compilation successful ================= #
 	# See "$(OUT)/registryDAO_storage.tz" for compilation result #
 	#
 
 $(OUT)/treasuryDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
-$(OUT)/treasuryDAO_storage.tz : token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/treasuryDAO_storage.tz : governance_token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/treasuryDAO_storage.tz : governance_token_id = 0n
 $(OUT)/treasuryDAO_storage.tz : frozen_scale_value = 0n
 $(OUT)/treasuryDAO_storage.tz : frozen_extra_value = 0n
 $(OUT)/treasuryDAO_storage.tz : max_proposal_size = 0n
@@ -109,7 +112,7 @@ $(OUT)/treasuryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/treasuryDAO_storage.tz: src/**
 	# ============== Compiling TreasuryDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/treasuryDAO_storage.tz src/treasuryDAO.mligo base_DAO_contract 'default_treasury_DAO_full_storage(("$(admin_address)": address), ("$(token_address)": address), (${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value), $(min_xtz_amount), $(max_xtz_amount)), $(now_val), $(metadata_map))'
+	$(BUILD_STORAGE) --output-file $(OUT)/treasuryDAO_storage.tz src/treasuryDAO.mligo base_DAO_contract 'default_treasury_DAO_full_storage(("$(admin_address)": address), {address=("$(governance_token_address)": address); token_id=$(governance_token_id)}, (${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value), $(min_xtz_amount), $(max_xtz_amount)), $(now_val), $(metadata_map))'
 	# ============== Compilation successful ============== #
 	# See "$(OUT)/treasuryDAO_storage.tz" for compilation result #
 	#
--- a/haskell/test/Test/Ligo/BaseDAO/FA2.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/FA2.hs
@@ -10,13 +10,13 @@ module Test.Ligo.BaseDAO.FA2
   , lowBalanceScenario
   , noSourceAccountScenario
   , badOperatorScenario
-  , noForeignMoneyScenario
-  , validTransferOwnerScenario
+  -- , noForeignMoneyScenario
+  -- , validTransferOwnerScenario
   , updatingOperatorScenario
   , balanceOfOwnerScenario
   , validateTokenOwnerScenario
   , lowBalanceOwnerScenario
-  , noForeignMoneyOwnerScenario
+  -- , noForeignMoneyOwnerScenario
   , adminTransferScenario
   , adminTransferFrozenScenario
   ) where
@@ -28,21 +28,21 @@ import Lorentz.Test hiding (withSender)
 import Morley.Nettest
 import Util.Named
 
-import Test.Ligo.BaseDAO.Common
 import Ligo.BaseDAO.Types
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+import Test.Ligo.BaseDAO.Common
 
 {-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
 
 zeroTransferScenario :: MonadNettest caps base m => OriginateFn m -> m ()
 zeroTransferScenario originateFn = do
   nonexistent :: Address <- newAddress "nonexistent"
-  ((owner1, _), _, dao, _) <- originateFn
+  ((owner1, _), _, dao, _,  _) <- originateFn
   let params = [ FA2.TransferItem
         { tiFrom = nonexistent
         , tiTxs = [ FA2.TransferDestination
             { tdTo = owner1
-            , tdTokenId = unfrozenTokenId
+            , tdTokenId = frozenTokenId
             , tdAmount = 0
             } ]
         } ]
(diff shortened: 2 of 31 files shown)


src/types.mligo
```diff
@@ -313,8 +313,8 @@ type return_with_full_storage = operation list * full_storage
 
 let nil_op = ([] : operation list)
 
-let unfrozen_token_id: nat = 0n
+let unfrozen_token_id: nat = 1n
```
pasqu4le, line 316: I think this should be removed altogether?

pasqu4le: Same for the `unfrozen_token_id` field in `storage`, which should not longer be necessary.

src/proposal.mligo
```diff
@@ -134,10 +148,25 @@ let stake_tk(token_amount, addr, store : nat * address * storage): storage =
   in { store with freeze_history = new_freeze_history }
 
 [@inline]
-let unfreeze_on_ledger (tokens, addr, ledger, total_supply, unfrozen_token_id, frozen_token_id : nat * address * ledger * total_supply * token_id * token_id): (ledger * total_supply) =
+let unfreeze_on_ledger (tokens, addr, ledger, total_supply, frozen_token_id, token_address : nat * address * ledger * total_supply * token_id * address): (operation * ledger * total_supply) =
+  // Call transfer on token_contract to transfer `token` number of
+  // tokens from `addr` to the address of this contract. Right now token_id `0`
+  // is used, but we should probably confirm this TODO@SRAS
```
pasqu4le, line 154: Indeed we should not assume that `0` is always going to be the correct `token_id`, we should instead have this in the storage and known at origination alongside `token_address`, I'm thinking that we should replace `token_address` with an `governance_token` that itself contains an `address` and `token_id` fields.

src/types.mligo
pasqu4le: Note that there are probably several connected changes, e.g. the `fa2.mligo` code still checks for `unfrozen_token_id` that are no longer necessary/correct

sras: I initially removed it completely. But then we seem to have a lot of tests that required an unfrozen token id. For example to test transfer proposal in `registryDAO`. So added it back with token id of 1.

So if we are to remove it completely, how can we include such tests?

src/proposal.mligo
sras: Sounds good. Have replaced `token_address` with `governance_token` field of type `{address: address, token_id: nat}`, Have changed code, makefile and docs accordingly.

src/types.mligo
sras: @pasqu4le Do you mean that we can consider all tokens except frozen token id as unfrozen implicitly?

src/types.mligo
```diff
@@ -151,10 +151,15 @@ type last_period_change =
 
 // -- Storage -- //
 
+type governance_token =
+  { address : address
+  ; token_id : nat
```
pasqu4le, line 156: LGTM, but we also have the `token_id` type, which is just a type synonym for `nat` (and will result in the same  code, but since we have it let's use that.

Makefile
```diff
@@ -66,18 +66,20 @@ endif
 
 $(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
 $(OUT)/trivialDAO_storage.tz : token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/trivialDAO_storage.tz : token_id = 0n
```
pasqu4le, line 69: Perhaps we should add `governance` to these options as well, otherwise we risk having some ambiguity (if not now soon), I propose simply, `governance_token_address` and `governance_token_id`.

src/types.mligo
pasqu4le: That's a good question, those tests should be handled depending on the cases.

We might soon have other available `token_id` in the contract because of #54 (and those would not count as "frozen"), but that should be left to be handled by the PR for that issue, right now as of this issue description we should not have an unfrozen token in the contract at all.

So for example: the tests you mentioned for transfer proposals in `registryDAO` should indeed be modified.
Because we no longer have an unfrozen token in the DAO itself, transferring `unfrozen_token_id`s for the DAO in there should result in a failure, however transferring to the governance FA2 token should succeed (I see you already have a dummy contract in there, which I think should be used for these tests as well).

Some other tests (and check in the code) no longer make sense from now on.
So if now we have a check that e.g a transfer can be made only on:
1. `unfrozen_token_id`s and 
2. some cases of `frozen_token_id`s (e.g. by `admin`)

then from now on we should only check for (2) and disallow anything else (in the tests as well), including (1).
Note: if some tests seem useful for the #54 let's only comment them and leave a TODO for it.

src/types.mligo
sras: Have changed.

src/types.mligo
sras: I think I have covered this. 

One difficulty was with the tests for update operators call for the baseDAO contract. This right now had to be a always failing code, since frozen token id was prohibited as token id for associated with the operator. Since it appeared that we cannot have always failing code in branches, I have changed it to allow adding operator for frozen tokens as long as the operator is the admin.

Apart from this, I had to comment a couple of tests completely like the ones that tests owner and operator FA2 transfers.

Makefile
sras: Changed

src/types.mligo
pasqu4le: > I have changed it to allow adding operator for frozen tokens as long as the operator is the admin

Why was that necessary? `admin` was already always able to act as an "operator" for anyone because it already had freedom to transfer `frozen_token`s.

haskell/test/Test/Ligo/BaseDAO/FA2.hs
```diff
@@ -31,88 +31,90 @@ import Lorentz.Test hiding (withSender)
...
       { briOwner = owner2
-      , briTokenId = unfrozenTokenId
+      , briTokenId = frozenTokenId
       } ] consumer)
 
   checkStorage (AddressResolved $ toAddress consumer)
-    (toVal [[((owner2, 0 :: Natural), 110 :: Natural)]] )
+    (toVal [[((owner2, frozenTokenId), 110 :: Natural)]] )
+
+-- Disabling this test since there is no unfrozen token available to test it.
```
pasqu4le, line 84: Please add a `TODO #54` on kind of tests, as they'll need to be reactivated by that issue.

haskell/test/Test/Ligo/BaseDAO/Integrational/Common.hs
```diff
@@ -10,18 +10,28 @@ import Universum
...
+  tokenContract <- lOriginate dummyFA2Contract "TokenContract" [] (toMutez 0)
   let addresses = take addrCount $ NE.toList $ genesisAddresses
-  let storage = storageFn addresses
+  let storageInitial = storageFn addresses
+  let storage = storageInitial { fsStorage = (fsStorage storageInitial)
+    { sGovernanceToken = GovernanceToken
+        { gtAddress = unTAddress tokenContract
+        , gtTokenId = FA2.TokenId 0
+        }
+    } }
```
pasqu4le, line 34: Why not have `storageFn` take care of changing the `GovernanceToken` as well?

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -478,7 +482,7 @@ tokenTransferType contractAddr fromAddr toAddr = DAO.Token_transfer_type DAO.Tok
       , tiTxs =
         [ FA2.TransferDestination
           { tdTo = toAddr
-          , tdTokenId = unfrozenTokenId
+          , tdTokenId = frozenTokenId
```
pasqu4le, line 485: Am I correct that here and in other places you replaced this to be able to still make transfers on the DAO itself?

AFAIU this limits them to be executed by the `admin` as `from`, which is not very useful and I'd say an uncommon usage (although possible).
Since most transfers are going to happen on the (now-external) governance token I'd say that's what we should test for instead.

haskell/test/Test/Ligo/TreasuryDAO.hs
```diff
@@ -76,31 +76,20 @@ validProposal = uncapsNettest $ withFrozenCallStack do
   withSender (AddressResolved owner1) $
     call dao (Call @"Propose") (ProposeParams proposalSize proposalMeta)
 
-  checkTokenBalance (frozenTokenId) dao owner1 166
-  checkTokenBalance (unfrozenTokenId) dao owner1 34
+  checkTokenBalance frozenTokenId dao owner1 366
```
pasqu4le, line 80: Connected to the previous point: AFAIU this results now changed because there's only transfers of `frozenTokenId` going on, however I think we'll need to have a function similar to `checkTokenBalance` to fetch the status of a `dummyFA2` instead here.

All things considered, as far as tests are concerned, we might just want to originate a dummy FA2 contract every time we originate a DAO, just so to have it available in the tests themselves. What do you think?

haskell/test/Test/Ligo/BaseDAO/FA2.hs
```diff
@@ -269,71 +269,71 @@ validateTokenOwnerScenario originateFn = do
...
-        { tiFrom = owner2
-        , tiTxs = [ FA2.TransferDestination
-            { tdTo = owner1
-            , tdTokenId = unfrozenTokenId
-            , tdAmount = 10
-            } ]
-        } ]
-  withSender (AddressResolved owner1) $ call dao (Call @"Transfer") params
-    & expectCustomError_ #fA2_NOT_OPERATOR
+-- Disabled for now since there is no unfrozen token available
```
michaeljklein, line 283: Please add a `TODO` as in https://github.com/tqtezos/baseDAO/pull/181/files#r608147607

src/proposal.mligo
```diff
@@ -112,12 +112,25 @@ let check_proposal_limit_reached (config, propose_params, store : config * propo
...
+  // Call transfer on token_contract to transfer `token` number of
+  // tokens from `addr` to the address of this contract.
+  let
+    transfer_param : transfer_params =
+      [{ from_ = addr; txs = [ { to_ = Tezos.self_address; token_id = governance_token.token_id; amount = tokens } ] }] in
+  let token_contract =
+    begin
+      match (Tezos.get_entrypoint_opt "%transfer" governance_token.address : ((transfer_params contract) option)) with
+        | Some (c) -> c
+        | None -> (failwith "Token contract address failed typecheck" : (transfer_params contract))
```
michaeljklein, line 126: Please convert this error (and its occurrence in `unfreeze_on_ledger`) to the same form as the other errors and add documentation, e.g. as with `NOT_ENOUGH_FROZEN_TOKENS`.

michaeljklein (changes requested):

haskell/test/Test/Ligo/BaseDAO/Integrational/Common.hs
sras: `storageFn` was pure as of now, so couldn't orignate the FA2 contract. Also the need for it didn't arise yet. But I have changed it now anyway, since the change required was small.

haskell/test/Test/Ligo/TreasuryDAO.hs
sras: >fetch the status of a dummyFA2 instead here

Here the proposal is not actually flushed, so the transfer does not actually happen. The only transfer call the token contract get is the token transfer for `freeze` call. And the test AFAIU tests that we can raise a valid proposal. So is there any point in checking the token contract state here. We have other tests for BaseDAO contract that check that the transfer in proposal is actually executed by checking the token contract state to ensure that it received a transfer call as expected.

>we might just want to originate a dummy FA2 contract every time we originate a DAO, just so to have it available in the tests themselves. What do you think?

We are actually doing that right now and the `OriginateFn` type alias has been changed to include the FA2 Contract's address in the tuple it returns. 

https://github.com/tqtezos/baseDAO/blob/sras/%23170-byo-extension/haskell/test/Test/Ligo/BaseDAO/Common.hs#L44

haskell/test/Test/Ligo/BaseDAO/FA2.hs
sras: Done.

haskell/test/Test/Ligo/RegistryDAO.hs
sras: I have changed this to a hard coded token id of `0`.

haskell/test/Test/Ligo/BaseDAO/FA2.hs
sras: Done.

src/proposal.mligo
sras: Have renamed the error to `BAD_TOKEN_CONTRACT`.

michaeljklein (approved):
LGTM once @pasqu4le's comments are resolved

src/proposal.mligo
```diff
@@ -134,10 +147,24 @@ let stake_tk(token_amount, addr, store : nat * address * storage): storage =
...
+  let
+    transfer_param : transfer_params =
+      [{ from_ = Tezos.self_address; txs = [ { to_ = addr; token_id = governance_token.token_id; amount = tokens } ] }] in
+  let token_contract =
+    begin
+      match (Tezos.get_entrypoint_opt "%transfer" governance_token.address : ((transfer_params contract) option)) with
+        | Some (c) -> c
+        | None -> (failwith "BAD_TOKEN_CONTRACT" : (transfer_params contract))
+    end in
+  let operation = Tezos.transaction transfer_param 0mutez token_contract in
```
pasqu4le, line 162: A good part of this code is duplicated here in `freeze_on_ledger` as well as in `transfer_contract_tokens` (in `token.mligo`).

We should abstract this to a function, it being used in several places will cause Ligo to compile it into a lambda and save some costs.

IMO we can replace the `FAIL_TRANSFER_CONTRACT_TOKENS` error in `transfer_contract_tokens` with `BAD_TOKEN_CONTRACT` as well as they effectively are the same thing (remember to update the spec tho).

haskell/test/Test/Ligo/TreasuryDAO.hs
```diff
@@ -213,7 +198,7 @@ tokenTransferType contractAddr fromAddr toAddr = Token_transfer_type TokenTransf
       { tiFrom = fromAddr
       , tiTxs = [ FA2.TransferDestination
           { tdTo = toAddr
-          , tdTokenId = unfrozenTokenId
+          , tdTokenId = frozenTokenId
```
pasqu4le, line 201: For the same reason as in `RegistryDAO` I think this should be changed from to be `FA2.TokenId 0`.

Actually even better I think it's more clear if we use `FA2.theTokenId` in both places instead.

haskell/test/Test/Ligo/TreasuryDAO.hs
pasqu4le: Ah I was confused and didn't check properly, thanks for the explanation.

> And the test AFAIU tests that we can raise a valid proposal. So is there any point in checking the token contract state here

I agree, no point in doing so, let's keep them like this :ok_hand:

haskell/test/Test/Ligo/TreasuryDAO.hs
sras: I have change this where possible.

src/proposal.mligo
sras: Have changed.

pasqu4le (approved):
Only one minor nitpick, feel free to squash it with the rest of the changes.

Speaking of which, many of the commit messages don't have the issue "tag" and none have the `Problem`/`Solution` format, please fix these too.

lorentz/docs/specification.md
```diff
@@ -204,7 +204,6 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 | `NOT_ADMIN`                     | The sender is not the administrator                                                                         |
 | `NOT_PENDING_ADMIN`             | Authorized sender is not the current pending administrator                                                  |
 | `NOT_TOKEN_OWNER`               | Trying to configure operators for a different wallet which sender does not own                              |
-| `FAIL_TRANSFER_CONTRACT_TOKENS` | Trying to cross-transfer DAO tokens to another contract that does not exist or is not a valid FA2 contract. |
```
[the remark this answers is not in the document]
sras: Have added it back.

veqtor:
Any documentation on this?
Looking to deploy this for an existing FA2 token

PR #182, opened 2021-04-05
[#173] Remove migration functionality

### LINKED ISSUE #173: Remove migration entrypoints/storage

Full migration of a DAO would require extensive migrations of the ledgers, users, proposals, etc.
This can effectively only be done "lazily", i.e. on a per-user basis, but:
- Such a lazy migration is possible without explicit migration functionality
- In this case, there's no need to lock the contract
- TreasuryDAO may transfer all funds and RegistryDAO can set registry entries to indicate migration

# Acceptance criteria

Remove the following entrypoints, along with their storage values and related functionality:
- `migrate`
- `confirm_migration`

PruStephan:
## Description

Removing migration entrypoints, because in future all migrations will happen manually

## Related issue(s)

Resolves #173 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Ligo/BaseDAO/Management.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Management.hs
@@ -6,8 +6,6 @@
 -- `withOriginated` function
 module Test.Ligo.BaseDAO.Management
   ( test_BaseDAO_Management
-
-  , expectMigrated
   ) where
 
 import Universum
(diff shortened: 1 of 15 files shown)


template/templateDAO.cabal
```diff
@@ -0,0 +1,35 @@
+cabal-version: 2.2
```
pasqu4le, line 1: Not sure why you added this (I guess it should be git-ignored too perhaps), but it is not necessary.

src/types.mligo
```diff
@@ -242,8 +235,6 @@ type forbid_xtz_params =
   | Drop_proposal of proposal_key
```
pasqu4le, line 235: As I was suggesting elsewhere, I think that `migratable_parameter` (and its comment) should no longer exist, and instead be "merged" into the `parameter` itself.

So IMO parameter should look like:
```ocaml
type parameter =
  (allow_xtz_params, "", forbid_xtz_params, "") michelson_or
```
where `migratable_parameter` no longer exists and `forbid_xtz_params` stays the same, but `allow_xtz_params` also contains `transfer_contract_tokens_param`:
```ocaml
type allow_xtz_params =
  | CallCustom of custom_ep_param
  | Propose of propose_params
  | Transfer_contract_tokens of transfer_contract_tokens_param
```
(because it allows xtz transfers).

src/base_DAO.mligo
```diff
@@ -53,14 +51,7 @@ let allowing_xtz (param, store, config : allow_xtz_params * storage * config) =
  *)
 let requiring_no_migration (param, store, config : migratable_parameter * storage * config)
```
pasqu4le, line 52: note that with the other suggestion this can be merged into `base_DAO_contract` (as there are no longer entrypoints requiring no migration). Indeed its comment above is already no longer accurate.

pasqu4le (commented):
Ligo part mostly looks good to me (aside from a couple comments).

Will wait for the haskell/tests update as well

template/templateDAO.cabal
PruStephan: I added it, because one of the checks mantioned, that I should update cabal files

pasqu4le: I think it's a reminder on your system from a previous version of the repo.

As you can see in `master` there is no longer a `./template` dir, because it has been moved in `./lorentz/template/`, so you can just delete this file (as well as the `template` line you added to the `.gitignore`

src/types.mligo
```diff
@@ -262,22 +253,17 @@ type forbid_xtz_params =
...
- *)
-type migratable_parameter =
-  (allow_xtz_params, "", forbid_xtz_params, "") michelson_or
+  | Transfer_contract_tokens of transfer_contract_tokens_param
 
 (*
  * Full parameter of the contract.
  * Made up of entrypoints that could get migrated, and 'Transfer_contract_tokens'
  * which should keep working even after migration.
  *)
```
pasqu4le, line 262: This should be updated:
```suggestion
(*
 * Full parameter of the contract.
 * Separated into entrypoints that forbid Tz transfers,
 * and those that allow Tz transfers
 *)
```

src/base_DAO.mligo
```diff
@@ -75,8 +55,8 @@ let base_DAO_contract (param, full_store : parameter * full_storage)
   let store : storage = full_store.0 in
   let config : config = full_store.1 in
   let (ops, new_store) = match param with
-    // entrypoints that require the contract to not be migrated
-    | M_left p -> requiring_no_migration (p, store, config)
-    // entrypoint that also work when the contract has been migrated
-    | M_right p -> transfer_contract_tokens(p, store)
+    // The entrypoint that allows tz to be send was called
+    | M_left p -> allowing_xtz (p, store, config)
+    // The entrypoint that require no tz to be send was called
```
pasqu4le, line 60:
```suggestion
    // An entrypoint that allows tz to be sent was called
    | M_left p -> allowing_xtz (p, store, config)
    // An entrypoint that require no tz to be sent was called
```
nitpick, just grammar

template/templateDAO.cabal
PruStephan: Is is strange though, that I don't have this folder. locally and I also don't see it in my branch here, but the check is still failing

pasqu4le: Which one? The `./template` one? You should indeed not have that :thinking: 

As for the check it seems that it's working correctly as in [your latest CI run](https://buildkite.com/serokell/basedao/builds/1565#e67ef4b4-8364-4448-92a4-c40b6f4db382) the cabal step is passing.

pasqu4le (changes requested):
Only one comment about tests, however the `docs/specification.md` needs to be updated as well to remove the migration-related documentation.

One last thing is also that there's Error instances for migration in `Ligo.BaseDAO.Types` that should now be removed.

haskell/test/Test/Ligo/BaseDAO/Management.hs
```diff
@@ -82,70 +80,7 @@ test_BaseDAO_Management =
...
-
-       , nettestScenarioCaps "throws error when there is no pending owner" $ do
-          now <- getNow
-          noPendingAdmin withOriginated (initialStorage now)
-
-       , nettestScenarioCaps "throws error when called by current admin, when pending owner is not the same" $ do
-          now <- getNow
-          pendingOwnerNotTheSame withOriginated (initialStorage now)
-
-      , nettestScenarioCaps "Respects migration state" $ do
```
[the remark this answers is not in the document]
PruStephan: Resolved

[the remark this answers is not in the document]
PruStephan: What about the typescript files?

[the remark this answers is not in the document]
pasqu4le: Yes you are right, the typescript files should be updated as well as the parameter has changed.

We should have a check for this, I'll open a new issue to add that :ok_hand:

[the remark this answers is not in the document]
pasqu4le: Found a small issue with our Makefile, so I pushed a fix with the changes as well, check it out.

Also opened #184 to avoid this from happening again

pasqu4le:
One last nitpick: in the commit message of c352894a4a53ca75706cdb156adce1b107e437df there's a `.` in the middle of the sentence that should be at the end (in the `Problem`).

Also, two of the most recent commits do not have a `Problem`/`Solution` in their message.

Feel free to fix that and squash your other commits as well, everything else LGTM.

pasqu4le (approved):
The `.` I mentioned above is still in the current commit message, please fix that before merging :ok_hand:

Issue #183, opened 2021-04-06
Add more option to the Makefile storage targets


pasqu4le:
# Clarification and motivation

We've been asked to add some more storage and configuration options to our `Makefile` target for DAO intial storages for:
- `fixed_proposal_fee_in_token`
- `ledger`
- `quorum_threshold`
- `voting_period`

Note that these are for any DAO, so we'll need to add them to the basic to all the targets (`$(OUT)/trivialDAO_storage.tz`, `$(OUT)/registryDAO_storage.tz` and `$(OUT)/treasuryDAO_storage.tz`) as well as the functions that are used to build these storages.

Additionally keep in mind that these options are inter-depended to other part of the storage for the correct functioning of the contract, namely:
- `ledger` if non empty should also cause `total_supply` to have entries (correctly calculated)
- `quorum_threshold` and `voting_period` have a `min_*` and `max_*` value as well, so we should consider adding options for these too or make some check to verify that the provided values are indeed in the limits.

# Acceptance criteria

All of the options above are added to the initial storage `Makefile` targets

pasqu4le:
Resolved by #185

PR #185, opened 2021-04-07
[#183] Add more options to Makefile

PruStephan:
## Description
We were asked to add some additional parameters to Makefile configuration for all our DAO's

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,10 +69,34 @@ $(OUT)/trivialDAO_storage.tz : governance_token_address = tz1QozfhaUW4wLnohDo6yi
 $(OUT)/trivialDAO_storage.tz : governance_token_id = 0n
 $(OUT)/trivialDAO_storage.tz : now_val = Tezos.now
 $(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
+$(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
+$(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
+$(OUT)/trivialDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
+$(OUT)/trivialDAO_storage.tz : voting_period = 11n
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage(("$(admin_address)": address), {address=("$(governance_token_address)": address); token_id=$(governance_token_id)}, $(now_val), $(metadata_map))'
+	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz \
+      src/base_DAO.mligo base_DAO_contract "default_full_storage( \
+        { storage_data = \
+          { admin = (\"$(admin_address)\" : address) \
+          ; governance_token = \
+            { address = (\"$(governance_token_address)\" : address) \
+            ; token_id = $(governance_token_id) \
+            } \
+          ; now_val = $(now_val) \
+          ; metadata_map = $(metadata_map) \
+          ; ledger_lst = $(ledger) \
+          ; quorum_threshold = $(quorum_threshold) \
+          ; voting_period = $(voting_period) \
+          } \
+        ; config_data = \
+          { max_quorum = {numerator = 99n; denominator = 100n} \
+          ; min_quorum = {numerator = 1n; denominator = 100n} \
+          ; max_period = 2592000n \
+          ; min_period = 1n \
+          } \
+        })"
 	# ================= Compilation successful ================= #
 	# See "$(OUT)/trivialDAO_storage.tz" for compilation result  #
 	#
@@ -89,10 +113,42 @@ $(OUT)/registryDAO_storage.tz : min_xtz_amount = 0mutez
 $(OUT)/registryDAO_storage.tz : max_xtz_amount = 100mutez
 $(OUT)/registryDAO_storage.tz : now_val = Tezos.now
 $(OUT)/registryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
+$(OUT)/registryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
+$(OUT)/registryDAO_storage.tz : ledger = ([] : ledger_list)
+$(OUT)/registryDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
+$(OUT)/registryDAO_storage.tz : voting_period = 11n
 $(OUT)/registryDAO_storage.tz: src/**
 	# ============== Compiling RegistryDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/registryDAO_storage.tz src/registryDAO.mligo base_DAO_contract 'default_registry_DAO_full_storage(("$(admin_address)": address), {address=("$(governance_token_address)": address); token_id=$(governance_token_id)}, ${frozen_scale_value}, $(frozen_extra_value), $(max_proposal_size), $(slash_scale_value), $(slash_division_value), $(min_xtz_amount), $(min_xtz_amount), $(now_val), $(metadata_map))'
+	$(BUILD_STORAGE) --output-file $(OUT)/registryDAO_storage.tz \
+      src/registryDAO.mligo base_DAO_contract "default_registry_DAO_full_storage( \
+        { base_data = \
+          { storage_data = \
+            { admin = (\"$(admin_address)\" : address) \
+            ; governance_token = \
+              { address = (\"$(governance_token_address)\" : address) \
+              ; token_id = $(governance_token_id) \
+              } \
+            ; now_val = $(now_val) \
+            ; metadata_map = $(metadata_map) \
+            ; ledger_lst = $(ledger) \
+            ; quorum_threshold = $(quorum_threshold) \
+            ; voting_period = $(voting_period) \
+            } \
+          ; config_data = \
+            { max_quorum = {numerator = 99n; denominator = 100n} \
+            ; min_quorum = {numerator = 1n; denominator = 100n} \
+            ; max_period = 2592000n \
+            ; min_period = 1n} \
+            } \
+          ; frozen_scale_value = $(frozen_scale_value) \
+          ; frozen_extra_value = $(frozen_extra_value) \
+          ; max_proposal_size = $(max_proposal_size) \
+          ; slash_scale_value = $(slash_scale_value) \
+          ; slash_division_value = $(slash_division_value) \
+          ; min_xtz_amount = $(min_xtz_amount) \
+          ; max_xtz_amount = $(max_xtz_amount) \
+          })"
 	# ================= Compilation successful ================= #
 	# See "$(OUT)/registryDAO_storage.tz" for compilation result #
 	#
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -68,10 +68,14 @@ $(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE
 $(OUT)/trivialDAO_storage.tz : token_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
 $(OUT)/trivialDAO_storage.tz : now_val = Tezos.now
 $(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
+$(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
+$(OUT)/trivialDAO_storage.tz : ledger = (Big_map.empty : ledger)
```
pasqu4le, line 72: We should calculate the correct `total_supply` for a given ledger.

In order to be able to do so we need to `fold` over the parameter here, but that's not possible to do with a `big_map`, so my suggestion is to take a `list` instead for this option and then in the Ligo code `List.fold` over it to create the actual `ledger` `Big_map` and the `total_supply` as well.

src/defaults.mligo
```diff
@@ -3,29 +3,45 @@
...
+  then a
+  else b
+
+let min (a, b : nat * nat) : nat =
+  if a < b
+  then a
+  else b
+
+
+let default_storage (admin, token_address, now_val, metadata, ledger, max_quorum, min_quorum, quorum_threshold, max_period, min_period, voting_period : address * address * timestamp * metadata_map * ledger * nat * nat * nat * nat * nat * nat) : storage = {
```
pasqu4le, line 36: One more suggestion about this is that we now have a bit too many parameter to this function (and the corresponding one for `registryDAO` and `treasuryDAO`).

So I suggest to create a new type just to carry this information, like:
```ocaml
type initial_data =
  { admin : address
  ; token_address : address
  ; now_val : timestamp
  <...>
```

In the `Makefile` we could then use it as:
```
	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage({admin = "$(admin_address)"; token_address = "$(token_address)"; now_val = $(now_val); <...> })'
```

src/defaults.mligo
```diff
@@ -3,29 +3,45 @@
 
 #include "types.mligo"
 
-let default_config : config = {
+let default_config (max_quorum, min_quorum, max_voting_period, min_voting_period : nat * nat * nat * nat): config = {
```
pasqu4le, line 6: see the suggestion before, I'd create here an `initial_config_data` type to carry this info (and use it as part of the `initial_data` type as well perhaps).

pasqu4le (commented):
Note: remember to change the description's `Resolves` to point at the issue.

Makefile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: LicenseRef-MIT-TQ
 
 # Ligo executable
-LIGO ?= ligo
+LIGO ?= /Users/stephan/Documents/ligo
```
pasqu4le, line 5: This should probably not be here.

Makefile
```diff
@@ -75,7 +75,7 @@ $(OUT)/trivialDAO_storage.tz : voting_period = 11n
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
-	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage(("$(admin_address)": address), ("$(token_address)": address), $(now_val), $(metadata_map), $(ledger), 1000n, 1n, $(quorum_threshold), 2592000n, 1n, $(voting_period))'
+	$(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz src/base_DAO.mligo base_DAO_contract 'default_full_storage({storage_data = {admin = ("$(admin_address)" : address); token_address = ("$(token_address)" : address); now_val = $(now_val); metadata_map = $(metadata_map); ledger = $(ledger); quorum_threshold = $(quorum_threshold); voting_period = $(voting_period)}; config_data = {max_quorum = 1000n; min_quorum = 1n; max_period = 2592000n; min_period = 1n}})'
```
pasqu4le, line 78:
```suggestion
  $(BUILD_STORAGE) --output-file $(OUT)/trivialDAO_storage.tz \
    src/base_DAO.mligo base_DAO_contract "default_full_storage( \
      { storage_data = \
        { admin = (\"$(admin_address)\" : address) \
        ; token_address = (\"$(token_address)\" : address) \
        ; now_val = $(now_val) \
        ; metadata_map = $(metadata_map) \
        ; ledger = $(ledger) \
        ; quorum_threshold = $(quorum_threshold) \
        ; voting_period = $(voting_period)\
      } \
      ; config_data = \
        { max_quorum = 1000n \
        ; min_quorum = 1n \
        ; max_period = 2592000n \
        ; min_period = 1n \
        } \
      } \
    })"
```
and something similar should be done with the other two below.

pasqu4le: Also I noticed now that we use tabs and as well as 2 spaces for indentation, can you please add a commit to only use 2 spaces instead?

src/registryDAO/types.mligo
```diff
@@ -52,3 +52,14 @@ type lookup_registry_param =
   }
 
 type lookup_registry_view = (registry_key * (registry_value option)) contract
+
+type initial_registryDAO_storage =
+  { trivial_data : initial_data
```
pasqu4le, line 57: `trivial_data` here is not the best name, as `registryDAO` and `trivialDAO` are both (different) instances of baseDAO, a more appropriate name could be `base_data` for example

src/treasuryDAO/types.mligo
```diff
@@ -29,3 +29,8 @@ type treasury_contract_extra =
   * tez // min_xtz_amount
   * tez // max_xtz_amount
   )
+
+type initial_treasuryDAO_storage =
+  { trivial_data : initial_data
+  ; contract_extra : treasury_contract_extra
+  }
```
pasqu4le, line 36: I wonder why you went with a different approach from `registryDAO` here, the `treasury_contract_extra` only use was for configuration, I think it should now be removed in favor of a single:
```suggestion
type initial_treasuryDAO_storage =
  { base_data : initial_data
  ; frozen_scale_value : nat
  ; frozen_extra_value : nat
  ; max_proposal_size : nat
  ; slash_scale_value : nat
  ; slash_division_value : nat
  ; min_xtz_amount : nat
  ; max_xtz_amount : nat
  }
```

PruStephan: I looked at the function definition an tried to use the same value as in definition. There we have all this data passed as one big tuple

Makefile
PruStephan: Oops, forgot about that

src/treasuryDAO/types.mligo
pasqu4le: Oh yeah, my suggestion is to update the definition as well in a way that's similar to registryDAO

Makefile
PruStephan: I am not familiar with Makefile indentation rules, but if I change a tab to 2 spaces or even 4 spaces, the markdown fails, and makefile doesn't compile

pasqu4le: I see you made these changes, but as you can see from CI they don't work.

The reason being that you need to wrap the whole string in `"` and not `'` for `\` to work properly, aka this:
```
      src/base_DAO.mligo base_DAO_contract 'default_full_storage( \
<...>
        })'
```
should be:
```
<...>
        })"
```

Perhaps this is the same reason why tabs/spaces are creating a problem?
AFAIK you should be able to replace each tab with 2 spaces and it should work.

pasqu4le:
Mostly LGTM now but:
1. `Resolves` in the PR description still needs to be updated
2. This is outdated, a few changes will be needed to rebase on `master` 
3. The `docs` should be updated too (I suggest you do this after rebasing tho)

Makefile
rvem: > Also I noticed now that we use tabs and as well as 2 spaces for indentation, can you please add a commit to only use 2 spaces instead?

https://stackoverflow.com/questions/2131213/can-you-make-valid-makefiles-without-tab-characters

pasqu4le: Oh thanks, I didn't know that.

There are however two places where we use four spaces (hence why I assumed it was ok), in the `test` and `typescript` targets, @PruStephan can you change those lines to use two tabs instead?

I suppose that the reason why that works is that they are "continuation" lines, but let's be consistent and use tabs everywhere.

PruStephan: A question about quorum_threshold. I have just rebased on latest master, and there is a change of quorum_threshold type. 
```
type quorum_threshold =
  [@layout:comb]
  { numerator : nat
  ; denominator : nat
  }
```

So the bound checks should check each parameter separately or remove the check at all

pasqu4le: The `quorum_threshold` has become a fraction, but so have the bounds.

They don't have to check each parameter separately, but as the fraction, you should still check that `quorum_threshold` is lower than `max_quorum_threshold` and bigger than `min_quorum_threshold`.

See how `cmp_qt` is used to do this.

src/defaults.mligo
```diff
@@ -2,42 +2,66 @@
...
+
+let max_vp (a, b : voting_period * voting_period) : voting_period =
+  if a > b
+  then a
+  else b
+
+let min_vp (a, b : voting_period * voting_period) : voting_period =
+  if a < b
+  then a
+  else b
```
pasqu4le, line 30: Given that this is only used to bound the `voting_period`, I'd replace it with a single:
```suggestion
let bound_vp (vp, min_vp, Max_vp : voting_period * voting_period * voting_period)
    : voting_period =
  if vp < min_vp
  then min_vp
  else
    if vp > max_vp
    then max_vp
    else vp 
```

src/proposal.mligo
```diff
@@ -79,6 +79,18 @@ let is_le_qt(qt_1, qt_2 : quorum_threshold * quorum_threshold): bool =
...
+let max_qt (a, b : quorum_threshold * quorum_threshold) : quorum_threshold =
+  if is_gt_qt(a, b)
+  then a
+  else b
+
+// Returns the greatests of two given quorum_thresholds
+let min_qt (a, b : quorum_threshold * quorum_threshold) : quorum_threshold =
+  if is_gt_qt(a, b)
+  then b
+  else a
```
pasqu4le, line 92: These aren't really used by this module and I think it's a bit odd to see them here.

I suggest moving them to `defaults` (no change in the importing will be needed) and make a single `bound_qt` (see previous comment).

pasqu4le (commented):
Only minor suggestions, however note that the [previous comment](https://github.com/tqtezos/baseDAO/pull/185#issuecomment-821179445) still applies.

Makefile
```diff
@@ -109,16 +165,49 @@ $(OUT)/treasuryDAO_storage.tz : min_xtz_amount = 0mutez
...
+          ; max_xtz_amount = $(max_xtz_amount) \
+          })"
+
 	# ============== Compilation successful ============== #
 	# See "$(OUT)/treasuryDAO_storage.tz" for compilation result #
 	#
 
 originate : storage = $(OUT)/trivialDAO_storage.tz
-originate : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+originate : admin_address = ("tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" : address)
```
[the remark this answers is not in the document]
pasqu4le: @PruStephan you left the quotes around the string, which is still not correct, as more are added below in the command call, please just revert this.

src/treasuryDAO/types.mligo
```diff
@@ -20,12 +20,13 @@
...
+type initial_treasuryDAO_storage =
+  { base_data : initial_data
+  ; frozen_scale_value : nat
+  ; frozen_extra_value : nat
+  ; max_proposal_size : nat
+  ; slash_scale_value : nat
+  ; slash_division_value : nat
+  ; min_xtz_amount : tez
+  ; max_xtz_amount : tez
+  }
```
pasqu4le, line 32: No eof newline

docs/building.md
```diff
@@ -95,7 +104,15 @@ make out/treasuryDAO_storage.tz \
...
+  voting_period=11n \
+  metadata_map=(Big_map.empty : metadata_map) \
 ```
 
 As for the other examples, all the arguments are optional and will be equal to
-the values above if not specified.
+the values above if not specified. If `ledger` is specified, than `total_supply`
+will be calculated, depending on values in it. If `quorum_threshold` is specified, it will be checked for being inside `[min_quorum_thresold, max_quorum_threshold]` interval.
+If it lies outside, `quorum_threshold` will be set to the nearest bound. The same is true for
+`voting_period`.
```
pasqu4le, line 118: I suggest a _separate_ paragraph for these options:
```suggestion
the values above if not specified. 

## Storage generation checks

The LIGO functions used by the `Makefile` targets above perform some automatic check, specifically:
If `ledger` is specified, than `total_supply` will be calculated, depending on its value.
If `quorum_threshold` is specified, it will be checked for being inside
`[min_quorum_thresold, max_quorum_threshold]` interval.
If it lies outside, `quorum_threshold` will be set to the nearest bound.
The same is true for `voting_period`.

pasqu4le (approved):
Aside from the two suggestions below all LGTM.

PR #186, opened 2021-04-07
[#175] make the quorum a fraction of Total Supply

### LINKED ISSUE #175: Quorum as Fraction of Total Supply

The quorum is currently a number of tokens. If the total supply changes, the quorum will need to be adjusted, however the primary mechanism for updating the quorum is by voting.

Adjusting the quorum to scale with the total supply solves this issue.

# Acceptance criteria

The quorum should be a fraction, stored in one of two formats:
- As `quorum_percentage : nat`
  * `quorum_percentage` is at most `99`
  * Represents the fraction `quorum_percentage / 100`
- As `numerator : nat` and `denominator : nat`
  * `numerator < denominator`
  * Represents the fraction `numerator / denominator`

Since Michelson lacks support for floating point numbers, the comparison should be calculated without division to avoid rounding:

```haskell 
votes_placed * denominator >= total_supply * numerator
```

pasqu4le:
## Description

Modifies the quorum threshold, as well as its settings, to be a fraction of the total supply of frozen tokens (instead of a fixed value).

## Related issue(s)

Resolves #175

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/typescript/baseDAO/src/generated/Set_quorum_threshold.ts
+++ b/typescript/baseDAO/src/generated/Set_quorum_threshold.ts
@@ -1,2 +1,5 @@
 import {Lambda} from '../common';
-export type Set_quorum_threshold = number;
+export interface Set_quorum_threshold {
+  numerator: number;
+  denominator: number;
+};
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -77,9 +77,9 @@ type config =
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   ; max_votes : nat
   // ^ Determine the maximum number of votes associated with a proposal including positive votes
-  ; max_quorum_threshold : nat
+  ; max_quorum_threshold : quorum_threshold
   // ^ Determine the maximum value of quorum threshold that is allowed to be set.
-  ; min_quorum_threshold : nat
+  ; min_quorum_threshold : quorum_threshold
   // ^ Determine the minimum value of quorum threshold that is allowed to be set.
   ; max_voting_period : nat
   // ^ Determine the maximum value of voting period that is allowed to be set.
@@ -93,6 +93,8 @@ type config =
 Note:
 - the `token_metadata` type matches the one defined in FA2.
 - the `proposal` type is defined below.
+- the `quorum_threshold` is expressed as a `nat/nat` fraction of the total supply
+  of frozen tokens, see [set_quorum_threshold](#set_quorum_threshold).
 - `storage` is the storage type of the contract without the configuration.
 - `full_storage` is instead the full storage of the contract, including its configuration,
 which is to say: `type full_storage = storage * config`.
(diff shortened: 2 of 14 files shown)


typescript/baseDAO/src/generated/Set_quorum_threshold.ts
```diff
@@ -1,2 +1,5 @@
 import {Lambda} from '../common';
-export type Set_quorum_threshold = number;
+export interface Set_quorum_threshold {
```
sras, line 2: We should probably also update https://github.com/tqtezos/baseDAO/blob/pasqu4le/%23175-quorum-fraction/typescript/baseDAO/src/api.ts#L203

Most probably it ll have to be something like https://github.com/tqtezos/baseDAO/blob/pasqu4le/%23175-quorum-fraction/typescript/baseDAO/src/api.ts#L188

sras:
LGTM except for one possibly missing Typescript update.

typescript/baseDAO/src/generated/Set_quorum_threshold.ts
pasqu4le: @sras do you mean because of the argument?

If so I ended up finding more such cases and simplified those too.

sras: @pasqu4le Earlier the entrypoint argument was just a number, so this call could work

```
      contract => contract.methods.set_quorum_threshold(arg));
```

But now the entrypoint accept two values, so I think we ll have to pass the numerator and denominator separately as `contract.methods.set_quorum_threshold(arg.qtNumerator, arg.qtDenominator))`.

I am not sure though. If you have this contract deployed somewhere, I can check it.

pasqu4le: Oh I see, I have made the opposite change in the last commit instead.

I don't have a deployed contract to share, if you have the time please originate one and do check this.

sras: Checked with a deployed contract, and the current code gives the following error

```
InvalidParameterError {
  smartContractMethodName: 'set_quorum_threshold',
  sigs: [ [ 'nat', 'nat' ] ],
  args: [ { qtNumerator: 1, qtDenominator: 2 } ],
  name: 'Invalid parameters error',
  message: 'set_quorum_threshold Received 1 arguments while expecting one of the following signatures ([["nat","nat"]])'
}
```

I could get a successful call by modifying the api code for `set_quorum_threshold` as

```
  set_quorum_threshold(arg: Set_quorum_threshold): Promise<string|void> {
    return this.withContract(
      contract => contract.methods.set_quorum_threshold(arg.qtNumerator, arg.qtDenominator));
  }
```

Here is the deployed contract and storage after the call   ```BaseDAO.set_quorum_threshold({qtNumerator: 1, qtDenominator: 2});```


https://edo2net.tzkt.io/KT1HABxCoj7QYoErXa7HX6j3qquTPRMJU19V/storage

pasqu4le: @sras thanks a lot for checking.

I changed it now (and reverted the incorrect changes), I also modified the `HasAnnotation` instance of `QuorumThreshold` to not contain the prefix, so it's `numerator` and `denominator`.

These changes are all in the (only) fixup commits so you can check them.

sras: Np, LGTM now.

sras (approved):

Issue #171, opened 2021-03-20
Voting Power DAO


michaeljklein:
# Clarification and motivation

The instructions `VOTING_POWER` and `TOTAL_VOTING_POWER` allow us to query the individual, total amount of rolls staked by bakers, respectively.

As with the BYO token DAO (https://github.com/tqtezos/baseDAO/issues/170), the (un)freeze entrypoints can be modified to consider the amount staked to be the amount of governance tokens held. In other words, a user with `VOTING_POWER = N` is considered to hold `N` governance tokens. 

# Acceptance criteria

Note: proposal and voting cycles must be aligned with the actual on-chain baking cycles.
This allows the results of `VOTING_POWER` and `TOTAL_VOTING_POWER` to be cached until the next voting/proposing cycle.

## Changes to BaseDAO

- The quorum becomes a fixed ratio or percentage of `TOTAL_VOTING_POWER`
- The duration of proposing and voting cycles are exact multiples of baking cycles (and must fit within 7 cycles)
- The proposal and voting cycles must also be synchronized with the baking cycles, since the `VOTING_POWER` can change at the end of a baking cycle
- There is no ledger of unfrozen governance tokens
- Regarding frozen tokens:
  * Users are only allowed to freeze exactly their `VOTING_POWER` number of tokens
  * This operation (freezing) must do nothing or fail if it’s called more than once before the end of the proposal cycle
  * All frozen tokens are invalidated at the end of the proposal cycle and must be re-frozen (triggering additional calls to `VOTING_POWER`, etc)

sras:
@michaeljklein We have some queries regarding the requirements.

> The proposal and voting cycles must also be synchronized with the baking cycles.

Since Michelson does not provide instructions to query the cycle or cycle position, it appears that we have to use the `level` instruction along with network constants, such as cycle length,  to compute/detect cycle transitions.

Since these constants can change between protocols, it also seems that we need some way for the admin to set these constants in storage.

Does this make sense?

michaeljklein:
@sras This contract shouldn't have an admin once it's actively accepting proposals: any required constant updates should be performed through proposals.

sras:
Some updates to this issue based on past discussion.

1. Where the description says, "The duration of proposing and voting cycles are exact multiples of baking cycles (and must fit within 7 cycles)",  it should fit in 5 cycles instead of 7

2. The limitation that only frozen tokens that were frozen in previous periods can be used to propose or vote is lifted.

PR #190, opened 2021-04-19
[#54] Allow token ids other than zero

### LINKED ISSUE #54: Make it possible to enable custom token_id

kkirka:
## Description
This PR adds support for token ids > 0. The administrator can only transfer users' tokens with id = 0. Since we want to fire `FA2_TOKEN_UNDEFINED`, I modified `validate_token_type` to check whether the token_id is present in the total supply big_map: this way, we can still distinguish supported tokens from unsupported ones, while not storing the set of token ids. Big map access would cost us some gas, though.

The test coverage in this PR is rather shady IMO. The existing tests for FA2 behavior have been evolving with the requirements, and currently, they are hard to make sense of: they test for similar functionality while leaving some code paths untested. However, I didn't want to introduce too many merge conflicts in this PR, as AFAIU other potentially complex PRs are coming soon. Moreover, there is some chance that token functionality will be removed soon. So I just uncommented/adjusted the existing tests and added a separate test case for admin trying to transfer users' tokens. However, I think that in the long run we should consider doing a rehaul to make the test suite better reflect the spec.

## Related issue(s)

Resolves #54 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


src/token/fa2.mligo
```diff
@@ -128,22 +128,22 @@ let balance_of (params, store : balance_request_params * storage): return =
 // Update operators entrypoint
 // -----------------------------------------------------------------
 [@inline]
-let validate_operator_token (token_id, frozen_token_id, is_operator_admin : token_id * token_id * bool): token_id =
-  if (token_id = frozen_token_id) then
+let validate_operator_token (token_id, store, is_operator_admin : token_id * storage * bool): token_id =
```
pasqu4le, line 131: Ok so the:
```
it seems that Michelson won't allow always failing code, so have included this branch
```
in `validate_operator_token` was due to the fact that there was no other valid `token_id` aside from `frozen_token_id` *and*  that was an `OPERATION_PROHIBITED` unless the operator is `admin` (which is always authorized to operate them, so doesn't actually need to be an operator explicitly).

In other terms, these two things make the check absurd, at least until this PR comes along.

Now that there are (potentially) more valid `token_id` we can revert this code and make it check that:
- for `frozen_token_id` no operator is valid (`admin` technically is, but doesn't need the `update_operators` call)
- for every other id, it must be one of the allowed ones

So I think this is the correct implementation now is:
```ocaml
let validate_operator_token (token_id, store : token_id * storage): token_id =
  if (token_id = frozen_token_id)
  then (failwith("OPERATION_PROHIBITED") : token_id)
  else validate_token_type (token_id, store)
```

src/token/fa2.mligo
```diff
@@ -10,8 +10,16 @@
...
+  if Big_map.mem token_id store.total_supply
+  then token_id
+  else
+    ([%Michelson ({| { FAILWITH } |} : string * unit -> token_id)]
+     ("FA2_TOKEN_UNDEFINED", ()) : token_id)
+
+[@inline]
+let check_sender (from_, token_id, store : address * nat * storage): address =
+  if (sender = store.admin && token_id = store.frozen_token_id) then from_
   else if (sender = from_) then from_
```
pasqu4le, line 23: Ah this check is kinda sad because (while it makes sense) it is repeated in `transfer_one`:
```ocaml
        if tx.token_id = store.frozen_token_id then (
          if sender = store.admin then tx.token_id
```
I think we can revert this and only call `check_sender` when we need to, aka when the `token_id` is not `frozen_token_id`.

So, my suggestion is to here just do:
```suggestion
let check_sender (from_, store : address * storage): address =
  if (sender = from_) then from_
```
and change the implementation of `transfer_item` to:
```ocaml
let transfer_item (store, ti : storage * transfer_item): storage =
    if tx.amount = 0n then store
    else
      let (valid_token_id, valid_from_) =
        if tx.token_id = store.frozen_token_id
        then
          if sender = store.admin
          then (tx.token_id, ti.from_)
          else (failwith("FROZEN_TOKEN_NOT_TRANSFERABLE") : token_id * address)
        else (validate_token_type (tx.token_id, store), check_sender (ti.from_, store))
      in
      let ledger = store.ledger in
      let total_supply = store.total_supply in
      let (ledger, total_supply) = debit_from(tx.amount, valid_from_, valid_token_id, ledger, total_supply) in
      let (ledger, total_supply) = credit_to(tx.amount, tx.to_, valid_token_id, ledger, total_supply) in
      { store with
        ledger = ledger
      ; total_supply = total_supply
      }
  in List.fold transfer_one ti.txs store
```

src/defaults.mligo
```diff
@@ -3,6 +3,8 @@
 
 #include "types.mligo"
 
+let frozen_token_id: nat = 0n
```
pasqu4le, line 6: This isn't really selectable tho (even tho it is in storage), well it can't be in all cases anyway.

As per FA2 we need to know that when there is only one `token_id` in the contract this has to be `0`.

We could either check that this is the case in `default_storage` (but than it's hard to say what to do when it is not the case) or not allow the decision at all, which is pretty fair.

IMO we can remove this and in `default_storage`  just set `frozen_token_id = 0n` instead.

(well done in removing this from the `types` module tho)

pasqu4le (commented):
I haven't checked the Haskell part yet as maybe the changes I'm proposing to the Ligo code will change the tests.

It looks like we forgot to properly update the spec in #170, good job on fixing that :ok_hand: 

>  think that in the long run we should consider doing a rehaul to make the test suite better reflect the spec.

Likewise, there have been too many heavy changes lately and this is largely the cause.
#171 should be the last one in the near future, so this will be done after that.

src/defaults.mligo
kkirka: > IMO we can remove this and in `default_storage` just set `frozen_token_id = 0n` instead

I left this as a separate binding because it's used in `total_supply = Map.literal [ (frozen_token_id, 0n) ];`. Although it isn't configurable, having `(0n, 0n)` in total supply harms readability IMO.

src/token/fa2.mligo
kkirka: I like the tuple trick, much cleaner :+1:

Now that I think about it once again, I see some problems still remain:

1. FA2 states that:
> If one of the specified `token_id`s is not defined within the FA2 contract, the entrypoint MUST fail with the error mnemonic `"FA2_TOKEN_UNDEFINED"`.
However, we check for zero transfer before validating the token id (in my current implementation as well). One may attempt to transfer zero to check if a certain token id is present, and we always succeed in this case.

2. The "`token_id` in `total_supply`" check is done three times: in `validate_token_type`, in `debit_from`, and in `credit_to`.
    - I can't remove the check from `debit_from` because `debit_from` actually updates the total supply (and we need to fail if it's `None`).
    - If I remove `validate_token_type`, I'll get `FA2_INSUFFICIENT_BALANCE` instead of `FA2_TOKEN_UNDEFINED`. I can adjust `debit_from` so that it first tries checks the total supply – I think this would be the best option.
    - I can, and I think I **should** remove this check from `credit_to`: it's extraneous in `transfer`, and it's plain wrong in `mint`: some DAO authors may want to mint new tokens with custom token ids, and mint uses `credit_to`.

So I'd rather:
- remove the `tx.amount = 0n` checks
- remove `validate_token_type`
- adjust the code in `debit_from` so that it checks for token type
- treat nonexisting total supply in `credit_to` as zero supply

src/token/fa2.mligo
sras: > So I think this is the correct implementation now is:

Agree.

sras (approved):
LGTM

src/defaults.mligo
pasqu4le: Oh I see, makes sense to me too, no problem then :ok_hand:

src/token/fa2.mligo
pasqu4le: > we check for zero transfer before validating the token id

This has often been a point of contention, in FA1/1.2 as well, although I think it's pretty clear in this case, given also the:
> Transfers of zero amount MUST be treated as normal transfers.

So regarding point 1, I also think we should remove the `tx.amount = 0n` check entirely.

Regarding point 2 instead:
> remove `validate_token_type`

Well we need that in other places as well, but if you only mean in `transfer_one` then yes, it makes sense to me as well :ok_hand: 

> adjust the code in `debit_from` so that it checks for token type

Likewise this also seem a good move, as you said it should be possible to get by moving the check up.

> treat nonexisting total supply in `credit_to` as zero supply

This I think is not correct tho: just as `debit_from` does, the `credit_to` function needs to update the total supply and to do so it requires the `current_total_supply`.
This means that the "check" doesn't really add a cost, as the `Map.find_opt` is needed anyway and I also think that it should fail when trying to increase the supply of an undefined token.

Note: I also think that should be the case for `mint` and not allow admins to add `token_id`s, moreover it seems that we'll likely remove the `mint` entirely, so let's keep this behavior.

haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -682,6 +682,13 @@ instance CustomErrorHasDoc "nOT_OWNER" where
   customErrDocMdCause =
     "The sender of transaction is not owner"
 
+type instance ErrorArg "oPERATION_PROHIBITED" = NoErrorArg
```
pasqu4le, line 685: :thinking: It seems that this is also missing from the `specification.md`, can you add it there please?

haskell/test/Test/Ligo/BaseDAO/Token/Common.hs
```diff
@@ -0,0 +1,72 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
```
pasqu4le, line 3: Somewhere in this file there's a trailing whitespace that is making CI sad.

pasqu4le (approved):
Two minor comments and needs a rebase, otherwise LGTM now.

Feel free to add these changes and merge after that.

PR #194, opened 2021-04-22
[#189] Update `proposal_metadata` to `bytes`

### LINKED ISSUE #189: Consider changing the proposal_metadata type to bytes only

The `proposal_metadata` is currently a `(string, bytes) map`, however that's probably not the best representation.

This type causes the unpacking logic to be quite complex and expensive as it has to:
- do an `UNPACK` call for every field of the proposal type
- keep in mind that different proposal types might have the same field

Additionally this complicates the (possible) case where two proposal types have exactly the same fields, as one would have to use some workaround with different keys or something like that. 

It's also not particularly convenient for users as they'll have to implement/operate more logic to `PACK` individual fields one by one.

So I propose to instead change the field to a `bytes` that needs a single `UNPACK` to obtain a sum-type instead.

# Acceptance criteria

1. `proposal_metadata` is a type synonym for `bytes`
2. there is a single proposal type per DAO implementation, that is a sum-type of all the different proposal types
3. such type is documented in the docs, including its Michelson representation
4. the unpacking logic for `registryDAO` and `treasuryDAO`'s proposal-related lambdas simply extracts that type

rinn7e:
## Description

Problem: The `proposal_metadata` is currently a `(string, bytes) map`,
however it is not the best representation due to unpacking issues.

Solution: Update `proposal_metadata` from `(string, bytes) map` to
`bytes`, make adjustment to all the tests.

## Related issue(s)

Resolves #189

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -43,24 +43,38 @@ However, an existing separate FA2 contract is used for governance.
 BaseDAO is a concrete smart contract, but also a framework to implement various DAOs.
 It can be configured at origination for any specific needs.
 
-In order to do so the contract has types that can contain arbitrary data:
-- `proposal_metadata` which is a type synonym for `(string, bytes) map`
-  (or in Michelson: `map string bytes`)
+In order to do so the contract has types that can contain arbitary data:
+- `proposal_metadata` which is a type synonym for `bytes`
 - `contract_extra` which is a type synonym for `(string, bytes) big_map`
   (or in Michelson: `big_map string bytes`)
 
 
 The former contains fields that are required to submit a proposal.
-The latter keeps global information like information about accepted proposals.
 
-In both cases we associate a `string` "name" to the `pack`ed representation of
+The latter keeps global information like information about accepted proposals.
+In this case, we associate a `string` "name" to the `pack`ed representation of
 the data, that can then be `unpack`ed by the contract code.
 
-For example a "treasury" style DAO would have a `proposal_metadata` containing:
-```
-"mutezAmount": <packed mutez>
-"recipientId": <packed string>
-"agoraPostId": <packed nat>
+For example, the `proposal_metadata` in a "treasury" style DAO would be the packed version of the type `transfer_proposal` :
+```ocaml
+type xtz_transfer =
+  { amount : tez
+  ; recipient : address
+  }
+
+type token_transfer =
+  { contract_address : address
+  ; transfer_list : transfer_item list
+  }
+
+type transfer_type =
+  | Xtz_transfer_type of xtz_transfer
+  | Token_transfer_type of token_transfer
+
+type transfer_proposal =
+  { agora_post_id : nat
+  ; transfers : transfer_type list
+  }
 ```
 and an empty `contract_extra`.
 
@@ -612,7 +626,7 @@ Parameter (in Michelson):
 ### **propose**
 
 ```ocaml
-type proposal_metadata = (string, bytes) map
+type proposal_metadata = bytes
 
 type propose_params =
   { frozen_token : nat
@@ -626,7 +640,7 @@ Parameter (in Michelson):
 ```
 (pair %propose
   (nat %frozen_token)
-  (map %proposal_metadata string bytes)
+  (bytes %proposal_metadata)
 )
 ```
 
--- a/haskell/test/Test/Ligo/BaseDAO/Common.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Common.hs
@@ -7,6 +7,7 @@
 
 module Test.Ligo.BaseDAO.Common
   ( OriginateFn
+  , TransferProposal(..)
   , totalSupplyFromLedger
 
   , mkFA2View
@@ -16,7 +17,6 @@ module Test.Ligo.BaseDAO.Common
   , addDataToSign
   , permitProtect
   , sendXtz
-  , ProposalMetadataFromNum (..)
 
   , createSampleProposal
   , originateLigoDaoWithBalance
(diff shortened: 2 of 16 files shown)


rinn7e:
@pasqu4le Took me quite sometimes to update the types in RegistryDAO tests, but I managed it now. Hope to get your reviews on this.

src/registryDAO.mligo
```diff
@@ -165,11 +122,12 @@ let registry_DAO_decision_lambda (proposal, extras : proposal * contract_extra)
     } in
   let proposal_key = to_proposal_key (propose_param, proposal.proposer) in
   let ops = ([] : operation list) in
-  match extract_proposal(proposal.metadata) with
-  | Normal_proposal diff ->
-      let extras = apply_diff(diff, extras) in
+  let registry_dao_proposal_metadata = unpack_proposal_metadata(proposal.metadata) in
+  match registry_dao_proposal_metadata with
```
pasqu4le, line 126:
```suggestion
  match unpack_proposal_metadata(proposal.metadata) with
```

src/registryDAO.mligo
```diff
@@ -181,12 +139,12 @@ let registry_DAO_decision_lambda (proposal, extras : proposal * contract_extra)
...
-      let m_max_proposal_size = unpack_nat_opt(find_map("max_proposal_size", proposal.metadata)) in
-      let m_slash_scale_value = unpack_nat_opt(find_map("slash_scale_value", proposal.metadata)) in
-      let m_slash_division_value = unpack_nat_opt(find_map("slash_division_value", proposal.metadata)) in
+  | Configuration_proposal cp ->
+      let m_frozen_scale_value = cp.frozen_scale_value in
+      let m_frozen_extra_value = cp.frozen_extra_value in
+      let m_max_proposal_size = cp.max_proposal_size in
+      let m_slash_scale_value = cp.slash_scale_value in
+      let m_slash_division_value = cp.slash_division_value in
       let new_ce = match m_frozen_scale_value with
```
pasqu4le, line 148: The lines above are an unnecessary re-binding now, we can remove them and use the values directly, e.g. here:
```suggestion
      let new_ce = match cp.frozen_scale_value with
```

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -423,8 +470,8 @@ test_RegistryDAO =
 
     defaultTokenBalance :: Natural = 0
 
-    metadataSize :: DynamicRec "pm" -> Natural
-    metadataSize md = fromIntegral $ BS.length $ lPackValueRaw md
+    metadataSize :: ProposalMetadata -> Natural
+    metadataSize md = fromIntegral $ BS.length $ md
```
pasqu4le, line 474:
```suggestion
    metadataSize = fromIntegral . BS.length
```

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -33,6 +33,70 @@ import Ligo.Util
...
+  , cpSlashDivisionValue :: Maybe Natural
+  , cpMaxProposalSize :: Maybe Natural
+  }
+
+instance HasAnnotation ConfigProposal where
+  annOptions = baseDaoAnnOptions
+
+
+-- Note: Somehow setting `[@layout:comb]` in ligo types does not work.
+-- Have to use `ligoLayout` instead.
```
pasqu4le, line 82: Were you able to tell what was going wrong?

Are the types produced by LIGO not respecting the layout selection?

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -138,8 +138,7 @@ instance ProposalMetadataFromNum Integer where
   proposalMetadataFromNum = fromIntegral
 
 instance ProposalMetadataFromNum ProposalMetadata where
```
pasqu4le, line 140: I wonder if this class is still necessary at all.

AFAIK it was intended to simplify the use because it was a `Map`, which is no longer the case, what do you think?

src/treasuryDAO/types.mligo
```diff
@@ -29,3 +29,21 @@ type treasury_contract_extra =
   * tez // min_xtz_amount
   * tez // max_xtz_amount
   )
+
+type normal_proposal =
```
pasqu4le, line 33: This is the same as the `transfer_type` in registryDAO.
So I think its definition could be moved to `src/common/types.mligo` instead.

pasqu4le (commented):
One more thing: the documentation for the DAOs in `docs/treasury.md` and `docs/registry.md` needs to be updated as well.

haskell/test/Test/Ligo/RegistryDAO.hs
rinn7e: I can't really tell what's going on. I make sure to make the ligo and haskell type looks exactly the same, added `[@layout:comb]` to all the related ligo types, but it does not work (it always result in fail to unpack). 

I do check the result type from ligo, it looks fine for me. It is quite trick but if you want me to investigate more into this, I could try.

src/treasuryDAO/types.mligo
rinn7e: Yeah, they are actually the same, I've made this change, as well as other changes, should be good for re-review.

haskell/test/Test/Ligo/RegistryDAO.hs
pasqu4le: No that's ok, was just wondering.

Using the `ligoLayout` is perfectly fine, we'll get back to this later if we need to.

docs/specification.md
```diff
@@ -38,8 +38,7 @@ BaseDAO is a concrete smart contract, but also a framework to implement various
 It can be configured at origination for any specific needs.
 
 In order to do so the contract has types that can contain arbitary data:
-- `proposal_metadata` which is a type synonym for `(string, bytes) map`
-  (or in Michelson: `map string bytes`)
+- `proposal_metadata` which is a type synonym for `bytes`
 - `contract_extra` which is a type synonym for `(string, bytes) big_map`
   (or in Michelson: `big_map string bytes`)
```
pasqu4le, line 44: A few lines below there's this:
```
For example a "treasury" style DAO would have a `proposal_metadata` containing:
```
which I think should be updated, as the `proposal_metadata` no longer contains `packed mutez` for each key, but the packed value for several fields. Can you update that example?

pasqu4le (commented):
Only one comment on the spec, I also still have the [previous request](https://github.com/tqtezos/baseDAO/pull/194#pullrequestreview-643189475), which is to also update the `docs/registry.md` and `docs/treasury.md`.

Both contain a brief description of the proposal metadata, but I think that now we need to also include the compiled Michelson type in there.

docs/specification.md
rinn7e: I have updated this now, I believe `docs/registry.md` and `docs/treasury.md` are mainly for the Lorentz version, so some of the information are still out of sync if comparing to the ligo implementation.

pasqu4le: > are mainly for the Lorentz version

That should no longer be the case and they should be based on the Ligo version.
I see that you have updated them now and I think it's enough for this PR :ok_hand:

pasqu4le (approved):

PR #195, opened 2021-04-23
[#170] Update build instructions to include governance_token

### LINKED ISSUE #170: BYO token BaseDAO extension

sras:
## Description

In the PR that implements BYO extension, we added Makefile
arguments `governance_token_address` and `governance_token_id`. But the
build instructions was not updated to reflect this. This PR updates
 `docs/building.md` file to include the new arguments in
`make` command samples. In addition the default token address was
changed to an address of the form `KT*` instead of `tz*`, to better
reflect the fact that it is an address of a smart contract. 

## Related issue(s)

Resolves #170

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- [x] My commits comply with [the following policy](https://www.notion.so/serokell/Commit-and-PR-olicy-4cf98e1a910a415d86b5f2491d9af1af).

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -19,7 +19,9 @@ The contract described here consists of two parts:
 - Token functionality ([FA2-based][FA2]).
 - DAO functionality with proposals and voting.
 
-These two parts are coupled into one smart contract because interaction between smart contracts in Tezos is expensive and hard to get right.
+These two parts are coupled into one smart contract to reduce interactions
+between smart contracts, which is expensive and hard to get right in Tezos.
+However, an existing separate FA2 contract is used for governance.
 
 # General Requirements
 
@@ -29,6 +31,10 @@ These two parts are coupled into one smart contract because interaction between 
 
 - The contract may store tokens with other token identifiers.
 
+- The contract must store a 'governance token'. This consist of the address of
+  an FA2 contract and a token id in that contract. The 'governance token'  is
+  used as part of the freeze/unfreeze process by making FA2 transfers on it.
+
 - The storage of the contract must have annotations for all fields
   and must be documented to make its interpretation easy for users.
 
@@ -37,7 +43,7 @@ These two parts are coupled into one smart contract because interaction between 
 BaseDAO is a concrete smart contract, but also a framework to implement various DAOs.
 It can be configured at origination for any specific needs.
 
-In order to do so the contract has types that can contain arbitary data:
+In order to do so the contract has types that can contain arbitrary data:
 - `proposal_metadata` which is a type synonym for `(string, bytes) map`
   (or in Michelson: `map string bytes`)
 - `contract_extra` which is a type synonym for `(string, bytes) big_map`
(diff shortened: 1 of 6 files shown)


pasqu4le (commented):
@sras the `docs/specification.md` also needs an update.

In particular:
1.There are still mentions of the unfrozen tokens (please see the changes in #190 to cause as little conflicts as possible with it)
2. There are no mention of the external FA2 governance token, so that needs to be included too.

docs/specification.md
```diff
@@ -221,8 +236,6 @@ The list of erros may be inaccurate and incomplete, it will be updated during th
 | `NOT_TOKEN_OWNER`               | Trying to configure operators for a different wallet which sender does not own                              |
 | `FAIL_PROPOSAL_CHECK`           | Throws when trying to propose a proposal that does not pass `proposalCheck`                                 |
 | `FROZEN_TOKEN_NOT_TRANSFERABLE` | Transfer entrypoint is called for frozen token by a non-admin sender                                        |
-| `PROPOSAL_INSUFFICIENT_BALANCE` | Throws when trying to propose a proposal without having enough unfrozen token                               |
-| `VOTING_INSUFFICIENT_BALANCE`   | Throws when trying to vote on a proposal without having enough unfrozen token                               |
```
pasqu4le, line 225: It looks like we still have error definitions for both of these in the `Ligo.BaseDAO.Types` module, can you remove them?

docs/specification.md
```diff
@@ -29,6 +29,10 @@ These two parts are coupled into one smart contract because interaction between
```
pasqu4le, line 29: A few lines above, in the overview there is:
```
These two parts are coupled into one smart contract because interaction between smart contracts in Tezos is expensive and hard to get right.
```
I think we should change it to:
```
These two parts are coupled into one smart contract to reduce interactions between smart contracts, which is expensive and hard to get right in Tezos.
However, an existing separate FA2 contract is used for governance.
```

docs/specification.md
```diff
@@ -174,7 +180,18 @@ Everyone can make a new proposal, however, you have to freeze some tokens for th
...
+For freezing the address should have corresponding amount of tokens of proper
+token type (token_id of storage.governance_token.token_id) in the governance
+FA2 contract. During 'Freeze' operation the BaseDAO contracts makes an FA2
+transfer on the governance contract to transfer tokens to its own address from
+the address who is freezing the tokens. Then it mints frozen tokens for the address.
+
+Unfreezing does it reverse, that is, the contract makes the FA2 transfer on governance
+contract to transfer tokens from its own address to the address that is doing unfreezing.
+It then burns the corresponding amount of tokens. Only frozen tokens that are not currently
+staked in a vote or proposal can be unfreezed.
```
pasqu4le, line 192: I think that this would be better in the `Stages` part just above, as it's not just about proposals.

sras: Yes, moved.

docs/specification.md
sras: Removed.

docs/specification.md
sras: Have changed.

pasqu4le (approved):
LGTM now

PR #197, opened 2021-04-26
[#192] Change the `custom_entrypoints` type to `big_map`

### LINKED ISSUE #192: Change the custom_entrypoints type to be a big_map

This issue is a sibling to #188 as the idea is the same: custom entrypoints should be stored into a `big_map` and not a `map`, as this way we can avoid the costs of evaluating the whole map every time and instead take advantage of the laziness of `big_map`s.

# Acceptance criteria

The `custom_entrypoints` type is a `(string, bytes) big_map` and the Haskell type is updated accordingly.

rinn7e:
## Description

Problem: The `custom_entrypoints` is currently defined as a `(string, bytes) map`.
This however poses a limitation, as `map` is strict and so the cost
in case of many key is too high.

Solution: Make `custom_entrypoints` as `big_map` instead of map, make
`DynamicRec` using `BigMap` instead, remove `DynamicRecBigMap`.

## Related issue(s)

Resolves #192 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -78,7 +78,7 @@ type transfer_proposal =
 ```
 and an empty `contract_extra`.
 
-Lastly, there is one more `(string, bytes) map` type synonym: `custom_entrypoints`,
+Lastly, there is one more `(string, bytes) big_map` type synonym: `custom_entrypoints`,
 used to execute arbitrary logic on the contract, see [its section](#custom-entrypoints)
 for more information on its content and usage.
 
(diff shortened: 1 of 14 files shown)


pasqu4le (commented):
Only one comment: there's a mention in the `specification.md` that needs to be updated, otherwise LGTM

rinn7e:
> Only one comment: there's a mention in the `specification.md` that needs to be updated, otherwise LGTM

I totally missed that, updated now

pasqu4le (approved):
LGTM, needs a rebase (I suggest waiting for #189 to be closed first tho)

PR #199, opened 2021-04-28
[#92] Add `governance_token` off-chain view

### LINKED ISSUE #92: Add token_address off-chain view

We used to have a `token_address` entrypoint, but we removed it in #85 because it's not very useful. We can instead make it an off-chain view (according to TZIP-16).

# Acceptance criteria

Metadata printed by the `print-metadata` command includes `token_address` off-chain view.

rinn7e:
## Description

Problem: We used to have a `token_address` entrypoint, but we removed
it in #85 because it's not very useful. We can instead make it an
off-chain view (according to TZIP-16).

Solution: Add off-chain view `token_address`, rename it to
`governance_token` view instead, add tests to cover this
view.

## Related issue(s)

Resolves #92 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
+++ b/haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
@@ -16,6 +16,7 @@ module Ligo.BaseDAO.TZIP16Metadata
   , tokenMetadataView
   , getTotalSupplyView
   , permitsCounterView
+  , governanceTokenView
   ) where
 
 import qualified Universum as U
@@ -86,6 +87,7 @@ baseDAOViews = U.sequence
   , isOperatorView
   , tokenMetadataView
   , getTotalSupplyView
+  , governanceTokenView
 
   , permitsCounterView
   ]
@@ -199,6 +201,20 @@ getTotalSupplyView MetadataSettings{ msConfig = MetadataConfig{} } = View
             unsafeCompileViewCode $ WithParam @FA2.TokenId $ do
               stGet #sTotalSupply
               ifSome nop $ failCustom_ #fA2_TOKEN_UNDEFINED
+      ]
+  }
+
+governanceTokenView :: DaoView
+governanceTokenView MetadataSettings{} = View
+  { vName = "governance_token"
+  , vDescription = Just
+      "Return the address and token_id of the associated FA2 contract."
+  , vPure = Just True
+  , vImplementations =
+      [ VIMichelsonStorageView $
+          mkMichelsonStorageView @Storage Nothing [] $
+            unsafeCompileViewCode $ WithoutParam $ do
+              stToField #sGovernanceToken
       ]
   }
 
--- a/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
@@ -111,4 +111,9 @@ mkFA2Tests storage mc = testGroup "FA2 off-chain views"
         checkTotalSupply storage frozenTokenId
           @?= Right 200
     ]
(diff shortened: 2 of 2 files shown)


rinn7e:
I've added the off-chain view of `tokenAddress` now, not sure if there is documentation that is needed to be updated or not. Hope to get your review on this @pasqu4le

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
```diff
@@ -205,6 +207,20 @@ getTotalSupplyView MetadataSettings{ msConfig = MetadataConfig{} } = View
...
+tokenAddressView MetadataSettings{} = View
+  { vName = "token_address"
+  , vDescription = Just
+      "Returns the address of the associated FA2 contract."
+  , vPure = Just True
+  , vImplementations =
+      [ VIMichelsonStorageView $
+          mkMichelsonStorageView @Storage Nothing [] $
+            unsafeCompileViewCode $ WithoutParam $ do
+              drop; selfAddress
```
pasqu4le, line 220: Ah, at the time this was removed `SELF` was indeed the address to the only meaningful token for baseDAO.

However since #170 we have now an external governance token, so I think we should change this view to return that (see the `storage.governance_token.address` field).

Perhaps it makes sense to rename the view as well? To `governance_token_address`?

rinn7e: I see, I didn't check this properly. I have updated this now. Let me know if you want me to rename the commit as well.

sras: Maybe the view can fetch the governance token id as well? @pasqu4le ?

rinn7e: I'm not quite familiar with the new governance token. But it does seem to make more sense to include them both if I'm not mistaken.

pasqu4le: Indeed, I agree that it would probably be more useful, let's return the whole governance token pair, both address and id :ok_hand:

rinn7e: I've pushed an update to this now.

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
```diff
@@ -202,6 +204,20 @@ getTotalSupplyView MetadataSettings{ msConfig = MetadataConfig{} } = View
       ]
   }
 
+governanceTokenView :: DaoView
+governanceTokenView MetadataSettings{} = View
+  { vName = "governance_token"
+  , vDescription = Just
+      "Returns the address of the associated FA2 contract."
```
sras, line 211: Could be slightly amended to indicate it also returns the token-id as well

sras (approved):

haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
```diff
@@ -111,4 +111,9 @@ mkFA2Tests storage mc = testGroup "FA2 off-chain views"
         checkTotalSupply storage frozenTokenId
           @?= Right 200
     ]
+  , testGroup "token_address" $
+    [ testCase "Get the address of the associated FA2 contract" $
```
pasqu4le, line 115: Same here:
```suggestion
    [ testCase "Get the address and token_id of the associated FA2 contract" $
```

pasqu4le (approved):
Same comment as Sandeep, but otherwise LGTM :ok_hand:

PR #202, opened 2021-04-28
[#184] Typescript checks in CI 

### LINKED ISSUE #184: Improve Typescript checks and utiities

In #154 the typescript API has been introduced, however there's two things that could be improved.

First: while CI checks that the API builds correctly, it does not check that the API has been re-generated to match code changes (when there are any).
We should verify that, I suggest doing so as in the `check cabal files` step, so by running the re-generation and check for no diff.
Additionally: the TS build step runs on scheduled builds as well, which is likely unnecessary, so we should reconsider that.

Second point: there's two small improvements that we could use in the top-level `Makefile`:

1. we should provide the env vars to `stack build` in the `typescript` target (see e.g. the `test` one)
Otherwise there may be issues when building the Haskell library for the first time.

2. consider removing the typescript dir "clearing" from the `clean` target.
That command was only used for git-ignored results of Ligo compilation, I think it should go back to that as the typescript stuff gets committed. Moreover, there is already a "cleaning" step in the `typescript` target, so this seems unnecessary.

# Acceptance criteria

1. CI checks that the TS API matches the one generated from the code.
2. (if deemed useful) the existing API build step is prevented from running on scheduled builds
3. the default values for the contract and storage(s) env-var are given to the `./Makefile` `typescript` target
4. (if deemed useful) the `clean` target does not remove the generated typescript files

sras:
## Description

Adds checks to ensure that the typescript files have been regenerated to match any change to contract interface.

## Related issue(s)

Resolves #184 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -115,7 +115,17 @@ steps:
   #     # from building the tests twice and 'ligo-test' running time is mostly that.
   #   commands: *ligo-nettest
 
+  - label: check typescript api
+    if: *not_scheduled
+    depends_on:
+      - build-ligo
+      - build-haskell
+    commands:
+      - nix run -f ci.nix packages.baseDAO-ligo-meta.exes.baseDAO-ligo-meta pkgs.diffutils -c ./scripts/ci/validate-typescript-api.sh
+
+
   - label: typescript-build
+    if: *not_scheduled
     depends_on:
       - build-ligo
       - build-haskell
--- a/Makefile
+++ b/Makefile
@@ -247,4 +247,3 @@ typescript: all
 
 clean:
 	rm -rf $(OUT)
-	rm -rf $(TS_OUT)/baseDAO/src/generated/*
--- a/scripts/ci/validate-typescript-api.sh
+++ b/scripts/ci/validate-typescript-api.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2021 Tocqueville Group
+#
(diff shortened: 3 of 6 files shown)


.buildkite/pipeline.yml
```diff
@@ -102,7 +102,16 @@ steps:
   #     ./result/bin/baseDAO-test --nettest-run-network
   #       --pattern '\$1 == "On network" || \$NF == "On network" || \$0 ~ /.On network./'
 
+  - label: check typescript api
+    depends_on:
```
pasqu4le, line 106: Perhaps this too should run on master only after a PR is merged and not in a scheduled run?
```suggestion
    if: *not_scheduled
    depends_on:
```

.buildkite/pipeline.yml
```diff
@@ -102,7 +102,16 @@ steps:
...
+      - build-ligo
+      - build-haskell
+    commands:
+      - nix-build ci.nix -A packages.baseDAO-ligo-meta.tests.baseDAO-test
```
pasqu4le, line 110: AFAICT this step is unnecessary, as the `baseDAO-ligo-meta` package is already built by `build-haskell` and tests aren't necessary to check the typescript diff.

pasqu4le (commented):
Two small comments, but it looks like the API is already outdated, so we need to fix that here as well.

.buildkite/pipeline.yml
sras: Turned out to be true. Have removed it.

pasqu4le (approved):

PR #204, opened 2021-04-29
[#198] Add more assets to the auto prerelease step

### LINKED ISSUE #198: Add more assets to the auto prerelease step

In CI we have an auto prerelease step that creates a prerelease based on master.
Currently this only includes the `baseDAO.tz` contract itself.

However, we can modify it to:
- compile all storage(s) with the default values
- generate the known metadata (for the ligo version only)

# Acceptance criteria

Both the metadata and all storage(s) with the default values are included in the automatic pre-relese by CI

rinn7e:
## Description

Problem: In CI we have an auto prerelease step that creates a
prerelease based on master. Currently this only includes the
`baseDAO.tz` contract itself.

However, we can modify it to:
- compile all storage(s) with the default values
- generate the known metadata (for the ligo version only)

Solution: Add all trivialDAO, treasuryDAO, and registryDAO storages,
and add `known_metadata.json` to the pre-release step as assets.

## Related issue(s)

Resolves #198

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -45,8 +45,14 @@ steps:
     commands:
     - nix-build ci.nix -A build-ligo -o ./ligo-out/
     - nix run -f ci.nix morley -c morley optimize --contract ./ligo-out/baseDAO.tz --output baseDAO.tz
+    - cp ./ligo-out/trivialDAO_storage.tz trivialDAO_storage.tz
+    - cp ./ligo-out/registryDAO_storage.tz registryDAO_storage.tz
+    - cp ./ligo-out/treasuryDAO_storage.tz treasuryDAO_storage.tz
     artifact_paths:
       - baseDAO.tz
+      - trivialDAO_storage.tz
+      - registryDAO_storage.tz
+      - treasuryDAO_storage.tz
 
   - label: build-haskell
     key: build-haskell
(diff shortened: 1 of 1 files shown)


rinn7e:
@pasqu4le This PR is still WIP, but I would some clarification:

> However, we can modify it to:
> 1. compile all storage(s) with the default values
> 2. generate the known metadata (for the ligo version only)

Point 1 is done, but I'm not sure about point 2. Could you help explaining what the known metadata mean in this case.

pasqu4le:
@rinn7e oh yes: see the `baseDAO-ligo-meta` package, aka the Haskell package for ligo in `haskell`.

The executable has the option to generate Metadata if provided with a `MetadataConfig`, I see the "known metadata" because this OFC cannot include more information that a DAO owner might want to add.

As you can see we also have a `defaultMetadataConfig`, so we can use that same info to generate and "example" metadata to include in releases.

rinn7e:
> @rinn7e oh yes: see the `baseDAO-ligo-meta` package, aka the Haskell package for ligo in `haskell`.
> 
> The executable has the option to generate Metadata if provided with a `MetadataConfig`, I see the "known metadata" because this OFC cannot include more information that a DAO owner might want to add.
> 
> As you can see we also have a `defaultMetadataConfig`, so we can use that same info to generate and "example" metadata to include in releases.

This PR is ready for review now, I included all the assets now. For the known metadata, I decide to get this from the executable via `print-metadata. 

All the assets can be view here: https://github.com/tqtezos/baseDAO/releases/tag/updating-auto-release

Let me know if you think this is good enough.

pasqu4le (approved):
LGTM :ok_hand: 

Don't forget the TODO to revert!

PR #216, opened 2021-05-03
[#213] Remove set_quorum_threshold and its limits from config

### LINKED ISSUE #213: Remove set_quorum_threshold

The motivation for this issue is specular to #212 but for the quorum threshold.

Here too we also won't need the limits anymore and we can remove them from the contract `storage`.

# Acceptance criteria

The `set_quorum_threshold` entrypoint is no longer part of the contract.

sras:
## Description

Remove the 'set_quorum_threshold' entrypoint and the associated code and the
max_quorum_threshold, min_quorum_threshold limits that are part of the
configuration in storage.

## Related issue(s)

Resolves #213 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -102,10 +102,6 @@ type config =
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   ; max_votes : nat
   // ^ Determine the maximum number of votes associated with a proposal including positive votes
-  ; max_quorum_threshold : quorum_threshold
-  // ^ Determine the maximum value of quorum threshold that is allowed to be set.
-  ; min_quorum_threshold : quorum_threshold
-  // ^ Determine the minimum value of quorum threshold that is allowed to be set.
   ; max_voting_period : nat
   // ^ Determine the maximum value of voting period that is allowed to be set.
   ; min_voting_period : nat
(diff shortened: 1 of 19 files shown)


docs/specification.md
```diff
@@ -689,33 +684,6 @@ Parameter (in Michelson):
...
-```
-(pair %set_quorum_threshold (nat %numerator) (nat %denominator))
-```
-
-- Update the quorum threshold which proposals have to meet to not get rejected.
-- The Quorum threshold is calculated as the proportion of total votes (upvotes
-  and downvotes) over the frozen token's total supply.
-- This affects all ongoing and new proposals.
-- Fails with `NOT_ADMIN` if the sender is not the administrator.
-- Fails with `OUT_OF_BOUND_QUORUM_THRESHOLD` if the quorum threshold value is out of the bound set by the configuration or the numerator is not smaller than the denominator.
```
pasqu4le, line 717: Not that this entrypoint has been removed the `OUT_OF_BOUND_QUORUM_THRESHOLD` is no longer used.

We should remove it from the `Errors` section in this doc as well as removing the instances in `Ligo.BaseDAO.Types`

src/proposal.mligo
```diff
@@ -285,17 +285,6 @@ let set_voting_period(new_period, config, store : voting_period * config * stora
   let store = { store with voting_period = new_period; last_period_change = vp_log } in
   (([] : operation list), store)
 
-// Update quorum_threshold. The new quorum_threshold affects
-// all ongoing and new proposals.
-[@inline]
-let set_quorum_threshold(new_threshold, config, store : quorum_threshold * config * storage): return =
-  let store = authorize_admin store in
-  if   is_le_qt (new_threshold, config.max_quorum_threshold)
```
pasqu4le, line 293: The definition of `is_le_qt` can now also be removed, as this was its only use.

pasqu4le (commented):
Only two small points.
Ah and the commits need to be fixed as well.

docs/specification.md
sras: Updated

src/proposal.mligo
sras: Updated

pasqu4le (approved):

PR #218, opened 2021-05-03
[#209] Remove the `burn` entrypoint

### LINKED ISSUE #209: Remove the `burn` entrypoint

The motivation for this issues are specular to the ones for `mint`, see #208 

Tl;dr: `admin` should not be able to use `burn` at any time and it's use can be replaced by using the right ledger at origination.

# Acceptance criteria

The `burn` entrypoint` is no longer part of the contract.

rinn7e:
## Description

Problem: `burn` is an admin-only entrypoint that allow
destroying tokens. This was intended to
be used mainly to set up the contract and then to not use them again.
However, it is possible to setup the ledger with correct initial value
at origination and we no longer need these entrypoints.

Solution: Remove `burn` entrypoint and remove all related
tests.

## Related issue(s)

Resolves #209

- Tests
        silently reappearing again.

- Documentation
    - Haddock


pasqu4le (approved):
One minor nitpick: in the commit message it says:
```
be used mainly to set up the contract and then to not use them again.
However, it is possible to setup the ledger with correct initial value
at origination and we no longer need these entrypoints.
```
however since we are speaking of a single entrypoint, it should say:
```
be used mainly to set up the contract and then to not be used again.
However, it is possible to setup the ledger with correct initial value
at origination and we no longer need this entrypoint.
```

I realized now that I made these mistakes in the issue description, so I'll fix them there.

PR #219, opened 2021-05-03
[#205] Update voting period in the Makefile from 11 seconds to 11 days

### LINKED ISSUE #205: Change the suggested voting period in the Makefile from 11 seconds to days

In the `Makefile` we have a default voting period of `11n`, since this value is in seconds, this is too low for a realistic use.

We should modify it to last a few days, probably 11 days would be a decent value.

Additionally, we should better document that "seconds" is the unit used here.

# Acceptance criteria

The default voting period length lasts at least a few days, the spec document clarifies that `voting_period` is a value expressed in seconds.

rinn7e:
## Description

Problem: In the `Makefile` we have a default voting period of `11n`,
since this value is in seconds, this is too low for a realistic use.

Solution: Update voting period in Makefile to 950400 seconds (11 days).

## Related issue(s)

Resolves #205 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ $(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
 $(OUT)/trivialDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
-$(OUT)/trivialDAO_storage.tz : voting_period = 11n
+$(OUT)/trivialDAO_storage.tz : voting_period = 950400n # 11 days
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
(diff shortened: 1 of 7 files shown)


pasqu4le (commented):
The `docs/building.md` needs to be updated as well, as it says e.g.:
> All its arguments are optional and will be equal to the values above if not
specified.

But no list has been updated.

The rest LGTM

pasqu4le:
Ah @rinn7e one last thing, in the description I mentioned that we should make sure that:
> he spec document clarifies that voting_period is a value expressed in seconds.

ATM the only place that says the value is in seconds is the Haskell code, I think we should add a comment in `types.mligo` too and mention at least once _somewhere_ in the `specification.md` the same thing.

rinn7e:
> Ah @rinn7e one last thing, in the description I mentioned that we should make sure that:
> 
> > he spec document clarifies that voting_period is a value expressed in seconds.
> 
> ATM the only place that says the value is in seconds is the Haskell code, I think we should add a comment in `types.mligo` too and mention at least once _somewhere_ in the `specification.md` the same thing.

I add a type alias `second` and used it everywhere in the ligo files. Let me know if this is good enough.

pasqu4le (approved):
@rinn7e I think it works well, my only suggestion is to change it to `seconds`, plural.

Issue #208, opened 2021-05-02
Remove the `mint` entrypoint


pasqu4le:
# Clarification and motivation

`mint` is an admin-only entrypoint that allows him to generate more tokens.

This was intended to be used mainly to set up the contract and then to not be used again, however ATM:
- nothing prohibits the `admin` from using it later on as well
- it's possible to set up the ledger with the correct initial value at origination

As such, we should remove the entrypoint entirely.

# Acceptance criteria

The `mint` entrypoint is no longer part of the contract

Issue #203, opened 2021-04-29
Include token_id in the sender check during transfer calls


sras:
# Clarification and motivation

In #54 we added support for multiple token_ids other than the frozen token. This means that as per FA2, operators should be tracked for each token_id. But as of now, we do not include this information in the `operators` structure. We should include token_id in the `operators` and use it to check that the operator is allowed to transfer token_ids in the request.

# Acceptance criteria

1. `operators` include `token_id` along with operator address, indicating the token_ids a certain operator is allowed to transfer for the ower.
2. During transfers, we should check that the sender is a valid operator by including token_id in the check.

PR #220, opened 2021-05-03
[#210] Remove the special `admin` permissions on `transfer`

### LINKED ISSUE #210: Remove the special `admin` permissions on `transfer`

The `transfer` entrypoint has a special permission that allows the `admin` (and only him) to transfer frozen tokens on the ledger.

This was intended as a shorthand for `mint` and `burn`, which granted him the same ability.
However, since these are to be removed (see #208 and #209) we should remove this permission as well.

Note: it's probably better to implement this after the two previous issues.

# Acceptance criteria

There is no special treatment for the `admin` in the `transfer` entrypoint logic.

rinn7e:
## Description

Problem: The `transfer` entrypoint has a special permission that
allows the `admin` (and only him) to transfer frozen tokens on the ledger.
This was intended as a shorthand for `mint` and `burn`, which granted him
the same ability. owever, since these have been removed (see #208 and #209) we should
remove this permission as well.

Solution: Remove the special permission which allow `admin` to transfer
frozen token.

## Related issue(s)

Resolves #210 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -170,7 +170,6 @@ whole contract (hence "global"):
 * **administrator**
   - Can re-assign this role to a new address.
   - Can perform administrative operations.
-  - Can transfer FA2 tokens owned or operated by this contract.
   - There always must be exactly one administrator.
 
 Additionally, the contract inherits the **operator** role from FA2.
@@ -247,7 +246,7 @@ The list of errors may be inaccurate and incomplete, it will be updated during t
 | `NOT_PENDING_ADMIN`             | Authorized sender is not the current pending administrator                                                  |
 | `NOT_TOKEN_OWNER`               | Trying to configure operators for a different wallet which sender does not own                              |
 | `FAIL_PROPOSAL_CHECK`           | Throws when trying to propose a proposal that does not pass `proposalCheck`                                 |
-| `FROZEN_TOKEN_NOT_TRANSFERABLE` | Transfer entrypoint is called for frozen token by a non-admin sender                                        |
+| `FROZEN_TOKEN_NOT_TRANSFERABLE` | Transfer entrypoint is called for frozen token                                                              |
 | `PROPOSAL_NOT_EXIST`            | Throws when trying to vote on a proposal that does not exist                                                |
 | `QUORUM_NOT_MET`                | A proposal is flushed, but there are not enough votes                                                       |
 | `VOTING_PERIOD_OVER`            | Throws when trying to vote on a proposal that is already ended                                              |
@@ -353,7 +352,7 @@ Parameter (in Michelson):
 
 - Permission logic follows the default permissions descriptor specified in FA2.
 
-- The administrator can transfer frozen tokens from any address to any address.
+- Fails with `FROZEN_TOKEN_NOT_TRANSFERABLE` when trying to transfer frozen tokens.
 
 ### **balance_of**
 
--- a/haskell/test/Test/Ligo/BaseDAO/Token/Transfer.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Token/Transfer.hs
@@ -133,11 +133,11 @@ transferTests = testGroup "Transfer:"
             transfer 10 unfrozenTokens owner2 owner1 dao
               & expectCustomError_ #fA2_NOT_OPERATOR dao
 
-      , nettestScenarioCaps "admin can transfer frozen tokens from any address to any address" $ do
+      , nettestScenarioCaps "admin cannot transfer frozen tokens from any address to any address" $ do
           ((owner1, _), (owner2, _), dao, _, admin) <- originateWithCustomToken
 
           withSender (AddressResolved admin) $
             transfer 10 frozenTokens owner1 owner2 dao
-          assertBalanceOf owner2 110 frozenTokens dao
+              & expectCustomErrorNoArg #fROZEN_TOKEN_NOT_TRANSFERABLE dao
       ]
   ]
--- a/src/token/fa2.mligo
+++ b/src/token/fa2.mligo
@@ -19,15 +19,13 @@ let validate_token_type (token_id, store : token_id * storage): token_id =
 
 [@inline]
 let check_sender (from_, token_id, store : address * nat * storage): address =
-  if (Tezos.sender = from_)
-  then from_
-  else
-    let key: operator = { owner = from_; operator = sender; token_id = token_id } in
-    if Big_map.mem key store.operators
-    then from_
-    else
-     ([%Michelson ({| { FAILWITH } |} : string * unit -> address)]
-        ("FA2_NOT_OPERATOR", ()) : address)
+  let key: operator = { owner = from_; operator = sender; token_id = token_id } in
+  if token_id = store.frozen_token_id then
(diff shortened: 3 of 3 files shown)


rinn7e:
@pasqu4le This is probably related to the issues you work on #203. So I'm not sure which one should be rebased on which. Nevertheless, this is ready for review.

docs/specification.md
```diff
@@ -359,8 +357,6 @@ Parameter (in Michelson):
 
 - Permission logic follows the default permissions descriptor specified in FA2.
 
-- The administrator can transfer frozen tokens from any address to any address.
```
pasqu4le, line 362: The logic change isn't entirely correct, perhaps my issue description is a bit ambiguous.

Currently nobody **except** the admin can transfer frozen tokens, what we should do is not allow anyone to transfer them, but rather remove the admin special permission so that nobody **including** the admin can make such transfer.

To put this in context, this issue basically reverts the change made long ago for #26 , so frozen tokens (that at the time had the id of `1`) should not be transferable at all once again.

docs/specification.md
```diff
@@ -251,7 +250,6 @@ The list of errors may be inaccurate and incomplete, it will be updated during t
 | `NOT_PENDING_ADMIN`             | Authorized sender is not the current pending administrator                                                  |
 | `NOT_TOKEN_OWNER`               | Trying to configure operators for a different wallet which sender does not own                              |
 | `FAIL_PROPOSAL_CHECK`           | Throws when trying to propose a proposal that does not pass `proposalCheck`                                 |
-| `FROZEN_TOKEN_NOT_TRANSFERABLE` | Transfer entrypoint is called for frozen token by a non-admin sender                                        |
```
pasqu4le, line 254: Following the previous comment, this error still needs to exist, the difference is that the `by a non-admin sender` needs to be removed, as it now applies to that role as well.

haskell/test/Test/Ligo/BaseDAO/Token/Transfer.hs
```diff
@@ -110,34 +103,11 @@ transferTests = testGroup "Transfer:"
             transfer 1 unfrozenTokens nonexistent owner1 dao
               & expectCustomError #fA2_INSUFFICIENT_BALANCE dao (#required .! 1, #present .! 0)
 
-      , nettestScenarioCaps "cannot transfer frozen tokens" $ do
-          ((owner1, _), (owner2, _), dao, _, _) <- originateWithCustomToken
-
-          withSender (AddressResolved owner1) $
-            transfer 10 frozenTokens owner1 owner2 dao
-              & expectCustomErrorNoArg #fROZEN_TOKEN_NOT_TRANSFERABLE dao
```
pasqu4le, line 118: I think both this and the one removed above should be kept given the other comments.

haskell/test/Test/Ligo/BaseDAO/Token/Transfer.hs
```diff
@@ -110,34 +103,11 @@ transferTests = testGroup "Transfer:"
...
-            transfer 10 unfrozenTokens owner2 owner1 dao
-              & expectCustomError_ #fA2_NOT_OPERATOR dao
-
-      , nettestScenarioCaps "admin can transfer frozen tokens from any address to any address" $ do
-          ((owner1, _), (owner2, _), dao, _, admin) <- originateWithCustomToken
-
-          withSender (AddressResolved admin) $
-            transfer 10 frozenTokens owner1 owner2 dao
-          assertBalanceOf owner2 110 frozenTokens dao
-      ]
```
pasqu4le, line 142: I would keep both of these, except the second one needs to be modified to check that not even the `admin` can transfer frozen tokens from other addresses and his call fails with `FROZEN_TOKEN_NOT_TRANSFERABLE` too.

src/token/fa2.mligo
```diff
@@ -80,14 +80,7 @@ let credit_to (amt, to_, token_id, ledger, total_supply : nat * address * nat *
...
   let transfer_one (store, tx : storage * transfer_destination): storage =
-    let valid_from_ =
-      if tx.token_id = store.frozen_token_id
-      then
-        if Tezos.sender = store.admin
-        then ti.from_
-        else (failwith("FROZEN_TOKEN_NOT_TRANSFERABLE") : address)
-      else check_sender (ti.from_, tx.token_id, store)
-      in
+    let valid_from_ = check_sender (ti.from_, tx.token_id, store) in
```
pasqu4le, line 83: As said above we should still keep the check, however since it is now simpler I suggest moving the `if tx.token_id = store.frozen_token_id then (failwith("FROZEN_TOKEN_NOT_TRANSFERABLE") : address)` inside `check_sender` itself.

pasqu4le (changes requested):
@rinn7e the change in logic is incorrect, see the comments below.

Quite a bit of changes are needed, so I won't base my PR on top of this (yet).
I don't think there will be many merge conflict anyway, so hopefully it won't be problem to merge them in either order.

docs/specification.md
rinn7e: I see, I definitely misunderstood this. Now that I think about it, it really does not make sense to allow frozen token transfer by everyone. I'll update this in a moment.

rinn7e: I've changed the logic and update the spec. Should be the correct logic now.

src/token/fa2.mligo
rinn7e: I've updated this as well.

rinn7e:
I've squashed the commits to better resolve some conflict from rebasing.

src/token/fa2.mligo
```diff
@@ -19,15 +19,18 @@ let validate_token_type (token_id, store : token_id * storage): token_id =
...
     then from_
     else
-     ([%Michelson ({| { FAILWITH } |} : string * unit -> address)]
-        ("FA2_NOT_OPERATOR", ()) : address)
+      let key: operator = { owner = from_; operator = sender; token_id = token_id } in
+      if Big_map.mem key store.operators
+      then from_
+      else
+        ([%Michelson ({| { FAILWITH } |} : string * unit -> address)]
+            ("FA2_NOT_OPERATOR", ()) : address)
```
pasqu4le, line 33: The logic seems correct to me, but I think this will be a little more readable (and maybe even more efficient) if implemented as:
```ocaml
let check_sender (from_, token_id, store : address * nat * storage): address =
  if token_id = store.frozen_token_id
  then (failwith("FROZEN_TOKEN_NOT_TRANSFERABLE") : address)
  else if (Tezos.sender <> from_)
    let key: operator = { owner = from_; operator = sender; token_id = token_id } in
    if Big_map.mem key store.operators
    then from_
    else ([%Michelson ({| { FAILWITH } |} : string * unit -> address)] 
            ("FA2_NOT_OPERATOR", ()) : address)
  else from_
```

pasqu4le (approved):
One more minor nitpick: in the commit message it says:
```
However, since these are to be removed (see #208 and #209) we should
remove this permission as well.
```
which is a bit outdated as both those have been closed already, I suggest replacing it with:
```
However, since these have been removed (see #208 and #209) we should
remove this permission as well.
```

Aside from that and the suggestion below all LGTM

PR #221, opened 2021-05-03
[#203] single token operator

### LINKED ISSUE #203: Include token_id in the sender check during transfer calls

pasqu4le:
## Description

As per FA2, operators should be tracked for each `token_id`

This PR adds it to the stored operators and fixed the check accordingly.

## Related issue(s)

Resolves #203

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
+++ b/haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
@@ -140,7 +140,6 @@ isOperatorView MetadataSettings{} = View
             unsafeCompileViewCode $ WithParam @FA2.OperatorParam $ do
               dip dup
               convertOperatorParam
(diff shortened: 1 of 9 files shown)


rinn7e (approved):
Seems good for me.

sras (approved):
LGTM

PR #222, opened 2021-05-04
[#212] Remove set voting period

### LINKED ISSUE #212: Remove `set_voting_period`

The voting period length can be set at origination and we no longer want the `admin` to be able to change it after that.

Additionally, this makes the voting period bounds, `min` and `max` limits, unnecessary and we can remove them from the contract `storage`.

# Acceptance criteria

The `set_voting_period` entrypoint is no longer part of the contract.

sras:
## Description

Problem: We have to remove 'set_voting_period' entrypoint from the
contract interface.

Solution: Remove the entrypoint, and update storage to remove the
'min_voting_period' and 'max_voting_period' values from configuration.
Also moves 'voting_period' configuration to the config section of
storage. The Makefile is also updated to remove reference to
voting_period limits. Also updates documentation.

## Related issue(s)

Resolves #212 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


src/types.mligo
```diff
@@ -104,6 +104,7 @@ type voter =
   ; vote_type : vote_type
   }
 
+type voting_period = nat
```
rinn7e, line 107: We probably on longer need this. I've removed this in #219, and uses `seconds` instead which gives better clarification.

rinn7e (changes requested):

src/types.mligo
sras: I see. I thought it got removed by mistake.

But isn't 'voting_period' more specific than 'seconds', thus carry more information than 'seconds'?

rinn7e: How I think about this is that, `voting_period` is measured in second. 

It was not obvious previously. The type `nat` or `type voting_period = nat` does not tell that this `nat` stands for second. It could be confused for `hour` or `millisecond` and so on.

sras: Ah, I see. Then it could be resolved by aliasing `voting_period` to be `seconds`.

rinn7e: Either ways, it's fine for me.

rinn7e (approved):
Seems fine for me. I've approved but maybe @pasqu4le want to check this as well.

Makefile
```diff
@@ -89,13 +89,11 @@ $(OUT)/trivialDAO_storage.tz: src/**
...
-          } \
-        ; config_data = \
-          { max_period = 2592000n \
-          ; min_period = 1n \
-          ; quorum_threshold = $(quorum_threshold) \
           } \
+          ; config_data = \
+            { voting_period = $(voting_period) \
+            ; quorum_threshold = $(quorum_threshold) \
+            } \
```
pasqu4le, line 96: I see you have increased the indentation, however `storage_data` and `config_data` are indeed part of the same record.

I think the correct indentation is the one that was used before (same for the other changes in this file).

src/proposal.mligo
```diff
@@ -26,15 +26,15 @@ let check_if_proposal_exist (proposal_key, store : proposal_key * storage): prop
 
 // Gets the current period counting how many `voting_period` s have passed since
 // the 'started_on` timestamp. The periods start from zero index.
-let get_current_period_num(last_period_change, vp_length : last_period_change * nat) : nat =
+let get_current_period_num(last_period_change, vp_length : last_period_change * voting_period) : nat =
   match is_nat((Tezos.now - last_period_change.changed_on) : int) with
         | Some (elapsed_time) -> last_period_change.period_num + elapsed_time/vp_length
         | None -> ([%Michelson ({| { FAILWITH } |} : string * unit -> nat)]
             ("STARTED_ON_IN_FUTURE", ()))
```
pasqu4le, line 33: IIRC the need for the `last_period_change` was only to be able to track changes made using `set_voting_period`.

Now that we no longer have that entrypoint this can be simplified, we should no longer need to store the `period_num` of a `last_period_change` and so we can remove the latter entirely.
In the storage we can store a:
```ocaml
  ; start_time : timestamp
```
and here we can simplify the calculation as well, the period number should be the result of `elapsed_time/vp_length`

haskell/test/Test/Ligo/TreasuryDAO.hs
```diff
@@ -100,8 +103,8 @@ flushTokenTransfer = uncapsNettest $ withFrozenCallStack $ do
   withSender (AddressResolved owner2) $
     call dao (Call @"Freeze") (#amount .! 20)
 
-  -- Advance one voting period (which is 1).
-  advanceTime (sec 1.5)
+  -- Advance one voting period (which is 10).
+  advanceTime (sec 20)
```
pasqu4le, line 107: This seems incorrect (at least the comment), I suppose the idea here was to advance a little more than a period (but not two).

pasqu4le (changes requested):
One more thing: in `docs/building.md` at the end it says:
```
The same is true for `voting_period`.
```
which should now be removed as no check is done on that.

src/proposal.mligo
sras: Ah, yes. yesterday I reminded myself it should be removed, but missed it today. Will add now.

src/types.mligo
pasqu4le: I think having both may be confusing and possibly not that helpful, ideally we should have a `newtype` like in Haskell.

AFAIK ligo doesn't really have that, but it is possible to use records with a single field, I suggest:
```ocaml
type voting_period = { length : seconds }
```

This makes the usage a bit more cumbersome (you need to do `voting_period.length` all the time), but it makes it easier to see when we are dealing with just seconds/length of time VS voting periods.
Note: AFAIK this is compiled without additional costs.

haskell/test/Test/Ligo/TreasuryDAO.hs
sras: >a little more than a period (but not two).

But since periods start with proposing, if only little more then a period was advanced, it would end up in a voting period, right? In that case we would not be able to raise the proposal, which appear to be the next step in this test.

I have updated it with the comment amended.

src/proposal.mligo
sras: Have replaced `last_period_change` from storage with only a timestamp and updated the calculations.

Makefile
sras: Have fixed it.

src/proposal.mligo
```diff
@@ -26,15 +26,15 @@ let check_if_proposal_exist (proposal_key, store : proposal_key * storage): prop
...
 // Gets the current period counting how many `voting_period` s have passed since
 // the 'started_on` timestamp. The periods start from zero index.
-let get_current_period_num(last_period_change, vp_length : last_period_change * voting_period) : nat =
-  match is_nat((Tezos.now - last_period_change.changed_on) : int) with
-        | Some (elapsed_time) -> last_period_change.period_num + elapsed_time/vp_length
+let get_current_period_num(start_time, vp_length : timestamp * voting_period) : nat =
+  match is_nat((Tezos.now - start_time) : int) with
+        | Some (elapsed_time) -> elapsed_time/vp_length
         | None -> ([%Michelson ({| { FAILWITH } |} : string * unit -> nat)]
             ("STARTED_ON_IN_FUTURE", ()))
```
pasqu4le, line 33:
```suggestion
  | Some (elapsed_time) -> elapsed_time/vp_length
  | None -> ([%Michelson ({| { FAILWITH } |} : string * unit -> nat)]
      ("STARTED_ON_IN_FUTURE", ()))
```

src/types.mligo
pasqu4le: @sras pinging in case you missed the comment above

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -225,12 +210,11 @@ test_BaseDAO_Proposal =
 
     , nettestScenarioOnEmulator "a proposer is returned a fee after the proposal succeeds" $
         \_emulated -> uncapsNettest $ do
-          ((proposer, _), (voter, _), dao, _, admin) <- originateLigoDao
+          ((proposer, _), (voter, _), dao, _, admin) <- originateLigoDaoWithConfigDesc dynRecUnsafe (ConfigDesc $ VotingPeriod 60)
 
           -- Use 60s for voting period, since in real network by the time we call
           -- the vote entrypoint 30s have already passed.
```
pasqu4le, line 216: It seems no longer relevant
```suggestion
```

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -305,12 +289,12 @@ burnsFeeOnFailure
   :: forall caps base m. (MonadNettest caps base m)
   => FailureReason -> m ()
 burnsFeeOnFailure reason = do
-  ((proposer, _), (voter, _), dao, _, admin) <- originateLigoDao
+  ((proposer, _), (voter, _), dao, _, admin) <-
+      originateLigoDaoWithConfigDesc dynRecUnsafe (ConfigDesc $ VotingPeriod 60)
 
   -- Use 60s for voting period, since in real network by the time we call
   -- the vote entrypoint 30s have already passed.
```
pasqu4le, line 296: Same here
```suggestion
```

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -529,13 +487,11 @@ flushNotAffectOngoingProposals
   :: (MonadNettest caps base m, HasCallStack)
   => (ConfigDesc Config -> OriginateFn m) -> m ()
 flushNotAffectOngoingProposals originateFn = do
-  ((owner1, _), _, dao, _, admin) <- originateFn testConfig
-
-
   -- Note: Cannot set to few seconds, since in real network, each
   -- calls takes some times to run. 20 seconds seem to be the ideal.
```
pasqu4le, line 491: Likewise
```suggestion
```
BTW we kinda ignored this here (with the new at-origination value), but let's get to that when we'll need to.

src/types.mligo
sras: @pasqu4le Yes, makes sense, since it provides a little more safety than using just an alias. Have updated.

Makefile
sras: I see. Infact I checked the other two instances originally, but somehow thought they looked alright.

I see now that the inner braces of config record does not alight with the field names above them. I believe this is what needs to be fixed and I have made that change. 

>P.s. also the review comment just above

Hope this refers to amending `docs/building.md`, in which case I have included it.

pasqu4le (approved):
LGTM now

docs/specification.md
```diff
@@ -102,10 +102,6 @@ type config =
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   ; max_votes : nat
   // ^ Determine the maximum number of votes associated with a proposal including positive votes
-  ; max_voting_period : nat
-  // ^ Determine the maximum value of voting period that is allowed to be set.
-  ; min_voting_period : nat
-  // ^ Determine the minimum value of voting period that is allowed to be set.
```
pasqu4le, line 105: Oh actually, one last thing: this needs to be updated, as `config` now contains the `voting_period` field as well.

pasqu4le (changes requested):

docs/specification.md
sras: Have added `voting_period` and `quorum_threshold` as well, which seemed to be missing as well.

pasqu4le: Ah great :ok_hand:

pasqu4le (approved):

PR #223, opened 2021-05-04
[#211] Remove set_fixed_fee entrypoint

### LINKED ISSUE #206: [BUG] fixed_proposal_fee_in_token option in the Makefile is not used in the storage
# Description

We have a `fixed_proposal_fee_in_token` option in the Makefile to set the value of the storage field with the same name in the storage.

However the option isn't correctly used and instead we set the value to `0n` [by default](https://github.com/tqtezos/baseDAO/blob/master/src/defaults.mligo#L71).

Probably this is a rebase artifact resulting from a recent update.

# To Reproduce

Running `make out/treasuryDAO_storage.tz` and `make fixed_proposal_fee_in_token=xxx out/treasuryDAO_storage.tz` produces the same result, regardless of the `xxx` given.

# Expected behavior

The storage contains the chosen `fixed_proposal_fee_in_token`

### LINKED ISSUE #211: Remove `set_fixed_fee_in_token`

The `fixed_fee_in_token` can be set at origination and we don't want the `admin` to be able to change it after that (which he can currently do).

As such we should remove the `set_fixed_fee_in_token` entrypoint.

# Acceptance criteria

The  `set_fixed_fee_in_token` entrypoint is no longer part of the contract.

sras:
## Description

Removes set_fixed_fee entrypoint and updates assocaited docs and tests.
Resolves #206 as well.

## Related issue(s)

Resolves #211, #206

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -107,6 +107,8 @@ type config =
   // ^ Determine the voting_period length.
   ; quorum_threshold : quorum_threshold
   // ^ Determine the quorum threshold that the proposals need to meet.
+  ; fixed_proposal_fee_in_token : nat
+  // ^ A base fee paid for submitting a new proposal.
 
   ; custom_entrypoints : custom_entrypoints
   // ^ Packed arbitrary lambdas associated to a name for custom execution.
@@ -132,9 +134,6 @@ type proposal =
   ; metadata : proposal_metadata
   ; proposer : address
   ; proposer_frozen_token : nat
-  ; proposer_fixed_fee_in_token : nat
-  // ^ A fee paid for submitting the proposal.
-  // Needed to refund correctly if the proposal is successful.
   ; voters : (address * nat) list
   // ^ List of voter addresses associated with the vote amount
   // Needed for `flush` entrypoint.
@@ -275,7 +274,6 @@ Full list:
 * [`transfer_ownership`](#transfer_ownership)
 * [`accept_ownership`](#accept_ownership)
 * [`propose`](#propose)
-* [`set_fixed_fee_in_token`](#set_fixed_fee_in_token)
 * [`vote`](#vote)
 * [`flush`](#flush)
 * [`drop_proposal`](#drop_proposal)
@@ -578,7 +576,7 @@ Parameter (in Michelson):
 ```
 
 - The proposal is saved under `BLAKE2b` hash of proposal value and sender.
-- The `Natural` value: `proposalTokenAmount` determines how many sender's frozen tokens will be staked in addition to the [fee](#set_fixed_fee_in_token).
+- The `Natural` value: `proposalTokenAmount` determines how many sender's frozen tokens will be staked in addition to the [fee](#configuration)
 - Sender MUST have enough frozen tokens (i. e. `≥ proposalTokenAmount + fee`) that are not already staked for a proposal or a vote.
 - Fails with `NOT_ENOUGH_FROZEN_TOKENS` if the unstaked frozen token balance of the SENDER
   is less than `proposalTokenAmount + fee`.
@@ -587,20 +585,6 @@ Parameter (in Michelson):
 - Fails with `MAX_PROPOSALS_REACHED` if the current amount of ongoing proposals is at max value set by the config.
 - Fails with `PROPOSAL_NOT_UNIQUE` if exactly the same proposal from the same author has been proposed.
 
-### **set_fixed_fee_in_token**
-
-```ocaml
-Set_fixed_fee_in_token of nat
-```
-
-Parameter (in Michelson):
-```
-(nat %set_fixed_fee_in_token)
-```
-
-- Update the fee that the proposers have to pay to submit a proposal (the fee is returned if the proposal is successful and burnt otherwise)
-- This affects only new proposals; the existing proposals store the fee paid for their submission.
-- Fails with `NOT_ADMIN` if the sender is not the administrator.
 
 ### **vote**
 
--- a/src/defaults.mligo
+++ b/src/defaults.mligo
@@ -10,6 +10,7 @@ let default_config (data : initial_config_data) : config = {
   decision_lambda = (fun (proposal, extras : proposal * contract_extra) -> (([] : (operation list)), extras));
 
   quorum_threshold = data.quorum_threshold;
+  fixed_proposal_fee_in_token = 0n;
   voting_period = data.voting_period;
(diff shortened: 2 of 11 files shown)


src/defaults.mligo
```diff
@@ -10,6 +10,7 @@ let default_config (data : initial_config_data) : config = {
   decision_lambda = (fun (proposal, extras : proposal * contract_extra) -> (([] : (operation list)), extras));
 
   quorum_threshold = data.quorum_threshold;
+  fixed_proposal_fee_in_token = 0n;
```
pasqu4le, line 13: @sras this is a bug reported in #206, this should be:
```suggestion
  fixed_proposal_fee_in_token = data.fixed_proposal_fee_in_token;
```
except that `initial_config_data` doesn't have that field.

If @rinn7e hasn't started working on that issue yet can you self-assign it and fix it here by adding a commit just for it?
I think it will avoid a merge conflict and be quicker to proceed this way.

src/proposal.mligo
```diff
@@ -183,7 +183,7 @@ let add_proposal (propose_params, store : propose_params * storage): storage =
     ; metadata = propose_params.proposal_metadata
     ; proposer = Tezos.sender
     ; proposer_frozen_token = propose_params.frozen_token
-    ; proposer_fixed_fee_in_token = store.fixed_proposal_fee_in_token
+    ; proposer_fixed_fee_in_token = config.fixed_proposal_fee_in_token
```
pasqu4le, line 186: This `proposer_fixed_fee_in_token` can now be removed and this operation is no longer necessary.

The reason `proposer_fixed_fee_in_token` exists is that before a proposal was flushed it was possible for the admin to change the fixed fee so it needed to be stored.

Since that's no longer a possibility we can just use the one from the `config` where needed.

docs/specification.md
```diff
@@ -579,7 +578,7 @@ Parameter (in Michelson):
 ```
 
 - The proposal is saved under `BLAKE2b` hash of proposal value and sender.
-- The `Natural` value: `proposalTokenAmount` determines how many sender's frozen tokens will be staked in addition to the [fee](#set_fixed_fee_in_token).
+- The `Natural` value: `proposalTokenAmount` determines how many sender's frozen tokens will be staked in addition to the fee.
```
pasqu4le, line 581: With the removal of this entrypoint and the `proposer_fixed_fee_in_token` (as suggested in a previous comment) there is no longer any documentation about the fixed fee.

Note that the `config` type in the `Configuration` section needs to be updated as it now contains a field for the fee, I suggest moving the comment that is now under `proposer_fixed_fee_in_token` there and to change the link here to point at that section.

pasqu4le (changes requested):

src/defaults.mligo
sras: Have updated

src/proposal.mligo
sras: Removed

docs/specification.md
sras: Have updated.

docs/specification.md
```diff
@@ -106,6 +106,8 @@ type config =
   // ^ Determine the maximum value of voting period that is allowed to be set.
   ; min_voting_period : nat
   // ^ Determine the minimum value of voting period that is allowed to be set.
+  ; fixed_proposal_fee_in_token : nat
+  // ^ A fee paid for submitting the proposal.
```
pasqu4le, line 110:
```suggestion
  // ^ A base fee paid for submitting a new proposal.
```
sounds better, I think.

src/proposal.mligo
```diff
@@ -309,15 +299,15 @@ let unstake_tk(token_amount, addr, store : nat * address * storage): storage =
...
   let (tokens, store) =
     if is_accepted
-    then (proposal.proposer_frozen_token + proposal.proposer_fixed_fee_in_token, store)
+    then (proposal.proposer_frozen_token + fixed_fee, store)
     else
       let slash_amount = rejected_proposal_return (proposal, store.extra) in
-      let fee = proposal.proposer_fixed_fee_in_token in
+      let fee = fixed_fee in
       let frozen_tokens = proposal.proposer_frozen_token + fee in
       let desired_burn_amount = slash_amount + fee in
```
pasqu4le, line 312:
```suggestion
      let frozen_tokens = proposal.proposer_frozen_token + fixed_fee in
      let desired_burn_amount = slash_amount + fixed_fee in
```

pasqu4le (commented):
Minor points, but LGTM.

Also can you add "Resolves #206" in the PR description as well?

It seems it will need a rebase soon too, as it is based on #222

src/proposal.mligo
```diff
@@ -170,9 +170,9 @@ let unfreeze_on_ledger (tokens, addr, ledger, total_supply, frozen_token_id, gov
   let (ledger, total_supply) = debit_from (tokens, addr, frozen_token_id, ledger, total_supply) in
   (operation, ledger, total_supply)
 
-let add_proposal (propose_params, voting_period, store : propose_params * voting_period * storage): storage =
+let add_proposal (propose_params, config, store : propose_params * config * storage): storage =
   let proposal_key = ensure_proposal_is_unique (propose_params, store) in
-  let current_period = get_current_period_num(store.start_time, voting_period) in
+  let current_period = get_current_period_num(store.start_time, config.voting_period) in
```
pasqu4le, line 175: This change seems unnecessary, we can still only use a `voting_period`

src/proposal.mligo
```diff
@@ -196,9 +195,9 @@ let add_proposal (propose_params, voting_period, store : propose_params * voting
 let propose (param, config, store : propose_params * config * storage): return =
   let store = check_is_proposal_valid (config, param, store) in
   let store = check_proposal_limit_reached (config, store) in
-  let amount_to_freeze = param.frozen_token + store.fixed_proposal_fee_in_token in
+  let amount_to_freeze = param.frozen_token + config.fixed_proposal_fee_in_token in
   let store = stake_tk(amount_to_freeze, Tezos.sender, config.voting_period, store) in
-  let store = add_proposal (param, config.voting_period, store) in
+  let store = add_proposal (param, config, store) in
```
pasqu4le, line 200: This too, can be reverted.

pasqu4le (approved):
One minor thing, feel free to squash and merge.

PR #224, opened 2021-05-05
[#214] Remove `get_total_supply` and `get_vote_permit_counter`

### LINKED ISSUE #214: Remove `get_total_supply` and `get_vote_permit_counter`

This is a continuation of #166 

In there we decided to remove these two `view`s and we ended up replacing them with `void` entrypoints.

These are currently needed in order to run tests properly, but are otheewise unnecessary to the contract.
So we want to eventually remove them, but we'll likely need to wait for #31 to do so.

# Acceptance criteria

Both the `get_total_supply` and `get_vote_permit_counter` entrypoints are no longer part of the contract.

rinn7e:
## Description

Problem: After #166, we no longer need `get_total_supply` and `get_vote_permit_counter`
entrypoints. We updated it to void instead of removing because the tests
need them to run properly. Currently, we found a way to check the
storage without using this entrypoints (By using `StorageView` instead)
so we should remove these entrypoints.

Solution: Remove `get_total_supply` and `get_vote_permit_counter`
entrypoints, add `FullStorageView`, update existing storage related type,
add `getTotalSupply` and `getVotePermitsCounter` storage check.

## Related issue(s)

Resolves #214

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
@@ -1,0 +1,76 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+module Test.Ligo.BaseDAO.Common.StorageHelper
+  ( getFullStorage
+  , getFullStorageView
+
+  , GetTotalSupplyFn
+  , getTotalSupplyEmulator
+  -- , getTotalSupplyNetwork
+
+  , GetVotePermitsCounterFn
+  , getVotePermitsCounterEmulator
+  -- , getVotePermitsCounterNetwork
+  ) where
+
+import Lorentz hiding (assert, (>>))
+import Universum
+
+import qualified Data.Map as M
+import Morley.Nettest
+import Morley.Nettest.Pure (PureM)
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+
+import Ligo.BaseDAO.Types
+
+getFullStorage :: AddressOrAlias -> EmulatedT PureM FullStorage
+getFullStorage addr =
+  fromVal @FullStorage <$> getStorage' @(ToT FullStorage) addr
+
+getFullStorageView :: (Monad m) => AddressOrAlias -> NettestT m FullStorageView
+getFullStorageView addr =
+  fromVal @FullStorageView <$> getStorage @(ToT FullStorageView) addr
+
+------------------------------------------------------------------------
+-- GetTotalSupply
+------------------------------------------------------------------------
+
+type GetTotalSupplyFn m = AddressOrAlias -> FA2.TokenId -> m Natural
+
+getTotalSupplyEmulator :: AddressOrAlias -> FA2.TokenId -> EmulatedT PureM Natural
+getTotalSupplyEmulator addr tokenId = do
+  fs <- getFullStorage addr
+  let result = case M.lookup tokenId $ sTotalSupply (fsStorage fs) of
+        Just v -> v
+        Nothing -> error "getTotalSupply: token_id does not exist."
+  pure result
+
+-- | Note: Not needed at the moment, due to all the tests that uses this run only in emulator
+-- anyway. Commented due to weeder.
+
+-- getTotalSupplyNetwork :: (Monad m) => AddressOrAlias -> FA2.TokenId -> NettestT m Natural
+-- getTotalSupplyNetwork addr tokenId = do
+--   fs <- getFullStorageView addr
+--   let result = case M.lookup tokenId $ sTotalSupply (fsStorage fs) of
+--         Just v -> v
+--         Nothing -> error "getTotalSupply: token_id does not exist."
+--   pure result
+
+------------------------------------------------------------------------
+-- GetVotePermitsCounter
+------------------------------------------------------------------------
+
+type GetVotePermitsCounterFn m = AddressOrAlias -> m Nonce
+
+getVotePermitsCounterEmulator :: AddressOrAlias -> EmulatedT PureM Nonce
+getVotePermitsCounterEmulator addr = do
+  fs <- getFullStorage addr
+  pure $ sPermitsCounter (fsStorage fs)
+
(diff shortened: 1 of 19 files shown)


haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -74,24 +74,25 @@ test_BaseDAO_Proposal =
       \_emulated ->
         uncapsNettest $ insufficientTokenVote (originateLigoDaoWithConfigDesc dynRecUnsafe)
 
-  , nettestScenario "cannot propose with insufficient tokens" $
-      uncapsNettest $ insufficientTokenProposal (originateLigoDaoWithConfigDesc dynRecUnsafe)
+  -- Note: When checking storage, we need to split the test into 2 (emulator and network) as demonstrated below:
+  , nettestScenarioOnEmulatorCaps "cannot propose with insufficient tokens (emulator) " $
+      insufficientTokenProposal (originateLigoDaoWithConfigDesc dynRecUnsafe) (\addr -> (length . sProposalKeyListSortByDate . fsStorage) <$> getFullStorage addr)
+  , nettestScenarioOnNetworkCaps "cannot propose with insufficient tokens (network) " $
+      insufficientTokenProposal (originateLigoDaoWithConfigDesc dynRecUnsafe) (\addr -> (length . sProposalKeyListSortByDate . fsStorage) <$> getFullStorageView addr)
```
rinn7e, line 81: @pasqu4le So this took me a while to figure everything out. The changes are quite big, so I'd like to clarify a few things:

- I'm using `StorageView` approach and I follow the pattern according to the stablecoin: https://github.com/tqtezos/stablecoin/blob/9a924a9a5d699d72345439d65fd5c6f12dcd506b/haskell/src/Lorentz/Contracts/Stablecoin/Types.hs#L251.
  - We have quite a more changes, since we use `BigMap` in multiple places, and this pattern requires updating all of them
  
- To do the storage checking, I use this approach from Diogo advice: https://gitlab.com/morley-framework/morley/-/issues/346#note_567052563
  - The `2.1` approach, basically, we split the tests into 2 (emulator and network).
  
So I manage to make this works now by using all the approaches described above, however most of the tests that uses `get_total_supply` and `get_vote_permits_counter` entrypoints are emulator only tests, so we don't actually need `StorageView` to solve this particular issues.

Nevertheless, I think this `StorageView` will be helpful in other tests, and would help solving the issues #31. I add a simple demonstration on how to splits the tests as seen above.

pasqu4le: I see, I agree on the approach, it looks like a good way to handle this :ok_hand: 

One question: how come you didn't go with the second variant instead from Diogo's suggestion:
> Create a function that retrieves a single field from the storage, and works on both the emulator and the network.

It seems to me it would fit your current code very well, but be easier to replace in the future, e.g. with the same same `insufficientTokenProposal` implementation above we could have a single:
```haskell
  , nettestScenarioCaps "cannot propose with insufficient tokens" $
      insufficientTokenProposal (originateLigoDaoWithConfigDesc dynRecUnsafe) getProposalAmount
```
as long as we implement a:
```haskell
getProposalAmount
  -> AddressOrAlias -> m Int
getProposalAmount addr = length <$>
  getStorageField (sProposalKeyListSortByDate . fsStorage) (sProposalKeyListSortByDate . fsStorage)
```
and a
```haskell
getStorageField
 :: MonadNettest caps base m
 => Address
 -> (FullStorage -> a)
 -> (FullStorageView -> a)
 -> m a
getStorageField addr whenEmulator whenNetwork = do
 storageExpr <- getStorageExpr addr
 case fromVal @FullStorage <$> fromExpression storageExpr of
   Right st -> whenEmulator st
   Left err1 ->
     case fromVal @FullStorageView <$> fromExpression storageExpr of
       Right st -> whenNetwork st
       Left err2 ->
         failure $ unlinesF 
           [ "Failed to deserialize storage as 'FullStorage'"
           , indentF 2 $ build err1
           , "Failed to deserialize storage as 'FullStorageView'"
           , indentF 2 $ build err2
           ]
```
(adapted from that suggestion).

pasqu4le: In fact ideally if `getProposalAmount` works on both emulator and network as intended, there is no need for `insufficientTokenProposal` to take it in input, which would be another advantage.

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -0,0 +1,76 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+module Test.Ligo.BaseDAO.Common.StorageHelper
```
pasqu4le, line 3: I currently leave no other comment on this module as I think it would be much improved by the discussion above (if possible).

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
rinn7e: The problem that I run into is that, this functions return a cryptic error message related to `Error: Expect 'True to be equal to 'False`. I think this may be from this particular line: `fromVal @FullStorage <$> fromExpression storageExpr`. 

Basically, maybe it complains about `FullStorage` containing `BigMap`. 

I'm not particularly sure exactly and it starts to take too much time, so I decide to go with the current approach instead. If you have some ideas on how to solve this, let me know.

pasqu4le: Got it, I looked into this and the problem is indeed the one you mentioned; `fromExpression` requires the `UnpackedValScope` here, which `FullStorage` cannot satisfy because of the `BigMap`s in it.

The problem remains with `getStorage` which also requires the same constraint.

I tried to find a solution, including distinguishing the two cases in the `getStorageField` with `Monad.Capabilities.checkCap`, but it seems that doesn't work, it looks like `nettestScenarioCaps` does not actually provide the additional emulated capability to the test,

So it looks like your current approach may be the only one available at the moment, let's go with it.

pasqu4le (approved):
LGTM :ok_hand:

PR #225, opened 2021-05-07
[#207] Dynamic quorum

### LINKED ISSUE #207: Dynamic quorum percentage

As in Tezos' voting process, it is useful to automatically update the quorum percentage based on previous participation.

# Acceptance criteria

Two storage parameters representing percentages:
- `max_change_percent`, with `default = 19%`
- `change_percent`, with `default = 5%`

Each voting stage, a `new_quorum` is calculated from the `old_quorum`
and the previous participation (`number_of_frozen_tokens_last_voting_period / known_total_supply_of_unfrozen_tokens`).

```
possible_new_quorum =
  old_quorum * (1 - change_percent) + participation * change_percent

min_new_quorum =
  old_quorum / (1 + max_change_percent)

max_new_quorum =
  old_quorum * (1 + max_change_percent)

new_quorum =
  max min_new_quorum
  (min max_new_quorum
       possible_new_quorum)
```

Note: the quorum percentage should also respect the `min_quorum_threshold`, `max_quorum_threshold` respectively.

sras:
## Description

Problem: We should implement the dynamic update of the quorum threshold
based on the previous participation.

Solution: Introduce the concept of cycle, which is a proposing, voting
period pair. Then track the number of staked tokens in each cycle, and
use it to calculate the new value of quorum-threshold at the very first
propose call in each cycle. Since proposal can be flushed long after the
cycle in which it was proposed, we also store the quorum threshold for
the proposal cycle in the proposals. We use this stored quorum threshold
while evaluating the votes for this proposal.

Another main change in this PR is that we have stopped storing the
denominator for the fractions. This is to simplify the calculations,
mostly in update quorum procedure, and also to round off fractions to
the required precision.

Also updates documentation and specs.

## Related issue(s)

Resolves #207 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,8 +74,13 @@ $(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
 $(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
-$(OUT)/trivialDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
+$(OUT)/trivialDAO_storage.tz : quorum_threshold = 10n
+$(OUT)/trivialDAO_storage.tz : min_quorum = 1n
+$(OUT)/trivialDAO_storage.tz : max_quorum = 99n
 $(OUT)/trivialDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/trivialDAO_storage.tz : quorum_change = 5n
+$(OUT)/trivialDAO_storage.tz : max_quorum_change = 19n
+$(OUT)/trivialDAO_storage.tz : governance_total_supply = 100n
 $(OUT)/trivialDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
 $(OUT)/trivialDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
 $(OUT)/trivialDAO_storage.tz: src/**
@@ -95,11 +100,16 @@ $(OUT)/trivialDAO_storage.tz: src/**
           ; ledger_lst = $(call escape_double_quote,$(ledger)) \
           } \
         ; config_data = \
-          { voting_period = { length = $(voting_period) } \
+          { max_quorum = { numerator = ($(max_quorum) * quorum_denominator)/100n } \
+          ; min_quorum = { numerator = ($(min_quorum) * quorum_denominator)/100n } \
+          ; voting_period = { length = $(voting_period) } \
           ; proposal_flush_time = $(proposal_flush_time) \
           ; proposal_expired_time = $(proposal_expired_time) \
-          ; quorum_threshold = $(quorum_threshold) \
+          ; quorum_threshold = { numerator = ($(quorum_threshold) * quorum_denominator)/100n } \
           ; fixed_proposal_fee_in_token = $(fixed_proposal_fee_in_token) \
+          ; quorum_change = { numerator = ($(quorum_change) * quorum_denominator)/100n } \
+          ; max_quorum_change = { numerator = ($(max_quorum_change) * quorum_denominator)/100n } \
+          ; governance_total_supply = $(governance_total_supply) \
           } \
         })"
 	# ================= Compilation successful ================= #
@@ -121,10 +131,16 @@ $(OUT)/registryDAO_storage.tz : now_val = `date +"%s"`
 $(OUT)/registryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/registryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/registryDAO_storage.tz : ledger = ([] : ledger_list)
-$(OUT)/registryDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
+$(OUT)/registryDAO_storage.tz : quorum_threshold = 10n
+$(OUT)/registryDAO_storage.tz : min_quorum = 1n
+$(OUT)/registryDAO_storage.tz : max_quorum = 99n
+
 $(OUT)/registryDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/registryDAO_storage.tz : quorum_change = 5n
+$(OUT)/registryDAO_storage.tz : max_quorum_change = 19n
 $(OUT)/registryDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
 $(OUT)/registryDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
+$(OUT)/registryDAO_storage.tz : governance_total_supply = 100n
 $(OUT)/registryDAO_storage.tz: src/**
 	# ============== Compiling RegistryDAO storage ============== #
 	mkdir -p $(OUT)
@@ -143,13 +159,17 @@ $(OUT)/registryDAO_storage.tz: src/**
             ; ledger_lst = $(call escape_double_quote,$(ledger)) \
             } \
           ; config_data = \
-            { quorum_threshold = $(quorum_threshold) \
+            { max_quorum = { numerator = ($(max_quorum) * quorum_denominator)/100n } \
+            ; min_quorum = { numerator = ($(min_quorum) * quorum_denominator)/100n } \
             ; voting_period = { length = $(voting_period) } \
             ; proposal_flush_time = $(proposal_flush_time) \
             ; proposal_expired_time = $(proposal_expired_time) \
             ; fixed_proposal_fee_in_token = $(fixed_proposal_fee_in_token) \
-            } \
-          } \
+            ; quorum_threshold = { numerator = ($(quorum_threshold) * quorum_denominator)/100n } \
+            ; quorum_change = { numerator = ($(quorum_change) * quorum_denominator)/100n } \
+            ; max_quorum_change = { numerator = ($(max_quorum_change) * quorum_denominator)/100n } \
+            ; governance_total_supply = $(governance_total_supply) \
+            }} \
           ; frozen_scale_value = $(frozen_scale_value) \
           ; frozen_extra_value = $(frozen_extra_value) \
           ; max_proposal_size = $(max_proposal_size) \
@@ -177,10 +197,15 @@ $(OUT)/treasuryDAO_storage.tz : now_val = `date +"%s"`
 $(OUT)/treasuryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/treasuryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/treasuryDAO_storage.tz : ledger = ([] : ledger_list)
-$(OUT)/treasuryDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
+$(OUT)/treasuryDAO_storage.tz : quorum_threshold = 10n
+$(OUT)/treasuryDAO_storage.tz : min_quorum = 1n
+$(OUT)/treasuryDAO_storage.tz : max_quorum = 99n
 $(OUT)/treasuryDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/treasuryDAO_storage.tz : quorum_change = 5n
+$(OUT)/treasuryDAO_storage.tz : max_quorum_change = 19n
 $(OUT)/treasuryDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
 $(OUT)/treasuryDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
+$(OUT)/treasuryDAO_storage.tz : governance_total_supply = 100n
 $(OUT)/treasuryDAO_storage.tz: src/**
 	# ============== Compiling TreasuryDAO storage ============== #
 	mkdir -p $(OUT)
@@ -199,13 +224,17 @@ $(OUT)/treasuryDAO_storage.tz: src/**
             ; ledger_lst = $(call escape_double_quote,$(ledger)) \
             } \
           ; config_data = \
-            { quorum_threshold = $(quorum_threshold) \
+            { max_quorum = { numerator = ($(max_quorum) * quorum_denominator)/100n } \
+            ; min_quorum = { numerator = ($(min_quorum) * quorum_denominator)/100n } \
             ; voting_period = { length = $(voting_period) } \
             ; proposal_flush_time = $(proposal_flush_time) \
             ; proposal_expired_time = $(proposal_expired_time) \
             ; fixed_proposal_fee_in_token = $(fixed_proposal_fee_in_token) \
-            } \
-          } \
+            ; quorum_threshold = { numerator = ($(quorum_threshold) * quorum_denominator)/100n } \
+            ; quorum_change = { numerator = ($(quorum_change) * quorum_denominator)/100n } \
+            ; max_quorum_change = { numerator = ($(max_quorum_change) * quorum_denominator)/100n } \
+            ; governance_total_supply = $(governance_total_supply) \
+            }} \
           ; frozen_scale_value = $(frozen_scale_value) \
           ; frozen_extra_value = $(frozen_extra_value) \
           ; max_proposal_size = $(max_proposal_size) \
--- a/docs/building.md
+++ b/docs/building.md
@@ -52,7 +52,12 @@ make out/trivialDAO_storage.tz \
   metadata_map=(Big_map.empty : metadata_map)
   fixed_proposal_fee_in_token=0n \
   ledger=([] : ledger_list) \
-  quorum_threshold={numerator=1n; denominator=10n} \
+  quorum_threshold = 10n \
+  min_quorum = 1 \
+  max_quorum = 99 \
+  change_percent = 5n \
+  max_change_percent = 19n \
+  governance_total_supply = 100n \
   voting_period=950400n \
   proposal_flush_time = 1900801n \
   proposal_expired_time = 2851200n \
@@ -86,7 +91,12 @@ make out/registryDAO_storage.tz \
   metadata_map=(Big_map.empty : metadata_map) \
   fixed_proposal_fee_in_token=0n \
   ledger=([] : ledger_list) \
-  quorum_threshold={numerator=1n; denominator=10n} \
+  quorum_threshold = 10n \
+  min_quorum = 1n \
+  max_quorum = 99n
+  change_percent = 5n \
+  max_change_percent = 19n \
+  governance_total_supply = 100n \
   voting_period=950400n \
   proposal_flush_time = 1900801n \
   proposal_expired_time = 2851200n
@@ -116,7 +126,12 @@ make out/treasuryDAO_storage.tz \
   metadata_map=(Big_map.empty : metadata_map)
   fixed_proposal_fee_in_token=0n \
   ledger=([] : ledger_list) \
-  quorum_threshold={numerator=1n; denominator=10n} \
+  quorum_threshold = 10n \
+  min_quorum = 1n \
+  max_quorum = 99n \
+  change_percent = 5n \
+  max_change_percent = 19n \
+  governance_total_supply = 100n \
   voting_period=950400n \
   proposal_flush_time = 1900801n \
   proposal_expired_time = 2851200n \
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -105,6 +105,10 @@ type config =
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   ; max_votes : nat
   // ^ Determine the maximum number of votes associated with a proposal including positive votes
+  ; max_quorum_threshold : quorum_threshold
+  // ^ Determine the maximum value of quorum threshold that is allowed.
+  ; min_quorum_threshold : quorum_threshold
+  // ^ Determine the minimum value of quorum threshold that is allowed.
 
   ; voting_period : voting_period
   // ^ Determine the voting_period length.
@@ -121,6 +125,16 @@ type config =
   // ^ Determine the quorum threshold that the proposals need to meet.
   ; fixed_proposal_fee_in_token : nat
   // ^ A base fee paid for submitting a new proposal.
+
+  ; max_quorum_change : quorum_fraction
+  // ^ A percentage value that limits the quorum_threshold change during
+  // every update of the same.
+  ; quorum_change : quorum_fraction
+  // ^ A percentage value that is used in the computation of new quorum
+  // threshold value.
+  ; governance_total_supply : int
+  // ^ The total supply of governance tokens used in the computation of
+  // of new quorum threshold value at each stage.
 
   ; custom_entrypoints : custom_entrypoints
   // ^ Packed arbitrary lambdas associated to a name for custom execution.
@@ -149,6 +163,8 @@ type proposal =
   ; voters : (address * nat) list
   // ^ List of voter addresses associated with the vote amount
   // Needed for `flush` entrypoint.
+  ; quorum_threshold: quorum_threshold
+  // quorum threshold at the cycle in which proposal was raised.
   }
 ```
 
@@ -200,7 +216,7 @@ Each address can have any number of operators and be an operator of any number o
 
 The contract constantly cycles between two stages, a proposing stage and a voting stage.
 Both have the same same length, `voting_period` and alternate between each other,
-starting from "voting" for period number `0`.
+starting from "voting" for period number `0`. A proposal/voting period pair is called a cycle.
 
 The voting period is specified for the whole smart contract and never changes.
 
@@ -221,6 +237,30 @@ Unfreezing does it reverse, that is, the contract makes the FA2 transfer on gove
 contract to transfer tokens from its own address to the address that is doing unfreezing.
 It then burns the corresponding amount of tokens. Only frozen tokens that are not currently
 staked in a vote or proposal can be unfreezed.
+
+The quorum threshold is updated at every cycle change, based on previous participation using
+the formula:
+
+```
+previous participation = number_of_staked_tokens_last_cycle / config.governance_total_supply.
+
+possible_new_quorum =
+  old_quorum * (1 - config.quorum_change) + participation * quorum_change
+
+min_new_quorum =
+  old_quorum / (1 + config.max_quorum_change)
+
+max_new_quorum =
+  old_quorum * (1 + config.max_quorum_change)
+
+new_quorum =
+  max min_new_quorum
+  (min max_new_quorum
+       possible_new_quorum)
+```
+
+This will use the configuration values provided at origination, and the new quorum will be still
+bound by the max/min quorum threshold values provided in the config.
 
 ### Proposals
 
@@ -291,6 +331,7 @@ The list of errors may be inaccurate and incomplete, it will be updated during t
 | `BAD_TOKEN_CONTRACT`             | Throw when the token contract is not of expected type.                                                           |
 | `DROP_PROPOSAL_CONDITION_NOT_MET`| Throw when calling `drop_proposal` when the sender is not proposer or guardian and proposal is not expired.      |
 | `EXPIRED_PROPOSAL`               | Throw when trying to `flush` expired proposals.                                                                  |
+| `BAD_STATE`                     | Throw when storage is in an unexpected state, indicating a contract error.                                  |
 
 # Entrypoints
 
@@ -695,6 +736,9 @@ have the same timestamp due to being in the same block, are processed in the ord
     - The return amount for the proposer is equal to or less than the sum of the proposer frozen tokens and the fee paid for the proposal.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
 - If proposal is accepted, decision lambda is called.
+- The quorum_threshold at the cycle in which the proposal was raised will be stored in the proposal, and this
+  threshold will be used to check if the votes meet the quorum threshold. So any dynamic update to the quorum threshold shall not affect
+  the proposal.
 
 ### **drop_proposal**
 
@@ -769,6 +813,7 @@ Parameter (in Michelson):
 
 - Author MUST have tokens equal to `freeze_param` or more, in the governance contract, with token id
   'governance-token.token_id`.
+
 
 ### **unfreeze**
 
--- a/haskell/test/Test/Ligo/BaseDAO/Integrational/Proposal.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Integrational/Proposal.hs
@@ -21,7 +21,7 @@ import Test.Tasty.HUnit (testCase)
 import Ligo.BaseDAO.Types
 import Michelson.Test.Integrational
 import Michelson.Untyped (unsafeBuildEpName)
-import Test.Ligo.BaseDAO.Common (dummyFA2Contract)
+import Test.Ligo.BaseDAO.Common (dummyFA2Contract, makeProposalKey)
 import Test.Ligo.BaseDAO.Integrational.Common
 import Util.Named
 
@@ -32,13 +32,21 @@ test_IntegrationalBaseDAOProposal = testGroup "BaseDAO Integrational Tests"
   [ testGroup "Integrational Tests for checking storage"
       [ testCase "contract correctly tracks freeze-unfreeze history" $
           integrationalTestExpectation $ checkFreezeHistoryTracking
+      , testCase "contract correctly updates quorum threshold and resets outdated staked token count" $
+          integrationalTestExpectation $ checkQuorumThresholdDynamicUpdate
+      , testCase "contract correctly updates quorum threshold within max bound" $
+          integrationalTestExpectation $ checkQuorumThresholdDynamicUpdateUpperBound
+      , testCase "contract correctly updates quorum threshold within min bound" $
+          integrationalTestExpectation $ checkQuorumThresholdDynamicUpdateLowerBound
+      , testCase "contract correctly stores quorum threshold within the proposal" $
+          integrationalTestExpectation $ checkProposalSavesQuorum
       ]
   ]
 
 checkFreezeHistoryTracking :: IntegrationalScenarioM ()
 checkFreezeHistoryTracking  = do
   let frozen_scale_value = 2
-  let frozen_extra_value = 0
+  let frozen_extra_value = 10
 
   now <- use isNow
 
@@ -47,11 +55,14 @@ checkFreezeHistoryTracking  = do
     pure $ mkFullStorage
       ! #admin admin
       ! #votingPeriod 10
-      ! #quorumThreshold (QuorumThreshold 10 100)
+      ! #quorumThreshold (mkQuorumThreshold 10 100)
       ! #extra dynRecUnsafe
       ! #metadata mempty
       ! #now now
       ! #tokenAddress (unTAddress dodTokenContract)
+      ! #maxChangePercent 19
+      ! #changePercent 5
+      ! #governanceTotalSupply 100
       ! #customEps mempty
     ) $ \(admin:wallet1:_) baseDao -> let
       proposalMeta1 = ""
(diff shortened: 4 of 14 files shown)


src/types.mligo
```diff
@@ -146,6 +146,17 @@ type governance_token =
...
+type period_frozen_history =
+  { current : nat
+  ; last : nat
+  ; updated_period : nat
+  }
+
+type quorum_threshold_with_period =
+  { quorum_threshold : quorum_threshold
+  ; updated_period : nat
+  }
```
pasqu4le, line 158: I think this representation can be simplified a lot and as a result the code would be simpler as well.

AFAIU for a given period `x` all the information needed to calculate the `quorum_threshold` are:
- constants
- the `quorum_threshold` of period `x - 1` (aka `old_quorum`)
- the `total_supply` of frozen tokens at period `x - 1`

(the exception is period `0`, but that's not a problem as all the info is known at the beginning)

So what I think we should so is track all we need in a single type:
```ocaml
type quorum_threshold_at_stage =
  { quorum_threshold : quorum_threshold
  ; frozen_supply: nat
  ; last_updated_period : nat
  }
```
To keep the information in it correct (and enough) we need to update this lazily in its entirety whenever:
- we change the `total_supply` of frozen tokens (in `freeze`/`unfreeze`)
- we need the `quorum_threshold`

I'll write more comments about the updates below.

src/proposal.mligo
```diff
@@ -408,6 +484,39 @@ let drop_proposal (proposal_key, config, store : proposal_key * config * storage
   else
     (failwith("FAIL_DROP_PROPOSAL_NOT_OVER") : return)
 
+let update_period_frozen_history(amt, current_period, period_frozen_history: int * nat * period_frozen_history) : period_frozen_history
+  = match is_nat(current_period - period_frozen_history.updated_period) with
```
pasqu4le, line 488: If picking the representation in the comment above here we should do two things (in this order).

First: check if the `quorum_threshold_at_stage.last_updated_period` is not `current_period`.
In case it isn't, we need to:
1. calculate the new `quorum_threshold`, which we can do as we have all the info from a previous period
2. set the `last_updated_period` to `current_period`

Second, regardless of what we do above we now know that we are storing the info for the `current_period`, so we can update the `frozen_supply` with the given value.
Note: BTW in this representation we can "safely" subtract in case of unfreezing: if we track the `total_supply` we know that no value should not be higher than that,so if this this is not true we should fail as there's a bug in the contract.

src/proposal.mligo
```diff
@@ -448,4 +558,5 @@ let unfreeze (amt, config, store : unfreeze_param * config * storage) : return =
         ledger = ledger
       ; total_supply = total_supply
       ; freeze_history = new_freeze_history
+      ; period_frozen_history = update_period_frozen_history(int(amt), current_period, store.period_frozen_history)
```
pasqu4le, line 561: Note: this looks incorrect BTW, probably the it should be `- amt`

src/proposal.mligo
```diff
@@ -337,6 +375,42 @@ let delete_proposal
     Set.remove (start_date, proposal_key) store.proposal_key_list_sort_by_date
   }
 
+let update_quorum(store, config : storage * config): storage =
```
pasqu4le, line 378: If picking the representation in the comment above here we should do only one thing, which is the first step from the comment above for `update_period_frozen_history` and that's it.

src/types.mligo
```diff
@@ -161,6 +172,8 @@ type storage =
   ; freeze_history : freeze_history
```
pasqu4le, line 172: While writing the previous comments I realized that we don't need to store `frozen_supply` inside of `quorum_threshold_at_stage` at all (I didn't edit the previous comments as they maybe show the point better).

We already know that value right here, in `total_supply`.

If we remember to update the `quorum_threshold_at_stage` every time we `freeze` or `unfreeze`, we already know that in period `x` the `total_supply` of `frozen_token_id` here in the storage is the same as either:
1. the total_supply of frozen tokens at period `x - 1` (that we can use to update the quorum)
2. the total_supply of frozen tokens at period `x` (that we don't need to use as the quorum is already up-to-date)

pasqu4le (commented):
I left some major comments that change the representation used here, I suggest you read them all first when considering them.

sras:
@pasqu4le Your comments appear to make sense to me. 

But one more thing. Do you think we should normalize/reduce the `quorum_threshold` fraction at some point? Since we are storing it as numerator/denominator pairs, and use them in calculations as such, isn't there chance for them to become arbitrarly large after many such calculation?

src/types.mligo
sras: @pasqu4le I doubt.

>>We already know that value right here, in total_supply.

To be clear how I understand your sugeestion, is that in `freeze` entrypoint, we can possibly look at the `quorum_threshold_at_stage`, and when we see that its `updated_at` period is less than current one, we can use the current `total_supply` as the number of frozen_tokens in last period, and use it to update the quorum_threshold, and then proceed to actually freeze the tokens.

But doesn't `total_supply` at this point  include tokens that were frozen in all period upto the current one, and not limited to tokens that were frozen in (x -1)? 

Or doesn't that matter at all?

Makefile
```diff
@@ -91,9 +94,14 @@ $(OUT)/trivialDAO_storage.tz: src/**
           ; ledger_lst = $(call escape_double_quote,$(ledger)) \
           } \
         ; config_data = \
-          { voting_period = { length = $(voting_period) } \
+          { max_quorum = {numerator = 99n; denominator = 100n} \
+          ; min_quorum = {numerator = 1n; denominator = 100n} \
```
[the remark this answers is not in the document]
sras: Have added

Makefile
```diff
@@ -116,6 +124,9 @@ $(OUT)/registryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/registryDAO_storage.tz : ledger = ([] : ledger_list)
 $(OUT)/registryDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
 $(OUT)/registryDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/registryDAO_storage.tz : change_percent = 5n
+$(OUT)/registryDAO_storage.tz : max_change_percent = 19n
+$(OUT)/registryDAO_storage.tz : frozen_total_supply = 100n
```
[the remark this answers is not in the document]
sras: updated

src/proposal.mligo
sras: This has been removed.

pasqu4le (commented):
I left some comments about cost of operation, but I want to add that the Ligo implementation is correct AFAICT, just possibly too expensive.

src/proposal.mligo
```diff
@@ -337,6 +391,39 @@ let delete_proposal
...
+  if store.quorum_threshold_at_stage.last_updated_period = current_period
+    then store // Quorum has been updated in this period, so no change is required.
+    else
+      let last_period_frozen_tokens = match Map.find_opt 0n store.total_supply with
+        | Some (n) -> n
+            // We assume that this function will always be called before freezing
+            // tokens, so that if we are in this branch (where quorum_threshold_at_stage
+            // is outdated), we assume the total supply will be outdated as well and will
+            // have frozen tokens available in last/earlier period.
+        | None -> ((failwith "FROZEN_TOTAL_SUPPLY_NOT_FOUND") : nat) in
```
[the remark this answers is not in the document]
pasqu4le: well yes, but if we got here we are already in a catastrophically inconsistent state...

The idea is to keep the contract as simple as possible (to save on execution/size costs) and prove with tests that (in this case) none of the functions that modify the `total_supply` can remove this entry.

src/proposal.mligo
```diff
@@ -73,6 +79,54 @@ let is_ge_qt(qt_1, qt_2 : quorum_threshold * quorum_threshold): bool =
...
+// HCF of two naturals
+let hcf(n1, n2: nat * nat): nat =
+  if (n1 > n2) then hcf_(n1, n2) else hcf_(n2, n1)
+
+[@inline]
+// Normalize a fraction so that the numerator and denominator are smallest
+// possible values.
+let normalize_qt(qt: quorum_threshold): quorum_threshold = let
+  hcf = hcf(qt.numerator, qt.denominator)
+  in {numerator = qt.numerator / hcf; denominator = qt.denominator / hcf }
```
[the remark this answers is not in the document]
pasqu4le: Yeah, that may be a good idea too. There's many calculations involving percentages, I'd imagine that often the denominator could be divided by `100` and the nominator kept sufficiently detailed (Ideally however if we can avoid unnecessary multiplications it would be better).

Considering that the denominator needs to always be bigger than the numerator, we can pick one sufficiently large and assume that it's detailed enough to be meaningful.

src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: I think it is, the only case in which this could happen is:
1. the contract was originated with an empty `total_supply` in which case the behavior of `0n` is correct.
2. the contract has a serious bug and as a result there is no `total_supply` of frozen tokens

I mean this is an error that we cannot even test for, unless we purposefully originate the contract without a `total_supply` entry, in which case as I said is ok to behave like 0.

In case of (2) the contract:
- if with an error is stuck: as there is no way to `freeze`, `unfreeze`, `flush` or `drop_proposal`
- without an error produces a 0% `previous_participation`, which results in a lower quorum threshold

I agree that in general this is something to raise an error for, but this is one case where the disadvantage can be ignored, as the problem that caused it (no `total_supply`) has a far larger impact.
In other word, all the entrypoints "affected" by this returning `0n` don't work without a frozen `total_supply` anyhow:
- `freeze` will throw `FA2_TOKEN_UNDEFINED` in `credit_to`
- `unfreeze` will do the same in `debit_from`
- `flush` will also fail, because even in config proposals and such it has to give back/consume/burn tokens 
- `drop_proposal` will also throw `FA2_TOKEN_UNDEFINED` for either `credit_to` or `debit_from`

whatever this returns in such a case is in-influential, as the contract is utterly broken either way.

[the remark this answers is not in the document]
sras: First of all, I think it was a bad idea to re-purpose `total_supply` to indicate the support for various token ids. I think that is the root cause that we find ourself with disagreement here. Because in normal case `total_supply` to be zero would be perfectly fine. 

The problem with your reasoning that I see is this. Once we find ourselves in an inconsistent state at some point in future, any answer to the question as to how we ended up in that state, can only be speculative. Because assumptions that you make right now about what the rest of the code is doing, might not hold at that point in future.

Now, to take this discussion a bit in a different direction, I am not sure what problem you find with throwing error, instead of returning zero. You say that we want to keep the contract simple, but AFAIU, throwing this error does not add to its complexity in any way.

src/proposal.mligo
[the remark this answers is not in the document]
sras: > I'd imagine that often the denominator could be divided by 100 and the nominator kept sufficiently detailed 

I have included a rounding off behavior as you suggest here. Right now it normalizes and rounds off after it. May be only rounding off is enough. What do you think?

src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: > I think it was a bad idea to re-purpose total_supply to indicate the support for various token ids.
> in normal case total_supply to be zero would be perfectly fine.

That's true, however it's a very simple way to support this, a more clear alternative would have been to add a field in storage just for the set of supported `token_id`, but that's an added cost that we are so avoiding.

> any answer to the question as to how we ended up in that state, can only be speculative

That's fair, but we know for sure that it wasn't intended and my reasoning is based on the results of such a situation, not the cause. What happens to the entrypoints' logic is true regardless of how we got there.

> I am not sure what problem you find with throwing error, instead of returning zero.
> throwing this error does not add to its complexity in any way.

Well I don't have a problem with it really, we can go with it and as I said it's the proper thing to do.
My only motivations are that:
1. it adds cost, as `PUSH 0` is less expensive then `PUSH "FROZEN_TOTAL_SUPPLY_NOT_FOUND"; FAILWITH`.
    (both to originate and run)
2. it doesn't add any more safety, as the entrypoints affected could not be executed in that case regardless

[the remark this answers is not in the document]
pasqu4le: Your points are valid tho, IMO we can leave this for later as the discussion is more based on our use of the `total_supply` to indicate the support for tokens and a slight difference in efficiency.

We can go with your `FAILWITH` approach and get back to this discussion if we need to cut the cost down (or reconsider the `total_supply` field use), I'm ok with that :+1: 
If you do so however, do the same in `do_total_vote_meet_quorum_threshold` for consistency and remember to document this error in the spec as well, ok?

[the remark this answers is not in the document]
sras: > it doesn't add any more safety, as the entrypoints affected could not be executed in that case regardless

Just want to clarify. 

AFAIU you are talking about a case where the execution start from a corrupt state. But since we pass around storage through various funcions, and  if one of them modify storage to have this inconsistent state, and throwing the exception as soon as we come across it might save ourselfves from writing down an inconsistent version of the storage, permanantly to the contract storage.

So in that regard,  I think it's more safe, because it allows us to recover storage from a valid state, rather than a corrupted one. 

>If you do so however, do the same in do_total_vote_meet_quorum_threshold for consistency and remember to document this error in the spec as well, ok?

Will do.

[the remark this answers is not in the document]
sras: Have added the error throwing to `do_total_vote_meet_quorum_threshold` function and updated spec. I have renamed the error to just 'BAD_STATE', as a generic error.

src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: I think rounding off will be enough, avoiding normalization will likely save a considerable amount of gas.

src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: > throwing the exception as soon as we come across it might save ourselfves from writing down an inconsistent version of the storage

Well keeping in mind that the storage doesn't actually change if anything in the whole operation fails this is less of a concern here, where we one of the failing functions is called in the same entrypoint.
However that's a valid point, it would be difficult to verify that this is the case and keeps being the case in the future. :ok_hand:

src/proposal.mligo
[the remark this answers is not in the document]
sras: @pasqu4le I have removed the normalization function and only uses a recursive rounding function now. It is recursive because sometimes it appears that one iteration is not enough to bring the pair under the limit. May be it could be made more efficient.

src/proposal.mligo
```diff
@@ -337,6 +391,39 @@ let delete_proposal
...
+            // have frozen tokens available in last/earlier period.
+        | None -> ((failwith "FROZEN_TOTAL_SUPPLY_NOT_FOUND") : nat) in
+      let previous_participation = { numerator = last_period_frozen_tokens; denominator = config.governance_total_supply } in
+      let old_quorum = store.quorum_threshold_at_stage.quorum_threshold in
+      let one = {numerator = 1n; denominator = 1n} in
+      let possible_new_quorum =
+        let term1 = mul_qt(old_quorum, sub_qt(one, {numerator = config.change_percent; denominator = 100n})) in
+        let term2 = mul_qt(previous_participation, {numerator = config.change_percent; denominator = 100n}) in
+        add_qt(term1, term2) in
+      let max_change_percent = { numerator = config.max_change_percent; denominator = 100n } in
```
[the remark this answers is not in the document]
sras: Have included it.

src/proposal.mligo
```diff
@@ -337,6 +382,40 @@ let delete_proposal
...
+      let one_plus_max_change_percent = { numerator = 100n + config.max_change_percent; denominator = 100n } in
+      let min_new_quorum =
+        div_qt(old_quorum, one_plus_max_change_percent) in
+      let max_new_quorum =
+        mul_qt(old_quorum, one_plus_max_change_percent)
```
pasqu4le, line 409: I tried to reduce the cost of these calculations as much as possible,

Sadly it seems that using fraction-based operations adds a significant cost as probably ligo needs to pair and unpair many them continuously.

IMO we should go with a solution that takes as little as possible and leave comment to explain what's happening more clearly, here's what I could come up with:
```suggestion
      let old_quorum = store.quorum_threshold_at_stage.quorum_threshold in
      let old_quorum_numer = old_quorum.numerator in
      let old_quorum_denom = old_quorum.denominator in
      // participation = last_period_frozen_tokens / governance_total_supply
      // possible_new_quorum = old_quorum * (1 - change_percent) + participation * change_percent
      let possible_new_quorum =
        let term1 =
              { numerator = old_quorum_numer * sub_nat (100n, config.change_percent)
              ; denominator = old_quorum_denom * 100n
              }
        in
        let term2 =
              { numerator = last_period_frozen_tokens * config.change_percent
              ; denominator = config.governance_total_supply * 100n
              }
        in
        add_qt(term1, term2) in
      let one_plus_max_change_percent = 100n + config.max_change_percent in
      // min_new_quorum = old_quorum / (1 + max_change_percent)
      let min_new_quorum =
        { numerator = old_quorum_numer * 100n
        ; denominator = old_quorum_denom * one_plus_max_change_percent
        }
      in
      // max_new_quorum = old_quorum * (1 + max_change_percent)
      let max_new_quorum =
        { numerator = old_quorum_numer * one_plus_max_change_percent
        ; denominator = old_quorum_denom * 100n
        }
```
This cuts the size of the compiled contract quite a bit (`18982` -> `18818` bytes) and due to the lower number of operations I'd imagine the gas cost even more.

(note: you'll need a `sub_nat` which just throws an error if the subtraction is negative)

src/proposal.mligo
```diff
@@ -73,6 +79,45 @@ let is_ge_qt(qt_1, qt_2 : quorum_threshold * quorum_threshold): bool =
...
+    else qt
+
+let bound_qt (qt, min_qt, max_qt : quorum_threshold * quorum_threshold * quorum_threshold)
+    : quorum_threshold =
+  if is_gt_qt(qt, max_qt)
+  then max_qt
+  else
+    if is_le_qt(qt, min_qt)
+    then min_qt
+    else qt
```
pasqu4le, line 119: Now that we use this here and more than once it makes sense to inline it.

Also I think it can be redefined to only use `is_gt_qt`, which I suggest to do to have one less function in the code.
```suggestion
[@inline]
let bound_qt (qt, min_qt, max_qt : quorum_threshold * quorum_threshold * quorum_threshold)
    : quorum_threshold =
  if is_gt_qt(qt, max_qt)
  then max_qt
  else if is_gt_qt(min_qt, qt)
  then min_qt
  else qt
```

src/proposal.mligo
```diff
@@ -73,6 +79,45 @@ let is_ge_qt(qt_1, qt_2 : quorum_threshold * quorum_threshold): bool =
...
+[@inline]
+// Subtracts two fractions, and throws error if second one is larger.
+let sub_qt(qt_1, qt_2 : quorum_threshold * quorum_threshold): quorum_threshold = let
+  n = match is_nat((qt_1.numerator * qt_2.denominator) - (qt_2.numerator * qt_1.denominator)) with
+    | Some n -> n
+    | None -> ((failwith "UNEXPECTED_NEGATIGE_FRACTION_RESULT") : nat) in
+  { numerator = n; denominator = qt_1.denominator * qt_2.denominator }
+
+[@inline]
+let rec round(qt: quorum_threshold): quorum_threshold =
```
pasqu4le, line 107: I think `rec` is unnecessary here.

src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: @sras I think we should avoid recursive functions and that this can be implemented with a single execution (well enough).

Keeping in mind that the `numerator` is always smaller than the `denominator` we can:
1. calculate by what "factor" we want to reduce the fraction
2. divide both values by that "factor"

I also suggest to increase the max `denominator` to a million (which should still be small enough to not be near to the `nat` upper limit) and that such a number would be precise enough to not need the `numerator` check.

So, my suggestion is to use this:
```
[@inline]
let round(qt: quorum_threshold): quorum_threshold =
  let factor = qt.denominator / 1000000n in
  if factor > 1n
  else { numerator = qt.numerator / factor; denominator = qt.denominator / factor}
  then qt
```

Dealing with `nat` division, in the worst case we can obtain either:
- a `numerator` of one, which is still precise to the 0.0001%
- a `denominator` of 2 millions (or maybe less, not sure how `nat` division rounds) which is still not a problem

What do you think, does that sound correct?

src/proposal.mligo
sras: > Sadly it seems that using fraction-based operations adds a significant cost as probably ligo needs to pair and unpair many them continuously.

Would it be better to replace numerator denominator pairs with a single `nat` with a known precision. AFAIU this is how Tezos core does it. For example `1.95` will be stored as `195`. 

Can it significantly reduce the number of calculations?

docs/specification.md
```diff
@@ -744,6 +751,11 @@ Parameter (in Michelson):
 - Author MUST have tokens equal to `freeze_param` or more, in the governance contract, with token id
   'governance-token.token_id`.
 
+- Before actually freezing the tokens, this procedure also updates the quorum threshold
+  if required, after checking the last period it was updated at.
+  We do it here because this way, we will have access to the tokens that were frozen in
+  the last period, without having to track it separately.
```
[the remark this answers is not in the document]
sras: Changed

src/proposal.mligo
```diff
@@ -316,12 +361,12 @@ let is_voting_period_over (proposal, voting_period, store : proposal * voting_pe
...
 
 [@inline]
-let do_total_vote_meet_quorum_threshold (proposal, store, quorum_threshold : proposal * storage * quorum_threshold): bool =
+let do_total_vote_meet_quorum_threshold (proposal, store, voting_period, quorum_threshold : proposal * storage * voting_period * quorum_threshold): bool =
   let votes_placed = proposal.upvotes + proposal.downvotes in
   let total_supply =
         match Map.find_opt store.frozen_token_id store.total_supply with
         | Some v -> v
-        | None -> 0n
+        | None -> ((failwith "BAD_STATE") : nat)
```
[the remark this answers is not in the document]
sras: Yes, right. Done.

src/proposal.mligo
pasqu4le: @sras that's a very good idea, let's go with it! 

However I still think that percentage is not precise enough, I suggest once again that we use a per-million instead.

That should be safe as we know that the biggest numerator is a million as well and that the biggest number we can obtain within an operation is when multiplying two numbers, so `10 ^ 12`, which is still smaller than the max `nat` (that AFAIK is represented like a `63` bit unsigned int).

> same with division as well and a/b can be just `div (a*100) b`

Actually `a/b` should be just ... `a/b` too (the rest looks correct).

BTW I suggest documenting this properly and still keep a type synonym to better keep track of this:
```ocaml
// Quorum threshold that a proposal needs to meet in order to be accepted,
// expressed as a fraction of the total_supply of frozen tokens.
//
// For efficiency, we only keep a `nat` for the numerator, whereas the denominator
// is not stored and has a fixed value of `1000000`.
//
// Invariant: numerator < denominator
type quorum_threshold = nat
```

sras: >Actually a/b should be just ... a/b too (the rest looks correct).

But it seems that it would lose precision , because the decimal places will be lost. for ex `24.56/47.34` would be `2456/4734` which would be just `0`. But multiplying numerator with `100`, it would result in `51`, which the approximation we want in this representation of the actual answer, `0.5188`.

Agree with the rest.

pasqu4le: > it seems that it would lose precision

Oh yes, good point, I didn't think about the fact that it was a `nat` operation for a moment, never mind my comment there :+1:

sras: @pasqu4le I have included this change. In addition, I have also changed it to use `int` instead of `nat` so that it could momenerly accomodate negative values to simplify calculations. And consequently, I have changed some `config` values to be `int` so that it ll save some `nat`/`int` conversion in quorum_calculation.

pasqu4le: @sras instead of changing the type to be an `int`, which could store a potentially wrong value (because negative) I think we should keep the `nat` in `quorum_threshold`.

This should not be a problem as it's not necessary for us to hold a `nat`/`quorum_threshold` after every operation, we can use `int` in the middle steps and then make one conversion before saving the result.

docs/specification.md
```diff
@@ -749,6 +756,9 @@ Parameter (in Michelson):
 - Author MUST have tokens equal to `freeze_param` or more, in the governance contract, with token id
   'governance-token.token_id`.
 
+- This procedure also checks if quorum threshold requires updating, and updates
+  if found necessary.
```
pasqu4le, line 760: I think these can also be removed as an implementation detail
```suggestion
```

docs/specification.md
```diff
@@ -766,6 +776,8 @@ Parameter (in Michelson):
   on the governance contract from the address of the baseDAO contract to the address of the sender.
 - Fails with `NOT_ENOUGH_FROZEN_TOKENS` if the author does not have enough tokens that can be burned.
 
+- Similarly in the freeze procedure, we update the quorum threshold here as well.
+
```
pasqu4le, line 780:
```suggestion
```

docs/specification.md
```diff
@@ -692,6 +698,7 @@ proposal are returned to the proposer and voters in full. The decision lambda is
 - This entrypoint should only be used when there is a proposal that is stuck due to having a
 failing decision lambda.
 - Fails with `NOT_ADMIN` if the sender is not the administrator.
+- Similarly in the freeze procedure, we update the quorum threshold here as well, before doing anything.
```
pasqu4le, line 701:
```suggestion
```

docs/specification.md
```diff
@@ -673,6 +678,7 @@ have the same timestamp due to being in the same block, are processed in the ord
     - The return amount for the proposer is equal to or less than the sum of the proposer frozen tokens and the fee paid for the proposal.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
 - If proposal is accepted, decision lambda is called.
+- Similarly in the freeze procedure, we update the quorum threshold here as well, before doing anything.
```
pasqu4le, line 681:
```suggestion
```

docs/specification.md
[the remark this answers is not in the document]
pasqu4le: @sras I think you missed this part?
> we should document the expected result, which is that the quorum is updated at every stage (I would include it in that section)

To be clear I meant adding a line on the `stages` section of the spec that says something like:
```
The quorum threshold is updated at every stage change, based on the formula:
[...]
using the configuration values provided at origination.
```

src/types.mligo
```diff
@@ -122,14 +122,15 @@ type proposal =
 type seconds = nat
 type voting_period = { length : seconds }
 
+// For efficiency, we only keep a `nat` for the numerator, whereas the
+// denominator is not stored and has a fixed value of `1000000`.
+type fraction = { numerator : int }
+let fraction_denominator = 1000000
+let percent_to_fraction = 10000
```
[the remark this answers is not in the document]
sras: I think in the update_quorum procedure, we use this to represent percentages and some intermediate fractional values. It felt odd to use a type with 'quorum_` in it.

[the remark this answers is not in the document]
pasqu4le: > we use this to represent percentages and some intermediate fractional values

Yeah, but still only fractions for the same denominator :thinking: 

My point is that this cannot represent any fraction or rational number, but only the ones where the denominator is fixed at 1 million, which is a choice we made for `quorum`s for a practical reason.

This is why I still think it's ok to keep this as `quorum_fraction` (as "a fraction for quorum") and maintain the synonym of `quorum_threshold`.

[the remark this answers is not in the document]
sras: >My point is that this cannot represent any fraction or rational number

AFAIU its a fixed precision representation, so it can represent any fraction to the precision decided by the deniminator. 

But I am fine with changing the name :+1:

docs/specification.md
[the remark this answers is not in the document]
sras: Have added it.

src/proposal.mligo
sras: Have changed quorum threshold to be an Int.

src/proposal.mligo
```diff
@@ -337,6 +368,41 @@ let delete_proposal
...
+        // old_quorum * (1 - change_percent) + participation * change_percent
+        // simplified as
+        // old_quorum + (participation - old_quorum) * change_percent
+        fadd(old_quorum, fmul(change_percent_as_fraction, fsub(previous_participation, old_quorum))) in
+      let max_change_percent = { numerator = config.max_change_percent * percent_to_fraction } in
+      let one_plus_max_change_percent = { numerator = max_change_percent.numerator + fraction_denominator } in
+      let min_new_quorum =
+        fdiv(old_quorum, one_plus_max_change_percent) in
+      let max_new_quorum =
+        fmul(old_quorum, one_plus_max_change_percent)
```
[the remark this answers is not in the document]
sras: Have updated. Couldn't appy the suggestion as such because it conflicted with a commit I had pushed.

docs/specification.md
```diff
@@ -102,6 +102,10 @@ type config =
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   ; max_votes : nat
   // ^ Determine the maximum number of votes associated with a proposal including positive votes
+  ; max_quorum_threshold : quorum_threshold
+  // ^ Determine the maximum value of quorum threshold that is allowed.
+  ; min_quorum_threshold : quorum_threshold
+  // ^ Determine the minimum value of quorum threshold that is allowed.
```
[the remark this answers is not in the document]
sras: Have updated.

src/proposal.mligo
```diff
@@ -54,24 +54,36 @@ let ensure_proposal_is_unique (propose_params, store : propose_params * storage)
...
-
+// Multiply two quorum_fractions
+//
+// We store fractions by storing only the numerator and denominator is always
+// assumed to be quorum_denominator. So here qt_1, actually represents the
+// value qt_1.numerator/ quorum_denominator and qt_2 represents the value
+// qt_2.numerator/ quorum_denominator. So the product of the two is
+// qt_1.numerator * qt_2.numerator / (quorum_denominator *
+// quorum_denominator).  But since we store x as x * quorum_denominator,
+// the result we here would be qt_1.numerator * qt_2.numerator /
```
pasqu4le, line 65: `we here` seems a typo

src/proposal.mligo
```diff
@@ -54,24 +54,36 @@ let ensure_proposal_is_unique (propose_params, store : propose_params * storage)
...
+[@inline]
+let fmul(qt_1, qt_2 : int * int): int =
+   (qt_1 * qt_2) / quorum_denominator
+
+// Divide the first fraction by the second Here qt_1, actually represents the
+// value qt_1.numerator/ quorum_denominator and qt_2 represents the value
+// qt_2.numerator/ quorum_denominator. So the division can be expressed as
+// (qt_1.numerator * quorum_denominator) / (quorum_denominator *
+// qt_2.numerator) But since we store x as x * quorum_denominator, the result
+// we here would be (qt_1.numerator * quorum_denominator) / qt_2.numerator
```
pasqu4le, line 77: `we here` here as well.

src/proposal.mligo
```diff
@@ -316,18 +328,21 @@ let is_voting_period_over (proposal, voting_period, store : proposal * voting_pe
...
 
 [@inline]
-let do_total_vote_meet_quorum_threshold (proposal, store, quorum_threshold : proposal * storage * quorum_threshold): bool =
+let frozen_total_supply(store : storage): nat =
+    match Map.find_opt store.frozen_token_id store.total_supply with
+    | Some v -> v
+    | None -> ((failwith "BAD_STATE") : nat)
+
+[@inline]
+let do_total_vote_meet_quorum_threshold (proposal, store, voting_period, quorum_threshold : proposal * storage * voting_period * quorum_threshold): bool =
```
pasqu4le, line 337: `voting_period` was added here, but it's not used.

src/proposal.mligo
```diff
@@ -316,18 +328,21 @@ let is_voting_period_over (proposal, voting_period, store : proposal * voting_pe
...
   let votes_placed = proposal.upvotes + proposal.downvotes in
-  let total_supply =
-        match Map.find_opt store.frozen_token_id store.total_supply with
-        | Some v -> v
-        | None -> 0n
+  let total_supply = frozen_total_supply(store)
   in
```
pasqu4le, line 340:
```suggestion
  let total_supply = frozen_total_supply(store) in
```

src/proposal.mligo
```diff
@@ -316,18 +328,21 @@ let is_voting_period_over (proposal, voting_period, store : proposal * voting_pe
...
   // Note: this is equivalent to checking that the number of votes placed is
   // bigger or equal than the total supply of frozen tokens multiplied by the
   // quorum_threshold proportion.
-  let reached_quorum = {numerator = votes_placed; denominator = total_supply} in
-  is_ge_qt(reached_quorum, quorum_threshold)
+  let reached_quorum = (votes_placed * abs(quorum_denominator)) / total_supply in
+  (reached_quorum > quorum_threshold)
```
pasqu4le, line 345:
```suggestion
  (reached_quorum >= quorum_threshold)
```

src/types.mligo
```diff
@@ -122,14 +122,17 @@ type proposal =
 type seconds = nat
 type voting_period = { length : seconds }
 
+// For efficiency, we only keep a `nat` for the numerator, whereas the
+// denominator is not stored and has a fixed value of `1000000`.
+type quorum_fraction = { numerator : int }
+let quorum_denominator = 1000000
+let percent_to_quorum = 10000
```
pasqu4le, line 129: @sras [here](https://github.com/tqtezos/baseDAO/pull/225#discussion_r632692172) you said:
> Have changed quorum threshold to be an Int.

but looks like you changed `quorum_threshold` to be a `nat` and `quorum_fraction` to have an `int` in it?

I don't get the reason for this change, my point was that in these types we be better off using a `nat` as we want to
verify that the value isn't negative.

What I meant was that we can use `int`s as an intermediate type in the calculations of `update_quorum` without wrapping them in a `numerator` to be a `quorum_threshold` or `quorum_fraction` (as we indeed do now).

But IMO here we can have:
```suggestion
// For efficiency, we only keep a `nat` for the numerator, whereas the
// denominator is not stored and has a fixed value of `1000000`.
type quorum_fraction = { numerator : nat }
let quorum_denominator = 1000000n
let percent_to_quorum = 10000n
```
and just below do:
```ocaml
type quorum_threshold = quorum_fraction
```

and AFAICT `update_quorum` would remain the same, aside from moving some calculation of `int()`s.

Does this make sense to you?

sras: @pasqu4le It generally makes sense to me. But I don't really think it is a good idea to use plain `int` as intermediate for calculations (eventhough that is how it has been implemented in the last commit I pushed yesterday). AFAIU we can use a record with a single field for this without incurring no additional cost. The reasoning for this is that it is hard to track `int`s that should be treated as a numerator of a fraction, and ones that should be treated as such. I think a comment is too ineffective to consistently provide this distinction in a code base that is still evolving.

So what I have now is two types, `unsigned_quorum_fraction` which uses a `nat` for numerator and `quorum_fraction` which uses an `int` for the same. We use `unsigned_quorum_fraction` every where in storage and config, and use `quorum_fraction` for doing computation. As I said earlier, AFAIU it should be completely free, and the same as using raw `int`s, so long as we inline functons like `fsub` and `fadd`.

src/proposal.mligo
sras: Fixed

src/proposal.mligo
sras: Fixed.

src/proposal.mligo
sras: Not sure why it was added, but have removed it now.

src/proposal.mligo
sras: Have fixed this. Have also added a test for this, but I couldn't target an exact quorum since it is recomputed even during the first proposal.

src/types.mligo
```diff
@@ -275,9 +289,13 @@ type config =
...
   ; max_proposals : nat
   ; max_votes : nat
-  ; quorum_threshold : quorum_threshold
+  ; max_quorum_threshold : quorum_threshold
+  ; min_quorum_threshold : quorum_threshold
   ; voting_period : voting_period
   ; fixed_proposal_fee_in_token : nat
+  ; max_quorum_change : quorum_fraction
+  ; quorum_change : quorum_fraction
+  ; governance_total_supply : int
```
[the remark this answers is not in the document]
sras: Have changed

src/types.mligo
pasqu4le: That's a great point, using two different "newtype"s seems like the most elegant solution to me :ok_hand:

src/defaults.mligo
```diff
@@ -57,8 +67,18 @@ let default_storage (data, config_data : initial_storage_data * initial_config_d
...
-}
+    start_time = data.now_val;
+    quorum_threshold_at_stage =
+      { last_updated_period = 0n
+      ; quorum_threshold = to_unsigned(quorum_threshold)
+      };
+    period_staked =
+      { last_updated_period = 0n
+          // Staked tokens are tracked for a pair of proposal/voting periods.
+          // So should only contain period no of a proposing period.
```
pasqu4le, line 78: The comment seems to contradict the value, since `0n` is a `voting` stage.

However I think the comment is incorrect, while it is true that we want to track only per-cycle (proposal + voting stage) we probably want to allow for both odd and even number to be stored here, just for redundancy (e.g. what happens if nobody calls any entrypoint in a `proposing` stage?)

src/types.mligo
```diff
@@ -146,6 +152,16 @@ type governance_token =
...
 
+type quorum_threshold_at_stage =
+  { quorum_threshold : quorum_threshold
+  ; last_updated_period : nat
+  }
+
+type period_staked =
+  { last_updated_period : nat
+  ; staked : nat
+  }
```
pasqu4le, line 163: Can we not track these together?

AFAIU the quorum changes only per-cycle, so going from a `voting` stage to a `proposing` one, but not when going from a `proposing` to a `voting` one; and what is staked also needs to be tracked in a similar way, right?

If that's the case I propose using:
```suggestion
type quorum_threshold_at_stage =
  { quorum_threshold : quorum_threshold
  ; last_updated_stage : nat
  // ^ last stage number in which the quorum was updated
  ; total_staked_cycle : nat
  // ^ total number of tokens staked in the cycle (proposing + voting stage)
  // that the last_updated_stage is part of
  }
```

src/proposal.mligo
```diff
@@ -356,7 +439,7 @@ let handle_proposal_is_over
       && counter.current < counter.total // not finished
     then
       let counter = { counter with current = counter.current + 1n } in
-      let cond =    do_total_vote_meet_quorum_threshold(proposal, store, config.quorum_threshold)
+      let cond =    do_total_vote_meet_quorum_threshold(proposal, store, store.quorum_threshold_at_stage.quorum_threshold)
                 && proposal.upvotes > proposal.downvotes
```
pasqu4le, line 443: This function (`handle_proposal_is_over`) is used by `flush`, but I don't see `update_quorum` be called in the latter, should't it be?

src/proposal.mligo
```diff
@@ -341,6 +365,65 @@ let delete_proposal
...
+              | None -> (failwith("BAD_UNSIGNED_CONVERSION"):nat)
+    }
+
+let update_quorum(store, config : storage * config): storage =
+  let current_period = get_current_period_num(store.start_time, config.voting_period) in
+  if store.quorum_threshold_at_stage.last_updated_period = current_period
+    then store // Quorum has been updated in this period, so no change is required.
+    else
+      match is_nat(current_period - store.period_staked.last_updated_period) with
+        | Some elapsed -> if elapsed >= 2n
```
pasqu4le, line 389: As I was saying I suggest not relying on the `last_updated_period` to always be a proposing stage, instead I think it is more robust to allow `voting` stages as well and here do:
- if `elapsed >= 2n` or `elapsed = 1n && is_proposing_stage(current_period)` then update the `quorum_threshold_at_stage`
- otherwise if `elapsed = 1n` only update the `last_updated_period`

What do you think?

src/proposal.mligo
```diff
@@ -341,6 +365,65 @@ let delete_proposal
...
+                fmul(old_quorum, one_plus_max_change_percent) in
+
+              let config_min_qt = to_signed(config.min_quorum_threshold) in
+              let config_max_qt = to_signed(config.max_quorum_threshold) in
+              let new_quorum = bound_qt(bound_qt(possible_new_quorum, min_new_quorum, max_new_quorum), config_min_qt, config_max_qt)
+              in { store with quorum_threshold_at_stage =
+                   { quorum_threshold = to_unsigned(new_quorum)
+                   ; last_updated_period = current_period
+                   }
+                 }
```
pasqu4le, line 412: As an argument in favor of allowing both `voting` and `proposing` stage numbers, here what happens if nobody calls this function for the whole length of a `proposing` stage?

AFAIU in the following period the `elapsed >= 2` check will pass and here we'll store a `voting` stage number.

src/proposal.mligo
```diff
@@ -146,6 +170,14 @@ let freeze_on_ledger (tokens, addr, ledger, total_supply, frozen_token_id, gover
 
 let stake_tk(token_amount, addr, voting_period, store : nat * address * voting_period * storage): storage =
   let current_period = get_current_period_num(store.start_time, voting_period) in
+  let new_period_staked = match is_nat(current_period - store.period_staked.last_updated_period) with
+    | Some elapsed -> if elapsed >= 2n //TODO@SRAS Add test to check staked tokens in a proposing period and voting period after it are both accounted towards the period no of the proposing period
```
pasqu4le, line 174: As I was suggesting, storing the staked total in the `quorum_threshold_at_stage` might make this easier (not the test, but the logic).

We'll need to call `update_quorum` in the beginning of the `propose` and `vote` entrypoint, but then we can just add to the staked total after that.

src/defaults.mligo
```diff
@@ -13,18 +13,23 @@ let validate_default_config (data : initial_config_data) : unit =
...
+  max_quorum_threshold = data.max_quorum;
+  min_quorum_threshold = data.min_quorum;
+  max_quorum_change = data.max_quorum_change;
+  quorum_change = data.quorum_change;
+  governance_total_supply = data.governance_total_supply;
+  proposal_flush_time = data.proposal_flush_time;
+  proposal_expired_time = data.proposal_expired_time;
+
+  custom_entrypoints = (Big_map.empty : custom_entrypoints);
+}
```
pasqu4le, line 32: Nitpick, probably this is due to the rebase, but this seemed better with the previous indentation.

Makefile
```diff
@@ -74,8 +74,13 @@ $(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
 $(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
-$(OUT)/trivialDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
+$(OUT)/trivialDAO_storage.tz : quorum_threshold = { numerator = quorum_denominator/10n }
```
pasqu4le, line 77: This way it's going to be quite hard to provide a custom value here without knowing about implementation details like `quorum_threshold`, perhaps we should give an easier way to give these fractions?

Maybe only accept a `nat` for a percentage here and do the math in the LIGO code?

docs/building.md
```diff
@@ -53,6 +53,11 @@ make out/trivialDAO_storage.tz \
   fixed_proposal_fee_in_token=0n \
   ledger=([] : ledger_list) \
   quorum_threshold={numerator=1n; denominator=10n} \
+  min_quorum = {numerator = 1n; denominator = 100n} \
```
pasqu4le, line 56: related: these are no longer valid values, nor the default ones.

src/types.mligo
sras: >Can we not track these together?

@pasqu4le But then we will only have a single `last_updated_period`, right? But we need to track the last updated period for staked tokens and quorum threshold separately. 

Because for updating staked count, we should check if it was updated in older an older cycle (proposing + voting periods, and if yest, we ll have to overwrite what ever the current value was, or else we need to add to the current value.

Updating quorum_threhold should come in the next cycle. So if we use the last_updated_period, then we will only see the last period where propose or vote operations were called.

src/proposal.mligo
sras: Right now it is only called during `propose` call.  Do you think that might not be enough. May be I have missed something,

src/proposal.mligo
sras: I have changed them to count the cycles, calculated as `cycle = (period + 1) /2`

So period `0` should be cycle `0`, so it should be a valid value for initilization. Then both staked tokens and quorum threhold are tracked for cycles. In effect AFAIU it amounts to the same calculation that you do here.

src/proposal.mligo
sras: >AFAIU in the following period the elapsed >= 2 check will pass and here we'll store a voting stage number.

Basically since we call update_quorum in `propose` operation, I was counting on it not happening in a voting period. But now I have changed the whole thing to use cycles anyway.

src/proposal.mligo
sras: >and vote entrypoint,

Is calling `update_qourum` in `vote` required? given than `vote` cannot happen without a `propose` operation that preceds it, and we update the quorum for the whole cycle there?

src/types.mligo
pasqu4le: > we need to track the last updated period for staked tokens and quorum threshold separately.

Maybe I'm missing something, but I'm not sure we need to, this looks like the same situation as [the one we discussed before](https://github.com/tqtezos/baseDAO/pull/225#discussion_r628260543) for `frozen_total_supply`/`freeze`, except now it's for `staked_total_supply`/staking.

What I propose is that we update this value in the storage every time we `propose`, `vote` (because they update the staked total) and `flush` (because that's where the quorum is needed).

Say in the `last_updated_stage` we can have both `voting` and `proposing` numbers, then as before all we need to calculate the next `quorum_threshold` are the `total_staked_cycle` in the current stage.

The logic I'm proposing for `update_quorum` is:
- if we go from a `voting` stage to the `proposing` one following we do:
   1. update the `last_updated_stage` to the current one
   2. use `total_staked_cycle` to calculate and update the `quorum_threshold`
   3. set `total_staked_cycle` to `0`
- if we go from a `proposing` stage to the `voting` following (so in the same cycle) we do:
   1. update the `last_updated_stage` to the current one
   2. keep `total_staked_cycle` and `quorum_threshold` the same
 
Note that the advantage is that after doing this we can:
- in `vote` and `propose` just add to the `total_staked_cycle` (as we know it's for the current cycle)
- in `flush` use the `quorum_threshold`, as we know it's updated to the current stage/cycle already

src/proposal.mligo
pasqu4le: Yeah I think `flush` needs update the quorum as well, because there may be cases in which a `propose` wasn't called yet.

Take the case in which one `propose`s in a stage, `vote`s in the next and waits for this stage to end.
It's possible that now we are again in a `proposing` period and the proposal above can be `flush`ed.
If `flush` is called before `propose` than the quorum would be outdated and remain the one for the previous cycle.

src/types.mligo
pasqu4le: I saw now the rest of the comments about `vote` not needing this call, I think you are right in that we can store the `last_updated_cycle` instead of the `last_updated_stage`.

However the point above remains, I think the total staked in the cycle can (and should) be stored together.

src/proposal.mligo
pasqu4le: Oh I see, that's a good point about `vote`, we indeed don't need to check for an update there as it cannot happen in a different cycle than the last `propose` call.

src/proposal.mligo
pasqu4le: Just repeating: I agree that `update_quorum` is not needed in `vote`, but I still think we can store the total staked alongside the quorum.

In which IIUC we still don't need to check the period here in `stake_tk` and we can add to the `total_staked_cycle` that I'm proposing.

src/proposal.mligo
sras: > If flush is called before propose than the quorum would be outdated and remain the one for the previous cycle.

AFAIU the quorum is not outdated in this particular case. Because I think it should use the old quorum itself. Imagine `x` is the proposal that has not been flushed in the next  proposing period. When it is flushed, if we recompute the quorum, then it might be possible that the participation it recived has increased the quorum, and now the votes it recieved is no longer enough to meet the quorum. 

So basically, I think the quorum at the cycle in which the proposal was raised need to be used to evaluate it.

But I think this raise a case to actually store the quorum in the proposals itself. Or else in your example, if someone raise a proposal before `x` is flushed, then also the new quorum will come into play, and might stop `x` from getting passed.

Makefile
sras: Have updated to accept percentages in `nat` for these values.

src/defaults.mligo
sras: Fixed.

src/proposal.mligo
sras: Have updated.

src/proposal.mligo
sras: Have changed the proposals to include current quorum threhsolds and use them in quorum threshold checks.

src/defaults.mligo
sras: Have changed it to contain `1n` by default so that the inital quorum threshold won't get overwritten in cycle 1, when the first proposal is raised.

src/proposal.mligo
```diff
@@ -402,6 +486,7 @@ let flush(n, config, store : nat * config * storage): return =
 let drop_proposal (proposal_key, config, store : proposal_key * config * storage): return =
   let proposal = check_if_proposal_exist (proposal_key, store) in
   let proposal_is_expired = is_time_reached (proposal, config.proposal_expired_time) in
+  let store = update_quorum(store, config) in
```
pasqu4le, line 489: I think this is no longer necessary?

If I'm not mistaken the quorum threshold is not used here, nor the staked total updated.

src/proposal.mligo
```diff
@@ -341,6 +363,68 @@ let delete_proposal
...
+            else store
+
+let propose (param, config, store : propose_params * config * storage): return =
+  let store = check_is_proposal_valid (config, param, store) in
+  let store = check_proposal_limit_reached (config, store) in
+  let amount_to_freeze = param.frozen_token + config.fixed_proposal_fee_in_token in
+  let qt = store.quorum_threshold_at_cycle.quorum_threshold in
+    // We store the current quorum threshold before it gets updated in the following update_quorum call,
+    // and use the current quorum to initiate the proposal object.
+  let store = update_quorum(store, config) in
```
pasqu4le, line 421: Why did you add this, shouldn't the quorum be updated to the new cycle here exactly so that the new value is used?

Beside even if this wasn't the case this would work only for the first `propose` called in a cycle, not the following ones.

src/proposal.mligo
```diff
@@ -341,6 +363,68 @@ let delete_proposal
...
+      if current_cycle > store.quorum_threshold_at_cycle.last_updated_cycle
+          then
+              let previous_staked = store.quorum_threshold_at_cycle.staked in
+              let previous_participation = to_signed(fraction_to_quorum_fraction(previous_staked, config.governance_total_supply)) in
+              let old_quorum = to_signed(store.quorum_threshold_at_cycle.quorum_threshold) in
+              let quorum_change = to_signed(config.quorum_change) in
+              let possible_new_quorum =
+                // old_quorum + (previous_participation - old_quorum) * quorum_change
+                fadd(old_quorum, fmul(quorum_change, fsub(previous_participation, old_quorum))) in
+                //old_quorum + fmul(quorum_change, previous_participation - old_quorum) in
```
pasqu4le, line 396: This line looks like it was left behind by accident.

sras: Removed

src/proposal.mligo
sras: What a stupid mixup! Have fixed it now and added a test.

src/proposal.mligo
sras: Removed it.

haskell/test/Test/Ligo/BaseDAO/Integrational/Proposal.hs
```diff
@@ -32,13 +32,21 @@ test_IntegrationalBaseDAOProposal = testGroup "BaseDAO Integrational Tests"
   [ testGroup "Integrational Tests for checking storage"
       [ testCase "contract correctly tracks freeze-unfreeze history" $
           integrationalTestExpectation $ checkFreezeHistoryTracking
+      , testCase "contract correctly updates quorum threshold and rests outdated staked token count" $
```
pasqu4le, line 35:
```suggestion
      , testCase "contract correctly updates quorum threshold and resets outdated staked token count" $
```
looks like a typo

haskell/test/Test/Ligo/BaseDAO/Integrational/Proposal.hs
```diff
@@ -93,3 +104,196 @@ checkFreezeHistoryTracking  = do
...
+        let proposalMeta2 = "A"
+        tTransfer  (#from .! wallet1) (#to .! (unTAddress baseDao)) zeroMutez (unsafeBuildEpName "propose") (toVal (ProposeParams requiredFrozen proposalMeta2))
+
+        lExpectStorage @FullStorage baseDao $ \storage -> do
+            let qs = sQuorumThresholdAtCycle $ fsStorage storage
+            -- We start with quorumThreshold of 3/10, with participation as 10 and governance total supply as 100
+            -- participation = 10/100 = 1/10
+            -- so possible_new_quorum = 3/10 * (1 - 0.05)  + (1/10 * 5/100)
+            -- so 3/10 * (19/20) + (5/1000) = 57/200 + 1/200 = 58/200 = 29/100
+            let expected = QuorumThresholdAtCycle 2 (mkQuorumThreshold 290 1000) 10
```
pasqu4le, line 156: here (and in following places) these tests rely on these calculations to be correct. The comments are good, however this makes updating the tests on small changes quite complex, wouldn't it be possible to make a function to calculate the expected new threshold from some given arguments?

pasqu4le (approved):
Two comments left, the rest LGTM

haskell/test/Test/Ligo/BaseDAO/Integrational/Proposal.hs
sras: Have updated.

pasqu4le (approved):
LGTM now, remember that you need to squash and sign the commit to merge this.

Issue #215, opened 2021-05-02
Allow the DAO to automatically accept `admin` privileges of itself


pasqu4le:
# Clarification and motivation

We currently have a two-step process for the transfer of the `admin` role:
1. `admin` calls `transfer_ownership` to create a new `admin` candidate
2. the `admin` candidate calls `accept_ownership` to get the role

This is a valuable security feature to ensure that the `admin` is not incorrectly modified to an incorrect address and to check that the new `admin` can call the contract.

However, since this is a DAO, it is supposed to be governed only by its users after an initial period, so we should be able to make the DAO accept the `admin` role of itself.

To do that I suggest modifying the `transfer_ownership` entrypoint so that if the `admin` candidate is the DAO contract itself no further calls are needed.

# Acceptance criteria

The DAO is able to automatically accept ownership of itself.

sras:
Resolved by https://github.com/tqtezos/baseDAO/pull/226

PR #226, opened 2021-05-07
[#215] Accept admin privilege without accept_owenership if new owner if contract itself

sras:
Problem: In this issue, we are required to modify the behavior of
transfer_ownership entrypoint such that the new admin is set without
having to call accept_ownership call, if the new owner is the address of
the contract itsef.

Soltution: Implement the behavior and update tests and docs.

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/management.mligo
+++ b/src/management.mligo
@@ -13,7 +13,12 @@
 let transfer_ownership
     (param, store : transfer_ownership_param * storage) : return =
       let store = authorize_admin(store) in
-        (([] : operation list), { store with pending_owner = param ; })
+      let store =
+        if Tezos.self_address = param
+          then { store with admin = param ; } // If new admin is address of baseDAO,
+                                              // set as admin right away.
+          else { store with pending_owner = param ; }
+      in (([] : operation list), store)
 
 (*
  * Auth check for pending admin and copies the value in 'pending_owner' field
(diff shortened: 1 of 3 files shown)


rinn7e (approved):
Seems fine for me!

src/management.mligo
```diff
@@ -13,7 +13,9 @@
 let transfer_ownership
     (param, store : transfer_ownership_param * storage) : return =
       let store = authorize_admin(store) in
-        (([] : operation list), { store with pending_owner = param ; })
+        if Tezos.self_address = param
+          then (([] : operation list), { store with admin = param ; }) // If new admin is address of baseDAO, set as admin right away.
+          else (([] : operation list), { store with pending_owner = param ; })
```
pasqu4le, line 18: A bit of a nitpick, but it should be slightly more efficient to use the ops list once, also the comment takes the line too far out:
```
      let store =
            if Tezos.self_address = param
            then
               // If new admin is address of baseDAO, set as admin right away.
              { store with admin = param ; }
            else
              { store with pending_owner = param ; }
      in (nil_op, store)
```

pasqu4le (commented):
Note: there is a typo in the commit message: `Soltution`

Also as a reminder: this should me merged only after all the other `p0` issues

src/management.mligo
sras: Have updated.

PR #227, opened 2021-05-11
[#87] Improve handling of failed proposals

### LINKED ISSUE #87: Improve handling of failed proposals

This task follows #69. The broader problem is described [here](https://gist.github.com/gromakovsky/cd03f73f1f5dcc9640def18d0eee642d) along with a proposed solution. We are going to discuss it, reach consensus about the best solution and implement it.

# Acceptance criteria

Once there is a consensus for the problem described in the aforementioned gist document, we should implement the solution we agree upon.

rinn7e:
## Description

Problem: This task follows #69. The broader problem is described [here](https://gist.github.com/gromakovsky/cd03f73f1f5dcc9640def18d0eee642d) along with a proposed solution. We are going to discuss it, reach consensus about the best solution and implement it.

Solution: The solution follows what has been described in here: https://gist.github.com/gromakovsky/cd03f73f1f5dcc9640def18d0eee642d#gistcomment-3733825

Basically, the current changes include:
1. `flush` is able to skip proposals, and targets specific ones, so to avoid the ones that are currently failing

2. `flush` has a built-in expiry: a proposal `flush`ed after a certain period (keeping with the previous example `x + 3` for proposals created in `x`, for the same reason as explained above) is dropped.

3. `dropProposal` is removed entirely, as its duty is covered by `flush` itself now.

## Related issue(s)

Resolves #87

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -276,7 +276,6 @@ Full list:
 * [`propose`](#propose)
 * [`vote`](#vote)
 * [`flush`](#flush)
-* [`drop_proposal`](#drop_proposal)
 * [`freeze`](#freeze)
 * [`unfreeze`](#unfreeze)
 
@@ -668,28 +667,6 @@ have the same timestamp due to being in the same block, are processed in the ord
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
 - The lost of frozen tokens is due to the fact that the administrator has the right to `transfer` frozen tokens of any proposers or voters.
 - If proposal is accepted, decision lambda is called.
-
-### **drop_proposal**
-
-```ocaml
-type proposal_key = bytes
-
-Drop_proposal of proposal_key
-```
-
-Parameter (in Michelson):
-```
-(bytes %drop_proposal)
-```
-
-- Delete a finished and accepted proposal that is not flushed. Tokens frozen for this
-proposal are returned to the proposer and voters in full. The decision lambda is skipped.
-- This entrypoint should only be used when there is a proposal that is stuck due to having a
-failing decision lambda.
-- Fails with `NOT_ADMIN` if the sender is not the administrator.
-
-
-[FA2]: https://gitlab.com/tzip/tzip/-/blob/3a6464b1e641008b77a83807a0c102e7602c6af4/proposals/tzip-12/tzip-12.md
 
 ## Custom entrypoints
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -645,27 +645,27 @@ Parameter (in Michelson):
 ### **flush**
 
 ```ocaml
-Flush of nat
+type proposal_key = bytes
+
+Flush of [proposal_key]
 ```
 
 Parameter (in Michelson):
 ```
-(nat %flush)
+(bytes list %flush)
 ```
 
-- Finish voting process on an amount of proposals for which the voting period is over.
-- The order of processing proposals are from 'the oldest' to 'the newest'. The proposals which
-have the same timestamp due to being in the same block, are processed in the order of their proposal keys.
+- Finish voting process on proposals that are in the parameters for which the voting period is over.
 - Frozen tokens from voters and proposal submitter associated with those proposals are returned
   in form of tokens in governance token contract:
-  - If proposal got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
+  - If proposal became expired or got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
     - The return amount for the proposer is equal to or less than the slashed amount based on `rejectedProposalReturnValue`.
     - The paid fee is not returned to the proposer.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
   - If proposal got accepted:
     - The return amount for the proposer is equal to or less than the sum of the proposer frozen tokens and the fee paid for the proposal.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
-- The lost of frozen tokens is due to the fact that the administrator has the right to `transfer` frozen tokens of any proposers or voters.
+- The lost of frozen tokens is due to the fact that there can be proposals that `transfer` frozen tokens of any proposers or voters.
 - If proposal is accepted, decision lambda is called.
 
 ## Custom entrypoints
--- a/src/proposal.mligo
+++ b/src/proposal.mligo
@@ -337,27 +337,31 @@ let delete_proposal
     Set.remove (start_date, proposal_key) store.proposal_key_list_sort_by_date
   }
 
-// if proposal is proposed at 0 period_num, it will be expired at 3 period_num
+// If proposal is proposed at `period_num 1`, it will be expired starting at `period_num 5`.
+// Note: If a proposal is propose at `x`:
+// - `x + 1` it will be voted on
+// - `x + 2` it can be flushed, new proposals will be created in this stage but these cannot be flushed yet
+// - `x + 3` it can be flushed still, the proposals from `x + 2` are being voted on, so they still can't be flushed yet
+// - `x + 4` is too late: proposals from a newer period `x + 2` can not be flushed too, so the ones from `x` are expired
 [@inline]
-let is_proposal_expired (proposal, voting_period, store : proposal * voting_period * storage): bool =
-  let current_period = get_current_period_num(store.start_time, voting_period) in
-  current_period > proposal.period_num + 2n
+let is_proposal_expired (proposal, current_period : proposal * nat): bool =
+  current_period > proposal.period_num + 3n
 
 [@inline]
 let handle_proposal_is_over
-    (config, start_date, proposal_key, store, ops
-      : config * timestamp * proposal_key * storage * operation list
+    (config, proposal_key, store, ops
+      : config * proposal_key * storage * operation list
     )
-    : (operation list * storage * bool) =
+    : (operation list * storage) =
   let proposal = check_if_proposal_exist (proposal_key, store) in
+  let current_period = get_current_period_num (store.start_time, config.voting_period) in
   if    is_voting_period_over (proposal, config.voting_period, store)
   then
     let cond =
-          if (is_proposal_expired(proposal, config.voting_period, store))
-          then false // always fails when the proposal is expired regardless of actual votes
-          else
-            do_total_vote_meet_quorum_threshold(proposal, store, config.quorum_threshold)
-               && proposal.upvotes > proposal.downvotes
+          // always fails when the proposal is expired regardless of actual votes
+          not (is_proposal_expired(proposal, current_period))
+          && do_total_vote_meet_quorum_threshold(proposal, store, config.quorum_threshold)
+          && proposal.upvotes > proposal.downvotes
     in
     let store = unfreeze_proposer_and_voter_token
           (config.rejected_proposal_return_value, cond, proposal, config.voting_period, config.fixed_proposal_fee_in_token, store) in
@@ -370,9 +374,9 @@ let handle_proposal_is_over
     in
     let cons = fun (l, e : operation list * operation) -> e :: l in
     let ops = List.fold cons ops new_ops in
-    let store = delete_proposal (start_date, proposal_key, store) in
-    (ops, store, true)
-  else (ops, store, false)
+    let store = delete_proposal (proposal.start_date, proposal_key, store) in
+    (ops, store)
+  else (ops, store)
 
 // Elem function for list of `proposal_key`
 [@inline]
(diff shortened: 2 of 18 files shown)


rinn7e:
@pasqu4le This PR is still in progress, basically the things that are done:
- `flush` now can skip and target specific proposals. I also keep the `old flush with an amount` feature
- `flush` now takes into account whether a proposal is expired or not. If it is expired, it treats the proposal as if it is a failed proposal regardless of the actual votes.
- `dropProposal` is now removed.

The next stage that I'm doing is:
- optimizing the changes that have been made into `flush`
- add tests to cover all the new cases

src/proposal.mligo
```diff
@@ -363,50 +371,79 @@ let handle_proposal_is_over
...
-        handle_proposal_is_over (config, start_date, proposal_key, store, ops, counter)
-      in
-  let (ops, store, counter) =
-    Set.fold flush_one store.proposal_key_list_sort_by_date (([] : operation list), store, counter) in
-  (ops, store)
-
-// Removes an accepted and finished proposal by key.
-let drop_proposal (proposal_key, config, store : proposal_key * config * storage): return =
-  let store = authorize_admin store in
+// Flush all proposals that passed their voting period.
```
sras, line 384: Shouldn't this be amended?

src/proposal.mligo
```diff
@@ -363,50 +371,79 @@ let handle_proposal_is_over
...
+
+        let counter : counter =
+          { current = 0n
+          ; total = n
+          } in
+        let flush_one
+            (acc, e: (operation list * storage * counter) * (timestamp * proposal_key)) =
+              let (ops, store, counter) = acc in
+              let (start_date, proposal_key) = e in
+              if (counter.current < counter.total)
```
sras, line 403: Wonder if we can just use an integer counter, and just check if it greater than zero at every itereration, so that we don't have to pass `total` into this function and also can save the `counter` type.

src/proposal.mligo
```diff
@@ -363,50 +371,79 @@ let handle_proposal_is_over
...
+              then
+                let (ops, store, is_flush) = handle_proposal_is_over (config, start_date, proposal_key, store, ops) in
+                if (is_flush = true)
```
sras, line 406: Wouldn't `if (is_flush)` work?

src/types.mligo
```diff
@@ -206,17 +206,19 @@ type transfer_contract_tokens_param =
   ; params : transfer_params
   }
 
-
+type flush_param =
+    Flush_amount of nat // old behavior
+  | Flush_target of proposal_key list // target specific proposals
+  | Flush_skip of proposal_key list // skip specific proposals
```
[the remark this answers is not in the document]
pasqu4le: @rinn7e note that if you do this updating the tests will be much simpler.

src/proposal.mligo
```diff
@@ -337,18 +337,26 @@ let delete_proposal
     Set.remove (start_date, proposal_key) store.proposal_key_list_sort_by_date
   }
 
+// if proposal is proposed at 0 period_num, it will be expired at 3 period_num
+[@inline]
+let is_proposal_expired (proposal, voting_period, store : proposal * voting_period * storage): bool =
+  let current_period = get_current_period_num(store.start_time, voting_period) in
+  current_period > proposal.period_num + 2n
```
pasqu4le, line 344: Ah my explanation wasn't very clear, when I said:
> a proposal flushed after a certain period (keeping with the previous example `x + 3` for proposals created in `x`, for the same reason as explained above) is dropped.

I meant any period _after_ `x + 3` included, so `x + 4` or later.

My reasoning is that if a proposal is created at `x` then in
- `x + 1` it will be voted on
- `x + 2` it can be flushed, new proposals will be created in this stage but these cannot be flushed yet
- `x + 3` it can be flushed still, the proposals from `x + 2` are being voted on, so they still can't be flushed yet
- `x + 4` is too late: proposals from a newer period `x + 2` can not be flushed too, so the ones from `x` are expired

src/types.mligo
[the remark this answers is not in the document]
pasqu4le: > I believe this is what Flush_target achieve currently.

Yep, I propose to only keep that as it's the most comfortable to use (while keeping all the functionality needed) IMO.

> I agree that this hurt efficiency, so I'm good with removing them.
Cool :+1:

src/proposal.mligo
```diff
@@ -337,18 +337,26 @@ let delete_proposal
...
-     && counter.current < counter.total // not finished
   then
-    let counter = { counter with current = counter.current + 1n } in
-    let cond =    do_total_vote_meet_quorum_threshold(proposal, store, config.quorum_threshold)
+    let cond =
+          if (is_proposal_expired(proposal, config.voting_period, store))
+          then false // always fails when the proposal is expired regardless of actual votes
+          else
+            do_total_vote_meet_quorum_threshold(proposal, store, config.quorum_threshold)
                && proposal.upvotes > proposal.downvotes
```
pasqu4le, line 360:
```suggestion
    let cond =
          // always fails when the proposal is expired regardless of actual votes
          not (is_proposal_expired(proposal, config.voting_period, store))
          && do_total_vote_meet_quorum_threshold(proposal, store, config.quorum_threshold)
          && proposal.upvotes > proposal.downvotes
```

src/proposal.mligo
```diff
@@ -337,18 +337,26 @@ let delete_proposal
...
 let handle_proposal_is_over
-    (config, start_date, proposal_key, store, ops, counter
-      : config * timestamp * proposal_key * storage * operation list * counter
+    (config, start_date, proposal_key, store, ops
+      : config * timestamp * proposal_key * storage * operation list
     )
-    : (operation list * storage * counter) =
+    : (operation list * storage * bool) =
   let proposal = check_if_proposal_exist (proposal_key, store) in
   if    is_voting_period_over (proposal, config.voting_period, store)
```
pasqu4le, line 353: We now calculate the `current_period` twice, in `is_voting_period_over` as well as `is_proposal_expired`, let's calculate it once here and pass it as an argument instead:
```suggestion
  let current_period = get_current_period_num(store.start_time, config.voting_period) in
  if    is_voting_period_over (proposal, current_period)
```

src/proposal.mligo
pasqu4le: Likewise the specification for the entrypoint should be updated

rinn7e: Update the spec now, not sure if that's good enough.

src/proposal.mligo
```diff
@@ -363,50 +371,79 @@ let handle_proposal_is_over
...
-let drop_proposal (proposal_key, config, store : proposal_key * config * storage): return =
-  let store = authorize_admin store in
+// Flush all proposals that passed their voting period.
+let flush(param, config, store : flush_param * config * storage): return =
+  match param with
+    | Flush_amount n ->
+        let store =
+          if n = 0n
+            then
+              (failwith("BAD_ENTRYPOINT_PARAMETER") : storage)
```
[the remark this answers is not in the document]
rinn7e: I think this is simply a guard. We don't expect the param to be 0.

src/proposal.mligo
```diff
@@ -363,50 +371,79 @@ let handle_proposal_is_over
...
+            (acc, e: (operation list * storage) * (timestamp * proposal_key)) =
+              let (ops, store) = acc in
+              let (start_date, proposal_key) = e in
+              if elem_proposal_key_list (proposal_key, keys) = true then
+                let (ops, store, is_flush) = handle_proposal_is_over (config, start_date, proposal_key, store, ops)
+                in (ops, store)
+              else (ops, store)
+            in
+        let (ops, store) =
+          Set.fold flush_one store.proposal_key_list_sort_by_date (([] : operation list), store) in
```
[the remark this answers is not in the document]
rinn7e: With the new implementation, this makes sense. I've updated this now.

src/types.mligo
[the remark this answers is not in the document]
rinn7e: I push the changes according to feedback now, you can check this @pasqu4le .

src/proposal.mligo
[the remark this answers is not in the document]
sras: But the guard is not required, because zero input is a no-op right? So AFAIU there is nothing here to guard against.

src/proposal.mligo
```diff
@@ -337,19 +337,31 @@ let delete_proposal
     Set.remove (start_date, proposal_key) store.proposal_key_list_sort_by_date
   }
 
+// If proposal is proposed at `period_num 1`, it will be expired starting at `period_num 5`.
+// Note: If a proposal is propose at `x`:
+// - `x + 1` it will be voted on
+// - `x + 2` it can be flushed, new proposals will be created in this stage but these cannot be flushed yet
+// - `x + 3` it can be flushed still, the proposals from `x + 2` are being voted on, so they still can't be flushed yet
+// - `x + 4` is too late: proposals from a newer period `x + 2` can not be flushed too, so the ones from `x` are expired
```
pasqu4le, line 345: typo (likely mine)
```suggestion
// - `x + 4` is too late: proposals from a newer period `x + 2` can now be flushed too, so the ones from `x` are expired
```

src/proposal.mligo
```diff
@@ -362,51 +374,36 @@ let handle_proposal_is_over
...
+  else (ops, store)
 
-// Flush all proposals that passed their voting period.
-let flush(n, config, store : nat * config * storage): return =
+// Elem function for list of `proposal_key`
+[@inline]
+let elem_proposal_key_list (l, ls : proposal_key * proposal_key list) : bool =
+  let check_equal (acc, e : bool * proposal_key) =
+        if e = l then true else false
+  in List.fold check_equal ls false
```
pasqu4le, line 386: This looks like a reminder that can be now removed.

src/proposal.mligo
```diff
@@ -362,51 +374,36 @@ let handle_proposal_is_over
...
+
+// Flush proposals that are provided in the parameter. Other proposals are not affected.
+let flush(keys, config, store : proposal_key list * config * storage): return =
   let store =
-    if n = 0n
+    if List.length keys = 0n
       then
         (failwith("BAD_ENTRYPOINT_PARAMETER") : storage)
       else store
```
pasqu4le, line 394: I think there is no need to fail here and we can just let `List.fold` handle the empty case as well.

docs/specification.md
```diff
@@ -646,51 +645,29 @@ Parameter (in Michelson):
...
-  - If proposal got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
+  - If proposal became expired or got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
     - The return amount for the proposer is equal to or less than the slashed amount based on `rejectedProposalReturnValue`.
     - The paid fee is not returned to the proposer.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
   - If proposal got accepted:
     - The return amount for the proposer is equal to or less than the sum of the proposer frozen tokens and the fee paid for the proposal.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
-- The lost of frozen tokens is due to the fact that the administrator has the right to `transfer` frozen tokens of any proposers or voters.
+- The lost of frozen tokens is due to the fact that there can be proposals that `transfer` frozen tokens of any proposers or voters.
```
pasqu4le, line 668:
```suggestion
- The loss of frozen tokens is due to the fact that there can be proposals that `transfer` frozen tokens of any proposers or voters.
```

docs/specification.md
```diff
@@ -646,51 +645,29 @@ Parameter (in Michelson):
...
 ```
 
-- Finish voting process on an amount of proposals for which the voting period is over.
-- The order of processing proposals are from 'the oldest' to 'the newest'. The proposals which
-have the same timestamp due to being in the same block, are processed in the order of their proposal keys.
+- Finish voting process on proposals that are in the parameters for which the voting period is over.
 - Frozen tokens from voters and proposal submitter associated with those proposals are returned
   in form of tokens in governance token contract:
-  - If proposal got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
+  - If proposal became expired or got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
```
pasqu4le, line 661:
```suggestion
  - If the proposal expired or got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
```

I would also add here that a proposal is expired when more than 3 stages have passed since the one where it was created.

src/proposal.mligo
```diff
@@ -337,19 +337,31 @@ let delete_proposal
     Set.remove (start_date, proposal_key) store.proposal_key_list_sort_by_date
   }
 
+// If proposal is proposed at `period_num 1`, it will be expired starting at `period_num 5`.
+// Note: If a proposal is propose at `x`:
```
pasqu4le, line 341:
```suggestion
// Note: If a proposal is proposed at `x`:
```

src/proposal.mligo
pasqu4le: Yep, that's correct.

pasqu4le:
@rinn7e there's also this part of a previous comment:

> we have the `proposal_key_list_sort_by_date` accessible from the contract storage and I suggest we add a TZIP-16 view for convenience.

pasqu4le:
Closing this in favor of a [new logic for `drop_proposal`](https://github.com/tqtezos/baseDAO/issues/87#issuecomment-839716850) that was agreed as solution for the issue.

PR #228, opened 2021-05-11
Fix stages cycle to match the specification

pasqu4le:
## Description

Problem: there is a discrepancy between the specifiction and the
implementation (and tests) for the cycles: in the spec we say that
proposals stages are only odd and voting ones even, but the contract
behaves in the opposite way.

Solution: it seems to be more sensible to follow the specification,
although the current implementation is just as valid of an option, so
the contract code and tests are updated.
Additionally, tests have been commented more to show this more clearly.

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -182,8 +182,14 @@ The contract constantly cycles between two stages, a proposing stage and a votin
 Both have the same same length, `voting_period` and alternate between each other,
 starting from "voting" for period number `0`.
 
+The voting period is specified for the whole smart contract and never changes.
+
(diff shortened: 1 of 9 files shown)


sras:
LGTM. But I think it would be nice to include the rationale for why we are staring the contract in a voting period, in the spec doc.

sras (approved):

rinn7e (approved):
Seem fine for me!

pasqu4le:
@sras good point, added an explanation line to the spec.

PR #231, opened 2021-05-13
[#230] Fix Makefile to use proper current timestamp

### LINKED ISSUE #230: Replace `Tezos.now` in the Makefile with the current time

Currently we use `Tezos.now` to generate the contract storage, however `ligo compile-storage` is unable to derive the current time and as a result the starting time of the contract is set to `0` aka the first second of `1970`.

We should use the current time instead, calculated from the system running the command, likely we can use the `date` command for it.

# Acceptance criteria

Running `make` to produce a storage provides a start time that's as close to "now" as possible.

sras:
## Description

Problem: In the storage generation section of Makefile we use
'Tezos.now' to fetch the current timestamp. But it appears that the
'ligo compile' command is not able to fetch the current timestamp
properly.

Solution: Use the 'date' command from environment to fetch the unix
timestamp and use it as a timestamp during storage generation.

## Related issue(s)

Resolves #230 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


pasqu4le (approved):
:ok_hand: looks like it works just fine

PR #232, opened 2021-05-13
[#87] Improve handling of failed proposals (2nd Approach)

### LINKED ISSUE #87: Improve handling of failed proposals

rinn7e:
## Description

Problem: Problem: Basically, in a case when a proposal get accepted but it containsoperation that fails, we have the `drop_proposal` entrypoint to remove this proposal.This entrypoint can only be call by the `admin` currently. We no longer want this, andwe want it to be called by anyone. As a result, we need to update to the current `flush`
process, add expiry process, and update `drop_proposal` logic.

Original gist/discussion: https://gist.github.com/gromakovsky/cd03f73f1f5dcc9640def18d0eee642d

Solution: Add `proposal_flush_time_reached` and `proposal_expired_time` to
config, add `guardian` to `storage`, update `flush` and `drop_proposal`
logic.

Detail: https://github.com/tqtezos/baseDAO/issues/87#issuecomment-839716850

## Related issue(s)

Resolves #87

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ endif
 	#
 
 $(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/trivialDAO_storage.tz : guardian_address = KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh
 $(OUT)/trivialDAO_storage.tz : governance_token_address = KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5
 $(OUT)/trivialDAO_storage.tz : governance_token_id = 0n
 $(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
@@ -75,6 +76,8 @@ $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
 $(OUT)/trivialDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
 $(OUT)/trivialDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/trivialDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
+$(OUT)/trivialDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
@@ -82,6 +85,7 @@ $(OUT)/trivialDAO_storage.tz: src/**
       src/base_DAO.mligo base_DAO_contract "default_full_storage( \
         { storage_data = \
           { admin = (\"$(admin_address)\" : address) \
+          ; guardian = (\"$(guardian_address)\" : address) \
           ; governance_token = \
             { address = (\"$(governance_token_address)\" : address) \
             ; token_id = $(governance_token_id) \
@@ -92,6 +96,8 @@ $(OUT)/trivialDAO_storage.tz: src/**
           } \
         ; config_data = \
           { voting_period = { length = $(voting_period) } \
+          ; proposal_flush_time = $(proposal_flush_time) \
+          ; proposal_expired_time = $(proposal_expired_time) \
           ; quorum_threshold = $(quorum_threshold) \
           ; fixed_proposal_fee_in_token = $(fixed_proposal_fee_in_token) \
           } \
@@ -101,6 +107,7 @@ $(OUT)/trivialDAO_storage.tz: src/**
 	#
 
 $(OUT)/registryDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/registryDAO_storage.tz : guardian_address = KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh
 $(OUT)/registryDAO_storage.tz : governance_token_address = KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5
 $(OUT)/registryDAO_storage.tz : governance_token_id = 0n
 $(OUT)/registryDAO_storage.tz : frozen_scale_value = 1n
@@ -116,6 +123,8 @@ $(OUT)/registryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/registryDAO_storage.tz : ledger = ([] : ledger_list)
 $(OUT)/registryDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
 $(OUT)/registryDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/registryDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
+$(OUT)/registryDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
 $(OUT)/registryDAO_storage.tz: src/**
 	# ============== Compiling RegistryDAO storage ============== #
 	mkdir -p $(OUT)
@@ -124,6 +133,7 @@ $(OUT)/registryDAO_storage.tz: src/**
         { base_data = \
           { storage_data = \
             { admin = (\"$(admin_address)\" : address) \
+            ; guardian = (\"$(guardian_address)\" : address) \
             ; governance_token = \
               { address = (\"$(governance_token_address)\" : address) \
               ; token_id = $(governance_token_id) \
@@ -135,6 +145,8 @@ $(OUT)/registryDAO_storage.tz: src/**
           ; config_data = \
             { quorum_threshold = $(quorum_threshold) \
             ; voting_period = { length = $(voting_period) } \
+            ; proposal_flush_time = $(proposal_flush_time) \
+            ; proposal_expired_time = $(proposal_expired_time) \
             ; fixed_proposal_fee_in_token = $(fixed_proposal_fee_in_token) \
             } \
           } \
@@ -151,6 +163,7 @@ $(OUT)/registryDAO_storage.tz: src/**
 	#
 
 $(OUT)/treasuryDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
+$(OUT)/treasuryDAO_storage.tz : guardian_address = KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh
 $(OUT)/treasuryDAO_storage.tz : governance_token_address = KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5
 $(OUT)/treasuryDAO_storage.tz : governance_token_id = 0n
 $(OUT)/treasuryDAO_storage.tz : frozen_scale_value = 0n
@@ -166,6 +179,8 @@ $(OUT)/treasuryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/treasuryDAO_storage.tz : ledger = ([] : ledger_list)
 $(OUT)/treasuryDAO_storage.tz : quorum_threshold = {numerator = 1n; denominator = 10n}
 $(OUT)/treasuryDAO_storage.tz : voting_period = 950400n # 11 days
+$(OUT)/treasuryDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
+$(OUT)/treasuryDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
 $(OUT)/treasuryDAO_storage.tz: src/**
 	# ============== Compiling TreasuryDAO storage ============== #
 	mkdir -p $(OUT)
@@ -174,6 +189,7 @@ $(OUT)/treasuryDAO_storage.tz: src/**
         { base_data = \
           { storage_data = \
             { admin = (\"$(admin_address)\" : address) \
+            ; guardian = (\"$(guardian_address)\" : address) \
             ; governance_token = \
               { address = (\"$(governance_token_address)\" : address) \
               ; token_id =  $(governance_token_id) \
@@ -185,6 +201,8 @@ $(OUT)/treasuryDAO_storage.tz: src/**
           ; config_data = \
             { quorum_threshold = $(quorum_threshold) \
             ; voting_period = { length = $(voting_period) } \
+            ; proposal_flush_time = $(proposal_flush_time) \
+            ; proposal_expired_time = $(proposal_expired_time) \
             ; fixed_proposal_fee_in_token = $(fixed_proposal_fee_in_token) \
             } \
           } \
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -85,6 +85,9 @@ for more information on its content and usage.
 DAO configuration value parameters are captured by the `config` type:
 
 ```ocaml
+
+type seconds = nat
+
 type config =
   { proposal_check : propose_params * storage -> bool
   // ^ A lambda used to verify whether a proposal can be submitted.
@@ -105,6 +108,15 @@ type config =
 
   ; voting_period : voting_period
   // ^ Determine the voting_period length.
+
+  ; proposal_flush_time : seconds
+  // ^ Determine the minimum amount of seconds after the proposal is proposed to be able to be `flushed`.
+  // Have to be bigger than `voting_period * 2`
+
+  ; proposal_expired_time : seconds
+  // ^ Determine the minimum amount of seconds after the proposal is proposed to be considered as expired.
+  // Have to be bigger than `proposal_flush_time`
+
   ; quorum_threshold : quorum_threshold
   // ^ Determine the quorum threshold that the proposals need to meet.
   ; fixed_proposal_fee_in_token : nat
@@ -146,11 +158,19 @@ Some configuration values are specified in runtime and can be changed during the
 They must be provided on origination to construct the contract's initial storage.
 These values are:
 1. `admin : address` is the address that can perform administrative actions.
-2. `governance_token` is the FA2 contract address/token id pair that will be used as the governance token.
-3. `voting_period : nat` specifies how long the voting period lasts in seconds.
-4. `quorum_threshold : quorum_threshold` specifies what fraction of the frozen
+2. `guardian : address` is the address of a contract that have permission to call `drop_proposal` on any proposals.
+    - `guardian` contract cannot initiate the transaction that results in a call to `drop_proposal`
+3. `governance_token` is the FA2 contract address/token id pair that will be used as the governance token.
+4. `voting_period : nat` specifies how long the voting period lasts in seconds.
+5. `proposal_flush_time : nat`
+    - Specifies in seconds how long after the proposal is proposed it can be flushed.
+    - IMPORTANT: Must be bigger than `voting_period * 2`.
+6. `proposal_expired_time : nat`
+    - Specifies in seconds how long after the proposal is proposed it is considered expired.
+    - IMPORTANT: Must be bigger than `proposal_flush_time`.
+7. `quorum_threshold : quorum_threshold` specifies what fraction of the frozen
    tokens total supply of total votes are required for a successful proposal.
-5. `fixed_proposal_fee_in_token : nat` specifies the fee for submitting a proposal (in native DAO token).
+8. `fixed_proposal_fee_in_token : nat` specifies the fee for submitting a proposal (in native DAO token).
 
 # Contract logic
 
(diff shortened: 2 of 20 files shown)


haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -799,15 +839,31 @@ dropProposal originateFn = do
 
   key3 <- createSampleProposal 3 owner1 dao
 
-  withSender (AddressResolved admin) $ do
+  -- -- TODO: Change to `guardian`, but currently it is not working may be due to sender <> source
+  withSender (AddressResolved owner1) $ do
     call dao (Call @"Drop_proposal") key1
```
rinn7e, line 844: @pasqu4le I have added `proposal_flush_time`, and `proposal_expired_time`, update `flush` and `drop_proposal` entrypoints, and also update the tests.

One thing that I couldn't figure out is writing the test for `drop_proposal` such that the `guardian` can drop any proposals. 
- Currently, I set `guardian = admin` in the tests, and I think the check `sender <> source` cause this test to fail.
- I don't fully understand how `SOURCE` works, hope to get your advice on what the proper value should `guadian` be.

Makefile
```diff
@@ -82,6 +82,7 @@ $(OUT)/trivialDAO_storage.tz: src/**
       src/base_DAO.mligo base_DAO_contract "default_full_storage( \
         { storage_data = \
           { admin = (\"$(admin_address)\" : address) \
+          ; guardian = (\"$(admin_address)\" : address) \
```
sras, line 85: guardian is always same as admin address? The acceptance description seems to describe otherwise,

>this cannot be held by an implicit account (to be checked in drop_proposal)
>its permissions are different from the admins, it's duties are only to call drop_proposal where necessary

So it appears that we need to add another argument for the user to specify guardian.

Makefile
```diff
@@ -92,6 +93,8 @@ $(OUT)/trivialDAO_storage.tz: src/**
           } \
         ; config_data = \
           { voting_period = { length = $(voting_period) } \
+          ; proposal_flush_time = ($(voting_period) * 2n) \
+          ; proposal_expired_time = ($(voting_period) * 3n) \
```
sras, line 97: Here I am not sure if we should hard code `2n` and `3n`, and instead may be we should allow user to specify them, and also  probably check that flush time is greater then 2 * voting_period, and expiry time is greater than flush time.

docs/specification.md
```diff
@@ -105,6 +108,15 @@ type config =
 
   ; voting_period : voting_period
   // ^ Determine the voting_period length.
+
+  ; proposal_flush_time : seconds
+  // ^ Determine how long after the proposal is proposed to be able to be `flushed`.
```
sras, line 113: Probably needs a bit rephrasing for clarity. something like "minimum time that should pass after a proposal's submission before it can be flushed"..

docs/specification.md
```diff
@@ -105,6 +108,15 @@ type config =
...
+  // Have to be bigger than `voting_period * 2`
+
+  ; proposal_expired_time : seconds
+  // ^ Determine how long after the proposal is proposed to be considered as expired.
```
sras, line 117: Can use a bit of rephrasing here as well.

docs/specification.md
```diff
@@ -146,11 +158,19 @@ Some configuration values are specified in runtime and can be changed during the
 They must be provided on origination to construct the contract's initial storage.
 These values are:
 1. `admin : address` is the address that can perform administrative actions.
-2. `governance_token` is the FA2 contract address/token id pair that will be used as the governance token.
-3. `voting_period : nat` specifies how long the voting period lasts in seconds.
-4. `quorum_threshold : quorum_threshold` specifies what fraction of the frozen
+2. `guardian : address` is the address that have permission to call `drop_proposal`.
+    - `address` of guardian cannot be equal to `SOURCE`.
```
sras, line 162: I am not sure I haven't mixed up meaning of SENDER/SOURCE, but maybe a better way to put this might be something like "`guardian` cannot initiate the transaction that results in call to drop_proposal".

docs/specification.md
```diff
@@ -687,12 +708,11 @@ Parameter (in Michelson):
...
-proposal are returned to the proposer and voters in full. The decision lambda is skipped.
-- This entrypoint should only be used when there is a proposal that is stuck due to having a
-failing decision lambda.
-- Fails with `NOT_ADMIN` if the sender is not the administrator.
-
+- Delete a proposal when its `proposal_expired_time` is reached.
+- The proposer can call this entrypoint to delete his/her own proposal regardless of `proposal_expired_time`.
+- The `guardian` can call this entrypoint to delete any proposals at anytime regardless of `proposal_expired_time`.
+- Fails with `DROP_PROPOSAL_CONDITION_NOT_MET` when none of the conditions above are met.
+- Tokens that are frozen for this proposal are returned to the proposer and voters as if the proposal is rejected regardless of the actual votes.
```
sras, line 715: Shouldn't we also say that

>if called on an expired proposal, the paid fee is not returned to the proposer

haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -811,19 +817,12 @@ instance CustomErrorHasDoc "bAD_ENTRYPOINT_PARAMETER" where
...
   customErrDocMdCause =
-    "An error occurred when trying to drop a proposal due to the proposal's voting period is not over"
-
-type instance ErrorArg "fAIL_DROP_PROPOSAL_NOT_ACCEPTED" = NoErrorArg
-
-instance CustomErrorHasDoc "fAIL_DROP_PROPOSAL_NOT_ACCEPTED" where
-  customErrClass = ErrClassActionException
-  customErrDocMdCause =
-    "An error occurred when trying to drop a proposal due to the proposal is not an accepted proposal"
+    "Throw when trying to call `drop_proposal` when the sender is not proposer or guardian or proposal is not expired."
```
sras, line 825: > or proposal is not expired.

Not sure, but shouldn't this be "and proposal is not expired"?

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
sras: @rinn7e AFAIU you should set guardian address to the address of a contract that can call the `drop_proposal` entrypoint, and then start by calling the guardian. So the SOURCE would be the address that you used to call the guardian, but SENDER will be the address of the guardian itself.

Makefile
pasqu4le: > guardian is always same as admin address

It's not clear from the description, as it is _possible_ for the admin to be the guardian as well, but in most cases the two _should_ be different.

Note: the `guardian` is supposed to be a contract, so let's provide a `KT...` default value.

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
pasqu4le: @sras is correct, for tests all we need is a very simple `guardian` contract that has no checks and:
- stores the DAO address
- its whole parameter is the same as `drop_proposal`
- when called it will "pass the call forward", transferring to the DAO's `drop_proposal`

src/defaults.mligo
```diff
@@ -14,6 +14,8 @@ let default_config (data : initial_config_data) : config = {
   voting_period = data.voting_period;
   max_proposals = 500n;
   max_votes = 1000n;
+  proposal_flush_time = data.proposal_flush_time;
+  proposal_expired_time = data.proposal_expired_time;
```
pasqu4le, line 18: To continue @sras 's [comment](https://github.com/tqtezos/baseDAO/pull/232#discussion_r632452443) here's where we should make the checks on those values.

src/proposal.mligo
```diff
@@ -344,7 +352,7 @@ let handle_proposal_is_over
     )
     : (operation list * storage * counter) =
   let proposal = check_if_proposal_exist (proposal_key, store) in
-  if    is_voting_period_over (proposal, config.voting_period, store)
+  if    is_proposal_flush_time_reached (proposal, config.proposal_flush_time)
```
pasqu4le, line 355: We need to now check for another condition here, which is that:
> proposals cannot be flushed after `proposal_expired_time` has passed from their submission

and fail when that happens, so:
```suggestion
  if is_proposal_expired_time_reached (proposal, config.proposal_expired_time)
  then (failwith("EXPIRED_PROPOSAL") : (operation list * storage * counter))
  else if is_proposal_flush_time_reached (proposal, config.proposal_flush_time)
```

src/proposal.mligo
```diff
@@ -391,22 +399,22 @@ let flush(n, config, store : nat * config * storage): return =
...
 // Removes an accepted and finished proposal by key.
 let drop_proposal (proposal_key, config, store : proposal_key * config * storage): return =
-  let store = authorize_admin store in
-
   let proposal = check_if_proposal_exist (proposal_key, store) in
-  if is_voting_period_over(proposal, config.voting_period, store)
+
+  if   (sender = proposal.proposer)
+    || (sender = store.guardian && sender <> source) // Guardian cannot be equal to SOURCE
+    || is_proposal_expired_time_reached (proposal, config.proposal_expired_time)
```
pasqu4le, line 406: I suggest calculating the expired condition beforehand as it's needed later too (see next comment):
```suggestion
  let proposal_is_expired = is_proposal_expired_time_reached (proposal, config.proposal_expired_time) in

  if   (sender = proposal.proposer)
    || (sender = store.guardian && sender <> source) // Guardian cannot be equal to SOURCE
    || proposal_is_expired
```

src/proposal.mligo
```diff
@@ -391,22 +399,22 @@ let flush(n, config, store : nat * config * storage): return =
...
-    then
-      let store = unfreeze_proposer_and_voter_token
-            (config.rejected_proposal_return_value, true, proposal, config.voting_period, config.fixed_proposal_fee_in_token, store) in
-      let store = delete_proposal (proposal.start_date, proposal_key, store) in
-      (([] : operation list), store)
-    else
-      (failwith("FAIL_DROP_PROPOSAL_NOT_ACCEPTED") : return)
+    let store = unfreeze_proposer_and_voter_token
+          ( config.rejected_proposal_return_value
+          , false // Expired proposal is treated as rejected proposal regardless of its actual votes
```
pasqu4le, line 410: The comment is correct, however that's _only_ for expired proposals and here we may be dealing also with proposals that are dropped by their `proposer` or by the `guardian` but haven't expired yet, so I think this should be:
```suggestion
          , not proposal_is_expired // Expired proposal is treated as rejected proposal regardless of its actual votes
```

rinn7e: @pasqu4le Does that means if the proposers call `drop_proposal` on their proposals the token are return in full? 

This may pose a cheat where the proposer know the proposal is not gonna get accepted (thus their token got slashed), and call `drop_proposal` beforehand to avoid this. 

What do you think about this?

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
rinn7e: That makes sense. Turn out that, the current `OriginateFn m` which uses the tuple, is a bit troublesome to add a new field `guardian` to it. So I update it to be a record instead and this is quite a big change (mostly renaming), and I put it into a new commit.

The `guardian` contract is done, and I've added to the tests, hope to get your reviews on this.

Makefile
rinn7e: Guardian address being admin address was just a placeholder since I was note quite understand it.

Now I've updated to be a contract address, and add an argument for it as well.

Makefile
rinn7e: I've added arguments for these fields now.

docs/specification.md
rinn7e: Updated

docs/specification.md
rinn7e: Thanks, that definitely sounds better

docs/specification.md
rinn7e: With the current implementation, it treated the return token as if it is a rejected proposal. So it follows the `rejected_proposal_return_value`. My discussion here: https://serokell.slack.com/archives/C019PCGKTK2/p1620821574002100?thread_ts=1620761952.014000&cid=C019PCGKTK2

Now, I'm not sure if I have to state fully about this slash logic in here or not, since it is already stated in the `flush` entrypoints, so I simply state it "...  as if the proposal is rejected". Not sure if this could cause confusion or not.

src/defaults.mligo
rinn7e: Nice, I added the check now.

src/proposal.mligo
rinn7e: I see, but shouldn't we skip it instead? If it fails here, `flush` will keep failing if the amount of proposals to be `flush` include an expired proposal in it. 

Or we want to do it this way to keep the order of execution?

haskell/src/Ligo/BaseDAO/Types.hs
rinn7e: I think using `and` makes it sound like the proposals have to be expired. But it is not the case when the proposer or guardian make the call. They can drop the proposal regardless of whether or not it is expired.

src/proposal.mligo
pasqu4le: Agreed, will update the acceptance criteria as well, let's keep the current implementation.

haskell/src/Ligo/BaseDAO/Types.hs
sras: Yea, that is what I meant as well. The error happens only when caller is not proposer or guardian *and* the proposal is not expired. Right?

Using `or` there will make it as if it will throw error for proposal/guardian if the proposal is not expired. So right now it is like this `(not (proposer || guardian)) || (proposal is not expired)`, and when proposal is not expired this will always be true irrespective of the caller being proposer or guardian, which AFAIU is not correct.

docs/specification.md

haskell/src/Ligo/BaseDAO/Types.hs
rinn7e: Make sense, I've updated this now.

rinn7e: Just realize that the word `or` I use here, could be translated as `XOR` instead of `||`. Anyway, this could be an example case where natural language is quite ambiguous.

src/proposal.mligo
pasqu4le: Yep, we want to keep the order of execution and not allow expired proposals to be flushed.

`flush` will indeed fail if one (or more) proposals are expired, but that why we keep `drop_proposal` and someone will have to call it to move forward.

src/defaults.mligo
```diff
@@ -4,19 +4,26 @@
...
+    voting_period = data.voting_period;
+    max_proposals = 500n;
+    max_votes = 1000n;
+    proposal_flush_time = data.proposal_flush_time;
+    proposal_expired_time = data.proposal_expired_time;
+
+    custom_entrypoints = (Big_map.empty : custom_entrypoints);
+  }
+  else
+    (failwith("INVALID_CONFIG") : config)
```
pasqu4le, line 26: Two small suggestions here, I think we can move the failing validation to a separate function (I intend to extend it in other issues too), plus considering that the error message is only to be evaluated for storage (and it will never be actually part of the contract `code`) we can have longer and more esplicative error messages.

My suggestion:
```suggestion
let validate_default_config (data : initial_config_data) : unit =
  if data.proposal_expired_time <= data.proposal_flush_time then
    failwith("proposal_expired_time needs to be bigger than proposal_flush_time")
  else if data.proposal_flush_time <= (data.voting_period.length * 2n) then
    failwith("proposal_flush_time needs to be more than twice the voting_period length")
  else unit

let default_config (data : initial_config_data) : config =
  let u : unit = validate_default_config(data) in {
    proposal_check = (fun (params, extras : propose_params * contract_extra) -> true);
    rejected_proposal_return_value = (fun (proposal, extras : proposal * contract_extra) -> 0n);
    decision_lambda = (fun (proposal, extras : proposal * contract_extra) -> (([] : (operation list)), extras));

    quorum_threshold = data.quorum_threshold;
    fixed_proposal_fee_in_token = data.fixed_proposal_fee_in_token;
    voting_period = data.voting_period;
    max_proposals = 500n;
    max_votes = 1000n;
    proposal_flush_time = data.proposal_flush_time;
    proposal_expired_time = data.proposal_expired_time;

    custom_entrypoints = (Big_map.empty : custom_entrypoints);
  }
```

src/proposal.mligo
```diff
@@ -391,22 +396,23 @@ let flush(n, config, store : nat * config * storage): return =
...
```
pasqu4le, line 408: I think it would be more clear to say:
```suggestion
          , false // A dropped proposal is treated as rejected regardless of its actual votes
```
Just because the same happens to proposal that are dropped before they are expired.

docs/specification.md
```diff
@@ -146,11 +158,19 @@ Some configuration values are specified in runtime and can be changed during the
 They must be provided on origination to construct the contract's initial storage.
 These values are:
 1. `admin : address` is the address that can perform administrative actions.
-2. `governance_token` is the FA2 contract address/token id pair that will be used as the governance token.
-3. `voting_period : nat` specifies how long the voting period lasts in seconds.
-4. `quorum_threshold : quorum_threshold` specifies what fraction of the frozen
+2. `guardian : address` is the address of a contract that have permission to call `drop_proposal`.
```
pasqu4le, line 161:
```suggestion
2. `guardian : address` is the address of a contract that has permission to call `drop_proposal` on any proposal.
```

docs/specification.md
```diff
@@ -146,11 +158,19 @@ Some configuration values are specified in runtime and can be changed during the
...
+    - `guardian` contract cannot initiate the transaction that results in a call to `drop_proposal`
+3. `governance_token` is the FA2 contract address/token id pair that will be used as the governance token.
+4. `voting_period : nat` specifies how long the voting period lasts in seconds.
+5. `proposal_flush_time : nat`
+    - Specifies how long in seconds after the proposal is proposed such that it can be flushed.
```
pasqu4le, line 166:
```suggestion
    - Specifies in seconds how long after the proposal is proposed it can be flushed.
```

docs/specification.md
```diff
@@ -146,11 +158,19 @@ Some configuration values are specified in runtime and can be changed during the
...
+    - IMPORTANT: Must be bigger than `voting_period * 2`.
+6. `proposal_expired_time : nat`
+    - Specifies how long in seconds after the proposal is proposed such that it can be considered as expired.
```
pasqu4le, line 169:
```suggestion
    - Specifies in seconds how long after the proposal is proposed it is considered expired.
```
This is less of a nitpick as after this time it isn't just that it **can** be considered expired, but it **has** to.

docs/specification.md
```diff
@@ -660,7 +681,7 @@ Parameter (in Michelson):
 (nat %flush)
 ```
 
-- Finish voting process on an amount of proposals for which the voting period is over.
+- Finish voting process on an amount of proposals for which their `proposal_flush_time` is reached.
```
pasqu4le, line 684:
```suggestion
- Finish voting process on an amount of proposals for which their `proposal_flush_time` was reached, but their `proposal_expire_time` wasn't yet. 
```

docs/specification.md
pasqu4le: I think the current phrasing is correct, as we now established that dropped proposals are to be treated as failed.

pasqu4le: I think the current phrasing is correct, as we now established that dropped proposals are to be treated as failed.

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -50,7 +51,21 @@ import Ligo.BaseDAO.Types
 import Test.Ligo.BaseDAO.Common.StorageHelper as StorageHelper
 import Test.Ligo.BaseDAO.Proposal.Config (ConfigDesc, fillConfig)
 
-type OriginateFn m = m ((Address, Address), (Address, Address), TAddress Parameter, TAddress FA2.Parameter, Address)
+-- type m DaoOriginateData = m ((Address, Address), (Address, Address), TAddress Parameter, TAddress FA2.Parameter, Address)
+type OriginateFn m = m DaoOriginateData
+
+  -- pure ((dodOwner1, operator1), (dodOwner2, operator2), dodDao, dodTokenContract, admin)
```
pasqu4le, line 57: I think this is much better, let's remove these comments.

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -42,156 +42,167 @@ test_BaseDAO_Proposal =
...
-      , nettestScenarioOnEmulator "flush and run decision lambda" $
-          \_emulated ->
-            uncapsNettest $ flushDecisionLambda (originateLigoDaoWithConfigDesc dynRecUnsafe)
-      , nettestScenarioOnEmulator "can drop proposals" $
-          \_emulated ->
-            uncapsNettest $ dropProposal (originateLigoDaoWithConfigDesc dynRecUnsafe)
+      , nettestScenarioOnEmulatorCaps "flush and run decision lambda" $
+          flushDecisionLambda (originateLigoDaoWithConfigDesc dynRecUnsafe)
+
+      , nettestScenarioOnEmulatorCaps "can drop proposals" $
```
pasqu4le, line 118:
```suggestion
      , nettestScenarioOnEmulatorCaps "can drop proposals, only when allowed" $
```
I think this should be updated as I see you have extended the `dropProposal` test to include the check for forbidden cases as well.

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -42,156 +42,167 @@ test_BaseDAO_Proposal =
...
       -- TODO [#38]: Improve this when contract size is smaller
-      , nettestScenarioOnEmulator "flush and run decision lambda" $
-          \_emulated ->
-            uncapsNettest $ flushDecisionLambda (originateLigoDaoWithConfigDesc dynRecUnsafe)
-      , nettestScenarioOnEmulator "can drop proposals" $
-          \_emulated ->
-            uncapsNettest $ dropProposal (originateLigoDaoWithConfigDesc dynRecUnsafe)
+      , nettestScenarioOnEmulatorCaps "flush and run decision lambda" $
+          flushDecisionLambda (originateLigoDaoWithConfigDesc dynRecUnsafe)
+
```
pasqu4le, line 117: As a reminder connected to the other comment, there needs to also be a test to check that no proposal can be flushed after it expired.

src/proposal.mligo
rinn7e: I see, I've updated this now.

src/defaults.mligo
rinn7e: Definitely looks better, I've updated this now.

src/proposal.mligo
rinn7e: Updated

docs/specification.md
rinn7e: Updated

haskell/test/Test/Ligo/BaseDAO/Common.hs
rinn7e: Leftover, I've removed this now

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
rinn7e: Added a "flush fail on drop proposal" test case now.

src/proposal.mligo
```diff
@@ -344,27 +348,32 @@ let handle_proposal_is_over
...
-      else (([] : operation list), store)
-    in
-    let cons = fun (l, e : operation list * operation) -> e :: l in
-    let ops = List.fold cons ops new_ops in
-    let store = delete_proposal (start_date, proposal_key, store) in
-    (ops, store, counter)
-  else (ops, store, counter)
+
+  if is_time_reached (proposal, config.proposal_expired_time)
+    then (failwith("EXPIRED_PROPOSAL") : (operation list * storage * counter))
```
pasqu4le, line 353: `EXPIRED_PROPOSAL` should be added to the error list in the spec.

pasqu4le (approved):
One last thing, the rest LGTM

src/proposal.mligo
rinn7e: I've added this in a separate commit: [`4669c84` (#232)](https://github.com/tqtezos/baseDAO/pull/232/commits/4669c84ecec2ec93723844ae16bab7902ae9c0a3)

Have to do this, to avoid problem with rebasing.

Issue #233, opened 2021-05-17
Stabilize the contract and make a new release


pasqu4le:
# Clarification and motivation

After more recent changes in the contract the logic of the contract is fairly different from the last release.

We should make a new BYO-centered one, but before doing that we need to check if:
1. there are no leftovers (in `code` and `storage`) from previous changes
2. there are changes that don't alter the logic, but optimize it
3. there are changes that do alter the logic, but should be (re-)considered
4. everything is properly covered by tests
5. the documentation is correct and the usage is as comfortable as possible

# Acceptance criteria

- new issues are opened if any of the points above requires further consideration or code modifications
- a new release is made (after everything above is concluded)

PR #241, opened 2021-05-18
[#233] Bump Haskell dependencies and remove Lorentz version

pasqu4le:
## Description

Bumps Haskell dependencies to a newer versions.

Moreover it removes the outdated Lorentz version of the contract entirely.

## Related issue(s)

Part of #233 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


haskell/src/Ligo/Typescript.hs
```diff
@@ -315,6 +316,9 @@ mkTypesFor typename epType = case U.unwrapT epType of
   U.TBls12381G1 -> [TsType typename (TsAlias TsString)]
   U.TBls12381G2 -> [TsType typename (TsAlias TsString)]
   U.TNever -> [TsType typename (TsAlias TsVoid)]
+  U.TTicket a ->
+    let ticketFields = [("ticketer", toType U.TAddress), ("value", a), ("amount", toType U.TNat)]
+    in [TsType typename (TsInterface $ map (fst . mkTypesForField False) ticketFields)]
```
sras, line 321: Here you seem to be throwing away the `[TsDecl]` returned by  the `mkTypesForField`. This should be fine for the first and last elements for `ticketFields` as they are primitives. But for the second field, the type is `a`, which could require additional declarations. So if we throw the `[TsDecl]` away, the nested declarations for `a` won't be included. So I think something like the following might be required here (taken from the TPair branch of the same function.

```
        decls = (mkTypesForField False) <$> (indexEmptyFields allFields)
        recordFields = fst <$> decls
        thisType = TsType typename (TsInterface recordFields)
        thisDecls = snd <$> decls
        in thisType : (concat thisDecls)
```

```
        decls = (mkTypesForField False) <$> (indexEmptyFields allFields)
        recordFields = fst <$> decls
        thisType = TsType typename (TsInterface recordFields)
        thisDecls = snd <$> decls
        in thisType : (concat thisDecls)
```

sras (approved):
LGTM

rinn7e:
I couldn't build this locally, due to this error:
```
[1 of 6] Compiling Ligo.BaseDAO.Types
<command line>: /home/rinne/.stack/snapshots/x86_64-linux-tinfo6/9f1ebce10328f480dad134acff21eed56118d6ab46429e4dbe02498c01f0293a/8.10.4/lib/x86_64-linux-ghc-8.10.4/libHSmorley-metadata-0-Gk5LOjJkrb7FeQfRlzMWR2-ghc8.10.4.so: undefined symbol: lorentzzzm0zi11zi0zmDlcSfoFCsgnH10lVQsgBTs_LorentzzziMacro_unpair2_closure
```
Not sure how to fix this.

pasqu4le:
@rinn7e oddy I had a similar problem, not sure why but it's an issue with stack.

After I cleaned (part of) my `.stack` dir it recompiled and all worked correctly, I'm afraid you'll need to do the same.

rinn7e:
> @rinn7e oddy I had a similar problem, not sure why but it's an issue with stack.
> 
> After I cleaned (part of) my `.stack` dir it recompiled and all worked correctly, I'm afraid you'll need to do the same.

I have to nuke the entire dir, but it is working fine now.

rinn7e (approved):

sras (approved):

PR #243, opened 2021-05-19
[#238] Remove morley-large-originator 

### LINKED ISSUE #238: Remove morley-large-originator if possible

With the activation of `florence` the limit for the origination contract size is much higher and it should be possible to not need `morley-large-originator` to do so.

We should check if that's indeed the case and stop recommending its use if that's the case.

Additionally, we can also stop using it in our tests.

Related to #233 

# Acceptance criteria

If the contract can be originated without `morley-large-originator`:
1. we no longer suggest to use it in our documentation
2. we no longer use `originateLarge*` in our test suite

rinn7e:
## Description

Problem: With the activation of `florence` the limit for the origination contract
size is much higher and it should be possible to not need `morley-large-originator` to do so.
For the context:
- `baseDAO.tz` with `trivialDAO` storage is around `18kB`
- `baseDAO.tz` with `registryDAO` storage is around `26kB`
- `baseDAO.tz` with `treasuryDAO` storage is around `23kB`

The previous limit is `16kB`. With the new protocol, the limit is now increased to around `32kB`.

Solution: Remove `morley-large-orginator` from `Makefile`, remove
`originating.md`, replace the mention of it with `tezos-client` in
README, update the tests.

## Related issue(s)

Resolves #238

- Tests
        silently reappearing again.

- Documentation
    - Haddock


README.md
```diff
@@ -43,12 +43,13 @@ how to spend its XTZ and tokens.
...
+The contract can be originated with `tezos-client`:
 
-To overcome this limitation we advice to use the [morley-large-originator](https://gitlab.com/morley-framework/morley/-/tree/master/code/morley-large-originator)
-tool or the workaround that is described there.
+```sh
+tezos-client originate contract baseDAO transferring 0 from alice running BaseDAO.tz --init "$(<storage)" --burn-cap 17.0
+```
 
-For more information about this follow the [origination guide](docs/originating.md).
+Instructions on how to install `tezos-client` can be found at official [Tezos installation page](http://tezos.gitlab.io/introduction/howtoget.html) or in [our `tezos-packaging` repository](https://github.com/serokell/tezos-packaging).
```
rinn7e, line 52: @pasqu4le Took me a while to set up `tezos-client` with the new protocol to test if it works. The size now increase from 16kB to 32kB which allows originating baseDAO now (baseDAO size is 18kB).

Related to README.md, I'm not sure what words to put in here, so I just used what was written before and put it back in. Let me know if this is good enough.

pasqu4le: > baseDAO size is 18kB

That's the contract size on its own, right?
Did you try to originate the various instantiation (`registryDAO` and `trivialDAO`)?

I ask because the initial storage also has to be taken into account in the origination and in baseDAO with the lambdas it can be quite big (I don't think it will be big enough to reach the limit tho, but I'm just checking).

pasqu4le: > Related to README.md

Just checked it, I think it's fine the way you changed it :ok_hand:

stack.yaml.lock
```diff
@@ -32,10 +32,10 @@ packages:
   original:
     hackage: hex-text-0.1.0.0
 - completed:
-    hackage: named-0.3.0.1@sha256:2975d50c9c5d88095026ffc1303d2d9be52e5f588a8f8bcb7003a04b79f10a06,2312
+    hackage: named-0.3.0.1@sha256:418543272f0c844a77639b05276ee3906d5f305cd4cab46fafcb4f15b7f1e60d,2312
```
pasqu4le, line 35: It's quite odd that you had to update this, hopefully that's because you are based on an older version of my PR.

Since that's been merged a rebase should remove this need, I think.

pasqu4le (approved):
LGTM, feel free to rebase and merge this.

README.md
rinn7e: > That's the contract size on its own, right?

I just realize that it is only with the trivialDao storage. With registryDAO and treasuryDAO storage, they are around 26kB and 23kB respectively. 

Definitely not far from the max size and it could be a concern in the future.

pasqu4le: I think that's plenty of margin however.

Some simple contracts can entirely fit inside the difference there :sweat_smile: 

This should be more than enough for custom DAOs and the contract implementation is not supposed to grow anymore, so this is ok :ok_hand:

PR #246, opened 2021-05-20
[#229] store voting stage number in proposals

### LINKED ISSUE #229: Store offset period_num in proposals

`proposal.period_num` is not directly used, only `proposal.period_num + 1n`.

# Acceptance criteria

- Replace

```ocaml
    ; period_num = current_period
```

with

```ocaml
    ; period_num = current_period + 1n
```

in `add_proposal`

- Replace occurrences of `proposal.period_num + 1n` with `proposal.period_num`

pasqu4le:
## Description

Problem: when creating a new proposal we store the stage number in
which this was created, however we only need to know in which stage
number this can be voted on.

Solution: modify the proposal type to contain the voting period
stage.

Note: currently based on #245 

## Related issue(s)

Resolves #229

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/src/Ligo/BaseDAO/Types.hs
+++ b/haskell/src/Ligo/BaseDAO/Types.hs
@@ -437,7 +437,7 @@ data Proposal = Proposal
   { plUpvotes                 :: Natural
   , plDownvotes               :: Natural
   , plStartDate               :: Timestamp
-  , plPeriodNum               :: Natural
+  , plVotingStageNum          :: Natural
 
   , plMetadata                :: ProposalMetadata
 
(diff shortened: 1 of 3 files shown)


sras (approved):
AFAIU this only moves one `+` operation from `propose` operation to `vote` operation. Since `vote` is called many more times than `propose` this makes sense to me. But is there any other advantages?

rinn7e (approved):

pasqu4le:
> But is there any other advantages?

No, not really, that's the only reason as far as the code is concerned.

Perhaps it makes it easier to use when looking at the storage from an explorer too, but I;m not sure about that.

PR #247, opened 2021-05-20
[#233] small refactoring

pasqu4le:
## Description

Some refactoring and clean-up of the Ligo source code.

Note: currently based on #246

Also, the spec might be slightly outdated, but I plan to check and update it later today.

## Related issue(s)

Part of #233

- Tests
        silently reappearing again.

- Documentation
    - Haddock


haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -206,21 +206,12 @@ newtype VotingPeriod = VotingPeriod { unVotingPeriod :: Natural}
 instance HasAnnotation VotingPeriod where
   annOptions = baseDaoAnnOptions
 
-newtype MaxChangePercent = MaxChangePercent Natural
```
sras, line 209: These newtypes for `MaxChagePercent` and `ChangePercent` exists only because we might want to have different `IsConfigDescExt` instance for them to set them in the `Config` during tests. It appears that we don't need those instances as of now. So it might be fine.

haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -706,8 +697,8 @@ mkFullStorage
   :: "admin" :! Address
   -> "votingPeriod" :? VotingPeriod
   -> "quorumThreshold" :? QuorumThreshold
-  -> "maxChangePercent" :? MaxChangePercent
-  -> "changePercent" :? ChangePercent
+  -> "maxQuorumFraction" :? QuorumFraction
```
sras, line 700: "maxQuorumFraction" does not sound much like it is representing "maxChangePercent". Any particular reason why this was changed?

haskell/src/Ligo/BaseDAO/Types.hs
pasqu4le: Oh I see, I missed that point.

Since using the `IsConfigDescExt` will require to build the datatype (whichever we chose) I think this can be also used well by using `fillConfig` with ``percentageToFractionNumerator` (or something similar) if we ever need to :thinking: 

Anyway since as you said we don't need this now I would keep this change as it's closer to how the LIGO types are defined.

haskell/src/Ligo/BaseDAO/Types.hs
pasqu4le: No good reason, this was changed by accident :sweat_smile: 

Fixed now

src/proposal.mligo
```diff
@@ -260,32 +170,14 @@ let vote(votes, config, store : vote_param_permited list * config * storage): re
...
-// the fixed fee payed for the proposal. The case when the desired burn amount
-// is larger than the available frozen tokens is possible because the contract
-// administrator can transfer the proposer's frozen tokens.
-[@inline]
-let burn_what_possible (desired_burn_amount, frozen_tokens, addr, store : nat * nat * address * storage): storage =
-  let to_burn =
-    if frozen_tokens >= desired_burn_amount
-    then desired_burn_amount
-    else frozen_tokens
-  in burn_frozen_token (to_burn, addr, store)
```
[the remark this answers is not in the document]
pasqu4le: > the user might have already unfreezed some of their tokens

Yes, but not the ones he stakes to create the proposal, `unfreeze` will prevent that.

So since this needs only to be able to burn tokens that were previously staked to propose this should never result in an issue.

I added a `"cannot unfreeze staked tokens"` test for that now.

sras (approved):

PR #248, opened 2021-05-20
[#234] Throw more and better failure messages on storage creation

### LINKED ISSUE #234: Throw more and better failure messages on storage creation

In most cases we handle storage generation checks (e.g. on bounds) silently, by making sure that the necessary properties are respected.

This is not very good as it's not clear from the command execution that something required a change, we should use `validate_*` functions instead and throw an error message instead when a change is necessary.

Moreover, we know from the start that some of the `Makefile` target parameters should be provided (e.g. the `admin` address), however we provide default values for them.

This also is likely a bad idea because it may not make very clear that something is necessary. 

# Acceptance criteria

1. which of the `Makefile` storage targets argument should always be provided by the user is established and default values are no longer provided for these
2. storage validation checks are handled with failure messages and no longer with silent changes
3. the `building` documentation is updated accordingly.

Note: because this is storage generation and they will not end up in contract `code`, messages can be as long and expressive as we want.

rinn7e:
## Related issue(s)

Resolves #234

- Tests
        silently reappearing again.

- Documentation
    - Haddock

Problem: the last tested version of ligo is now quite old, so we want
to update to a newer one.

Solution: check the latest version, update nix sources for CI and
documentation.Problem: the last tested version of ligo is now quite old, so we want
to update to a newer one.

Solution: check the latest version, update nix sources for CI and
documentation.


rinn7e:
@pasqu4le This PR is still WIP but I'd like your view on my approach to this issue.

Basically, what I did is that, I turn all field of `init_config` and `init_storage` to be `option` so that we can print proper: `<some-field> cannot be empty`. We can also do some validation such as `proposal_expired_time` cannot be small than `proposal_flush_time` as part of this check.

Currently, I disable the storage generation for registry and treasury dao for now.

So hope to get your view on this, whether or not this is the right approach. 


Example usage:
+ Success command
```
make out/trivialDAO_storage.tz \
  admin_address="tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" \
  guardian_address="KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh" \
  governance_token_address="KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5" \
  governance_token_id=0n \
  now_val=`date +"%s"` \
  metadata_map="(Big_map.empty : metadata_map)" \
  ledger="([] : ledger_list)" \
  quorum_threshold=10n \
  min_quorum=1n \
  max_quorum=99n \
  voting_period=950400n \
  quorum_change=5n \
  max_quorum_change=19n \
  governance_total_supply=100n \
  proposal_flush_time=1900801n \
  proposal_expired_time=2851200n \
  ```
  
+ Fail command:
```
make out/trivialDAO_storage.tz \
  admin_address="tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af"
```

Error: `An error occurred while evaluating an expression: 'fixed_proposal_fee_in_token' cannot be empty.`

**Comment by pasqu4le** [comment]
@rinn7e your approach would work, but it seems very verbose and complicates things a lot.

What if instead of representing the optional property in the LIGO types we'd set the "default value" to the error in the `Makefile` itself?

E.g. for trivialDAO we can change the `admin_address` definition to:
```make
$(OUT)/trivialDAO_storage.tz : admin_address = failwith(\"admin_address must be supplied\")
```
and the result would be the same, if not provided the returned message is:
```
An error occurred while evaluating an expression: admin_address must be supplied
```

Note, you'll also need to remove the escaped surrounding double quotes a few lines below, al least in this case:
```
          { admin = ($(admin_address) : address) \
```

rinn7e:
> @rinn7e your approach would work, but it seems very verbose and complicates things a lot.
> 
> What if instead of representing the optional property in the LIGO types we'd set the "default value" to the error in the `Makefile` itself?

I see, this seems like a better approach, let me try to use this instead.

rinn7e:
>           { admin = ($(admin_address) : address) \

@pasqu4le I'm facing a problem with Makefile/shell removing the double quote.

Example: Even if I input arguments like this:
```
make out/registryDAO_storage.tz \
  admin_address="tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af" \
  guardian_address="KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh"
```

The double quote is somehow stripped, and result in :
```
ligo compile-storage --syntax cameligo --output-file out/registryDAO_storage.tz \
      src/registryDAO.mligo base_DAO_contract "default_registry_DAO_full_storage( \
        { base_data = \
          { storage_data = \
            { admin = (tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af : address) \
            ; guardian = (KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh : address) \
```

Do you have any advice on this? Trying to escape the quotes like this: `admin_address=\"tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af\"` doesn't work either.

pasqu4le:
>The double quote is somehow stripped, and result in :

Oh I know what's happening, it's because we use double quotes to enclose the whole string in the `Makefile`, see:
```make
<...>
          })"
```
> Trying to escape the quotes like this: admin_address=\"tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af\" doesn't work either.

:thinking: yeah that also won't work, I managed a workaround, but I think we shouldn't use it because it's inconvenient to do the escaping in the command line.

It took me a while, but I think the best solutions is to not provide a default and have `make` throw an error in such a case.

So I propose to add this definition (right below the one for `escape_double_quote`):
```make
# Utility function to require a variable to be defined
require_defined = $(if $(value $1),$(value $1),$(error "$1 needs to be defined"))
```
to remove the `admin_address` default definition:
```make
$(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
```
and instead use the new utility function below in the interpolation:
```
          { admin = (\"$(call require_defined,admin_address)\" : address) \
```

This seems to work fine as for `make out/trivialDAO_storage.tz` the result is a failure:
```
Makefile:82: *** "admin_address needs to be defined".  Stop.
```
but `make out/trivialDAO_storage.tz admin_address=tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af` works as expected.

PR #251, opened 2021-05-21
[#234] Throw more and better failure messages on storage creation

### LINKED ISSUE #234: Throw more and better failure messages on storage creation

rinn7e:
## Description

Problem: Currently, we provide `Makefile` arugments by default and we
handle some storage check silently. This is not really a good idea.

Solution: Remove default values in `Makefile`, and throw error instead.
Add more validate function to storage generation.

## Related issue(s)

Resolves #234 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ BUILD_PARAMETER = $(LIGO) compile-parameter --syntax cameligo
 escape_quote = $(subst ','\'',$(1))
 # Utility function to escape double quotes
 escape_double_quote = $(subst $\",$\\",$(1))
+# Utility function to require a variable to be defined
+require_defined = $(if $(value $1),$(value $1),$(error "$1 needs to be defined"))
 
 # Where to put build files
 OUT ?= out
@@ -33,10 +35,7 @@ TS_OUT ?= typescript
 .PHONY: all clean test typescript
 
 all: \
-	$(OUT)/baseDAO.tz \
-	$(OUT)/trivialDAO_storage.tz \
-	$(OUT)/registryDAO_storage.tz \
-	$(OUT)/treasuryDAO_storage.tz
+	$(OUT)/baseDAO.tz
 
 # Compile LIGO contract into its michelson representation.
 $(OUT)/baseDAO.tz: src/**
@@ -57,14 +56,10 @@ ifeq ($(OPTIMIZE), true)
 endif
 	#
 
-$(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af
-$(OUT)/trivialDAO_storage.tz : guardian_address = KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh
-$(OUT)/trivialDAO_storage.tz : governance_token_address = KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5
-$(OUT)/trivialDAO_storage.tz : governance_token_id = 0n
 $(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
-$(OUT)/trivialDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
+$(OUT)/trivialDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
-$(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
+$(OUT)/trivialDAO_storage.tz : ledger = []
 $(OUT)/trivialDAO_storage.tz : quorum_threshold = 10n
 $(OUT)/trivialDAO_storage.tz : min_quorum = 1n
 $(OUT)/trivialDAO_storage.tz : max_quorum = 99n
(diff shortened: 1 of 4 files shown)


rinn7e:
@pasqu4le Continue from the discussion from here: https://github.com/tqtezos/baseDAO/pull/248#issuecomment-845915783

I've made the changes according to your review, and it works fine now. I'm not sure what are the other validate functions to put in, currently I only add `validate_quorum_threshold` check for now. 

Hope to get your review on this.

Makefile
```diff
@@ -24,6 +24,8 @@ BUILD_PARAMETER = $(LIGO) compile-parameter --syntax cameligo
 escape_quote = $(subst ','\'',$(1))
 # Utility function to escape double quotes
 escape_double_quote = $(subst $\",$\\",$(1))
+# Utility function to require a variable to be defined
+require_defined = $(if $(value $1),$(value $1),$(error "$1 needs to be defined"))
```
sras, line 28: Does this work if `$1` is zero?

rinn7e: It seems to work fine if `$1` is `0`. It only throws error if the value `$1` is an empty string.

sras:
@rinn7e Not sure, but it appears that we still have default values for trivialDAO in the Makefile. Shouldn't we remove it as well?

https://github.com/tqtezos/baseDAO/blob/a04d1a3a735af25017d00e01b6e2c99c30fdb47d/Makefile#L59

rinn7e:
> @rinn7e Not sure, but it appears that we still have default values for trivialDAO in the Makefile. Shouldn't we remove it as well?
> 
> https://github.com/tqtezos/baseDAO/blob/a04d1a3a735af25017d00e01b6e2c99c30fdb47d/Makefile#L59

My bad, forgot to remove this. I've pushed an update now.

sras (approved):
LGTM now

Makefile
```diff
@@ -57,188 +56,196 @@ ifeq ($(OPTIMIZE), true)
...
-$(OUT)/trivialDAO_storage.tz : ledger = ([] : ledger_list)
-$(OUT)/trivialDAO_storage.tz : quorum_threshold = 10n
-$(OUT)/trivialDAO_storage.tz : min_quorum = 1n
-$(OUT)/trivialDAO_storage.tz : max_quorum = 99n
-$(OUT)/trivialDAO_storage.tz : period = 950400n # 11 days
-$(OUT)/trivialDAO_storage.tz : quorum_change = 5n
-$(OUT)/trivialDAO_storage.tz : max_quorum_change = 19n
-$(OUT)/trivialDAO_storage.tz : governance_total_supply = 100n
-$(OUT)/trivialDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
-$(OUT)/trivialDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
```
pasqu4le, line 76: Oh, you ended up making them all required, I was thinking of removing the default value only for some of them.

See the issue description again, I think it's pretty clear that what fields to require should be decided on first, especially from the first acceptance criteria:
> which of the Makefile storage targets argument should always be provided by the user is established and default values are no longer provided for these

I don't think they all fit the condition that they "should be provided" by the user.

I mean, my point is that with the default e.g. `admin_address` or `guardian_address` a real contract won't actually make sense, so the user should provide them in any case.
However, values like the one of `period` or `quorum_change` have decent default values which can be used for a real deployment, so we can keep giving defaults that can be overridden.

src/defaults.mligo
```diff
@@ -4,15 +4,21 @@
 #include "types.mligo"
 #include "proposal.mligo"
```
pasqu4le, line 6: One more thing to check is that the `quorum_threshold` is between the `min_quorum_threshold` and `max_quorum_threshold` values.

rinn7e: I've pushed an update to this now.

Makefile
pasqu4le: No problem, perhaps I should have made it more clear.

Yes, I think those should be all the required values :+1:

rinn7e: If that's the case, I've pushed the update. You can take a look at it.

Makefile
```diff
@@ -203,42 +190,62 @@ $(OUT)/treasuryDAO_storage.tz: src/**
...
+          ; max_xtz_amount = ($(call require_defined,max_xtz_amount) : tez) \
           })"
 
 	# ============== Compilation successful ============== #
 	# See "$(OUT)/treasuryDAO_storage.tz" for compilation result #
 	#
 
-test: all
+# For development and testing purpose
+storage:
```
pasqu4le, line 229: I suggest renaming this target `test-storage` instead

Makefile
```diff
@@ -203,42 +190,62 @@ $(OUT)/treasuryDAO_storage.tz: src/**
...
-          ; slash_division_value = $(slash_division_value) \
-          ; min_xtz_amount = $(min_xtz_amount) \
-          ; max_xtz_amount = $(max_xtz_amount) \
+          ; frozen_scale_value = ($(call require_defined,frozen_scale_value) : nat) \
+          ; frozen_extra_value = ($(call require_defined,frozen_extra_value) : nat) \
+          ; max_proposal_size = ($(call require_defined,max_proposal_size) : nat) \
+          ; slash_scale_value = ($(call require_defined,slash_scale_value) : nat) \
+          ; slash_division_value = ($(call require_defined,slash_division_value) : nat) \
+          ; min_xtz_amount = ($(call require_defined,min_xtz_amount) : tez) \
+          ; max_xtz_amount = ($(call require_defined,max_xtz_amount) : tez) \
```
pasqu4le, line 221: These changes to use `require_defined` are now unnecessary.

Since there is a default value there will never be a case in which there are undefined.

I mean, it could only happen if one purposefully calls e.g. `make out/treasuryDAO_storage.tz max_xtz_amount=""`, but LIGO will throw a better error in that case anyway.

Note: this comment applies to the other non-necessary changes like this, including the other storage targets.

Makefile
```diff
@@ -184,9 +171,9 @@ $(OUT)/treasuryDAO_storage.tz : slash_division_value = 0n
 $(OUT)/treasuryDAO_storage.tz : min_xtz_amount = 0mutez
 $(OUT)/treasuryDAO_storage.tz : max_xtz_amount = 100mutez
 $(OUT)/treasuryDAO_storage.tz : now_val = `date +"%s"`
-$(OUT)/treasuryDAO_storage.tz : metadata_map = (Big_map.empty : metadata_map)
+$(OUT)/treasuryDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/treasuryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
-$(OUT)/treasuryDAO_storage.tz : ledger = ([] : ledger_list)
+$(OUT)/treasuryDAO_storage.tz : ledger = []
```
pasqu4le, line 176: Even if you revert the changes below (as I suggested in the other comment) to remove the use of `require_defined` that are not necessary, I think we should keep this (and move only the type annotation in the call below).

It seems to me it makes things cleaner overall.

src/defaults.mligo
```diff
@@ -4,15 +4,22 @@
...
   else if data.proposal_flush_time <= (data.period.length * 2n) then
-    failwith("proposal_flush_time needs to be more than twice the period length")
+    failwith("'proposal_flush_time' needs to be more than twice the 'period' length.")
   else unit
 
+let validate_quorum_threshold_bound (data : initial_config_data) : unit =
+  if data.max_quorum >= data.quorum_threshold && data.quorum_threshold >= data.min_quorum then
+    unit
+  else
+    failwith("'quorum_threshold' value need be between 'max_quorum' and 'min_quorum'.")
```
pasqu4le, line 18: Let's be as detailed as possible here, I think we should split the check in two and give two different messages for "greater or equal then `min_quorum`" and "less than or equal to `max_quorum`".

Makefile
rinn7e: I kept it in case when we need it in the future. But I'm good with removing it for now. I pushed the update and handled other changes as well.

Makefile
```diff
@@ -203,42 +190,62 @@ $(OUT)/treasuryDAO_storage.tz: src/**
...
+		governance_token_address=KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5 \
+		governance_token_id=0n
+
+	make $(OUT)/registryDAO_storage.tz \
+		admin_address=tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE9Af \
+		guardian_address=KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh \
+		governance_token_address=KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5 \
+		governance_token_id=0n
+
+test: storage all
```
pasqu4le, line 248:
```suggestion
test: test-storage all
```

pasqu4le (approved):
Only one last minor thing, but LGTM now

Issue #47, opened 2020-11-11
Refine time-sensitive tests


gromakovsky:
# Clarification and motivation

Some logic of the contract depends on time (`NOW` value) and it's a bit tricky to write proper tests for this logic that would work in any network regardless of block interval. For example, in some tests we want to make sure that the time between `propose` and `vote` calls is less than the voting period, but we don't want to set the voting period to a _too big_ value because we also want this voting period to be over within the test. I recently opened a somewhat related issue: https://gitlab.com/morley-framework/morley/-/issues/433

One way around it is to wait for the aforementioned #433 and set voting period to `2 * block interval` and make sure `propose` and `vote` are called successfully. However, it's still a bit fragile because there can be more than `2 * block interval` between 2 blocks.

A better way in this case is to employ batching and put `vote` and `propose` calls in one batch. Then we can be sure they have the same `NOW` value. However, it's not clear when proper batching will work reliably in `morley-client` and `cleveland`.

# Acceptance criteria

1. Time-sensitive tests run in emulator and network.
2. They work reliably and don't fail if some block is missing.
3. It's fine to make assumptions like "if (average, approximate) block interval is 30 seconds, time between blocks is limited by 60 seconds". I think it's not guaranteed to always hold, but should fail very rarely. It's preferred to avoid assumption with non-zero probability of being wrong, of course.

rvem:
Perhaps https://gitlab.com/morley-framework/morley/-/merge_requests/737 can be useful

pasqu4le:
@sras since the morley MR above has been merged I think we can start working on this right away.

Because `florence` tests are disabled however we'll probably not merge the solution until #124 is closed

pasqu4le:
Update: time-sensitive tests (now `LEVEL`-sensitive) have been tweaked already to run as reliably as possible on the chain in #294 

We should however check that they work just as well in a non-local chain too, which is to be done by #124
If this is the case, this issue can be closed.

PR #252, opened 2021-05-21
[#47] Implement batch calls in tests where possible

sras:
Problem: In our tests, we sometimes send consecutive calls from the same
sender. And in some such places batching can be implemented to make it
faster.

Solution: Find such places, and use batching there. Three places have
been left out because it appear not easy to batch `createSampleProposal`
calls without duplicating some code.

## Related issue(s)

Resolves part of #47 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Ligo/BaseDAO/Common.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Common.hs
@@ -1,5 +1,6 @@
 -- SPDX-FileCopyrightText: 2021 TQ Tezos
 -- SPDX-License-Identifier: LicenseRef-MIT-TQ
+{-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE NumericUnderscores #-}
 
 module Test.Ligo.BaseDAO.Common
@@ -21,6 +22,7 @@ module Test.Ligo.BaseDAO.Common
   , sendXtz
 
   , createSampleProposal
+  , createSampleProposals
   , originateLigoDaoWithBalance
   , originateLigoDaoWithConfigDesc
   , originateLigoDao
@@ -36,6 +38,7 @@ import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Lorentz.Test (contractConsumer)
 import Michelson.Typed.Convert (convertContract, untypeValue)
 import Morley.Nettest
+import Morley.Nettest.Caps (MonadOps)
 import Named ((!))
 import Util.Named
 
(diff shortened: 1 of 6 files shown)


haskell/test/Test/Ligo/BaseDAO/Management.hs
```diff
@@ -191,24 +192,28 @@ notSetAdmin withOriginatedFn initialStorage =
...
 rewritePendingOwner withOriginatedFn initialStorage =
   withOriginatedFn 3 (\(owner:_) -> initialStorage owner) $
     \[owner, wallet1, wallet2] baseDao -> do
-      withSender owner $ do
-        call baseDao (Call @"Transfer_ownership")
-          (#newOwner .! wallet1)
-        call baseDao (Call @"Transfer_ownership")
-          (#newOwner .! wallet2)
+      withSender owner $
+        inBatch $ do
```
pasqu4le, line 211: Nitpick, but I think it's unnecessary to indent this further, I suggest:
```suggestion
      withSender owner . inBatch $ do
```
(same in the other places).

pasqu4le:
> it appear not easy to batch `createSampleProposal` calls without duplicating some code.

Even tho there are only a few places where it would be useful at the moment, I think there may be more in the future and this can be done.

A solution to avoid duplication would be to add a new function, `createSampleProposals` that takes a `counter_list` instead of a `counter` and then:
- generates `params` for each `counter` in the list
- makes a batch call for all the params
- return the list of keys, each generated with `makeProposalKey`.

Since this also works with one proposal to avoid duplication as well as modifying all the places where we only create one we can have something like:
```haskell
createSampleProposal counter dodOwner1 dao = head <$> createSampleProposals [counter] dodOwner1 dao
```

Note: perhaps instead of a list it would be better to have a non-empty list.

pasqu4le:
I spotted one more opportunity for batch calls, which are consecutive calls to `checkTokenBalance`.

This too can be modified into a function that accepts multiple addresses and returns a list of values, using the same `consumer`.

Possibly there are also other similar "common" functions that can be batched, but I haven't checked.

sras:
>A solution to avoid duplication would be to add a new function

I tried that already, but didn't worked out. May be I have missed something. WIll try again.

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -271,11 +274,30 @@ originateLigoDao =
 createSampleProposal
   :: (MonadNettest caps base m, HasCallStack)
   => Int -> Address -> TAddress Parameter -> m ProposalKey
-createSampleProposal counter dodOwner1 dao = do
+createSampleProposal counter dodOwner dao = do
+  let (pk, action) = createSampleProposal_ counter dodOwner dao
+  withSender dodOwner $ inBatch action
```
pasqu4le, line 279: The `inBatch` is not needed here:
```suggestion
  withSender dodOwner action
```

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -271,11 +274,30 @@ originateLigoDao =
...
+createSampleProposal_ counter dodOwner1 dao =
   let params = ProposeParams
         { ppFrozenToken = 10
         , ppProposalMetadata = lPackValueRaw @Integer $ fromIntegral counter
         }
+  in (makeProposalKey params dodOwner1, call dao (Call @"Propose") params)
 
-  withSender dodOwner1 $ call dao (Call @"Propose") params
-  pure $ (makeProposalKey params dodOwner1)
+createSampleProposals
```
pasqu4le, line 292: Can you please add this TODO, so we can reconsider it later if we have more use for it?
```suggestion
-- TODO consider making this polymorphic on the input/output size
createSampleProposals
```

pasqu4le (approved):

PR #253, opened 2021-05-24
[#239] Suggest storing the contract metadata on IPFS

### LINKED ISSUE #239: Suggest storing the contract metadata on IPFS

Currently we suggest storing the contract's TZIP-16 metadata on a separate on-chain contract, however it's getting more common to store it on IPFS for many tokens.

So, we should suggest doing so instead and provide some info in our documentation.

# Acceptance criteria

The documentation suggests and briefly explains how to store the contract metadata on IPFS.

rinn7e:
## Description

Problem: Currently we suggest storing the contract's TZIP-16 metadata
on a separate on-chain contract, however it's getting more common to
store it on IPFS for many tokens.

Solution: Mention IPFS in the specification and readme.

## Related issue(s)

Resolves #239 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,6 +251,19 @@ test-storage:
 		governance_token_address=KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5 \
 		governance_token_id=0n
 
+
+metadata : output = metadata.json
+metadata: test-storage all
+	$(MAKE) -C haskell build PACKAGE=baseDAO-ligo-meta \
+		STACK_DEV_OPTIONS="--fast --ghc-options -Wwarn" \
+	BASEDAO_LIGO_PATH=../$(OUT)/baseDAO.tz \
+	REGISTRY_STORAGE_PATH=../$(OUT)/registryDAO_storage.tz \
+	TREASURY_STORAGE_PATH=../$(OUT)/treasuryDAO_storage.tz
+
+	$(MAKE) -C haskell exec PACKAGE=baseDAO-ligo-meta \
+	EXEC_ARGUMENTS=print-metadata \
+  EXEC_OUTPUT=$(output)
+
 test: test-storage all
 	$(MAKE) -C haskell test PACKAGE=baseDAO-ligo-meta \
 	BASEDAO_LIGO_PATH=../$(OUT)/baseDAO.tz \
--- a/docs/building.md
+++ b/docs/building.md
@@ -24,6 +24,40 @@ which will use `ligo compile-contract` and save the result in `out/baseDAO.tz`.
 If you prefer to build it manually, the contract's main entrypoint is
 `base_DAO_contract`, located in [src/base_DAO.mligo](../src/base_DAO.mligo).
 
+## Generating a contract metadata
+
+BaseDAO metadata should be compliant with [TZIP-16](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md).
+
+You can generate the known metadata, but you'll need the `stack` tool installed, see
+[The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/README/) tutorial for instructions on how to obtain it.
+
+```bash
+make metadata --output=metadata.json
+```
+You can then modify the produced `metadata.json` if you want to add more information.
+
+This metadata should be stored either:
+- On IPFS: [Get started with IPFS](https://ipfs.io/)
+- As a separate contract: [The `tezos-storage` URI Scheme](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#the-tezos-storage-uri-scheme)
+
+
+To use this metadata, follow [TZIP-16 Contract storage](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#contract-storage)
+
+Specifically:
+> The encoding of the values must be the direct stream
+of bytes of the data being stored. For instance, an URI starting with `http:`
+will start with the 5 bytes `0x687474703a` (`h` is `0x68`, `t` is `0x74`,
+etc.). There is no implicit conversion to Michelson's binary format (`PACK`) nor
+quoting mechanism.
+
+After the conversion, the bytes could be use in the argument `metadata_map` of the `make <storage>` command
+(See the next section for more detail.)
+```
+make <storage.tz> \
+  ... \
+  metadata_map=(Big_map.literal [("", <metadata-bytes>)])
+```
+
 ## Generating a contract storage
 
 You can use `ligo` to also generate a contract initial storage, using the
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -872,13 +872,12 @@ Parameter (in Michelson):
 
 This contract implements [TZIP-016](https://gitlab.com/tzip/tzip/-/blob/21fb73fe01df8a744c9b03303e3d73b0f2265eb2/proposals/tzip-16/tzip-16.md).
 
-The DAO contract itself won't store the metadata, rather a dedicated contract will contain that.
-Motivation:
-* The baseDAO contract is very large and already over the Tezos hard limits.
-  * Providing the metadata during origination would increase the already excessive origination cost;
-  * Inserting the metadata after origination still requires mechanisms to manage the metadata, increasing the contract size even more.
+The DAO contract itself won't store the metadata, rather a file on IPFS (suggested), a dedicated contract or another external storage will contain that.
 
 ## Deployment with metadata
 
-The deployment of contract with metadata has an extra step: a dedicated contract for carrying metadata has to be originated first.
+The deployment of contract with metadata has an extra step, either:
+  - Uploading the contract metadata to IPFS.
+  - A dedicated contract for carrying metadata has to be originated first.
+
 Then the baseDAO contract should include the reference to a metadata key in the contract in order to be compliant with TZIP-016.
--- a/haskell/Makefile
+++ b/haskell/Makefile
@@ -14,6 +14,11 @@ STACK_DEV_TEST_OPTIONS = --fast --ghc-options -Wwarn
 STACK_CI_TEST_OPTIONS = --fast --ghc-options -Werror
 # Addtional (specified by user) options passed to test executable
 TEST_ARGUMENTS ?= ""
+# Addtional (specified by user) options passed to `exec` command
+EXEC_ARGUMENTS ?= ""
+# Addtional (specified by user) options for specifying output file
+# Mainly used when generating metadata
+EXEC_OUTPUT ?= ""
 # Packages to apply the command (build, test, e.t.c) for.
 PACKAGE = baseDAO templateDAO
 
@@ -28,6 +33,9 @@ build:
 
 test:
 	$(call call_test,"",$(STACK_DEV_TEST_OPTIONS))
+
+exec:
+	stack exec -- $(PACKAGE) $(EXEC_ARGUMENTS) > "../$(EXEC_OUTPUT)"
 
 # Like 'test' command, but enforces dumb terminal which may be useful to
 # workardoung some issues with `tasty`.


rinn7e:
Took me a while to understand how IPFS work and how it is compared to the traditional way (torrent). 

Since it is pretty straightforward to uploaded the file (simply download the client and upload a file), I only mention the IPFS website in the docs instead of writing a full tutorial.

docs/specification.md
```diff
@@ -882,3 +884,8 @@ Motivation:
 
 The deployment of contract with metadata has an extra step: a dedicated contract for carrying metadata has to be originated first.
 Then the baseDAO contract should include the reference to a metadata key in the contract in order to be compliant with TZIP-016.
+
+## Delployment with metadata that is on IPFS
```
sras, line 888: >Delployment

Possible typo.

docs/specification.md
```diff
@@ -882,3 +884,8 @@ Motivation:
...
+
+Before deploying the contract, the metadata should be uploaded to IPFS. Then the baseDAO contract should include the reference to the
+the metadata CID in the contract in order to be compliant with TZIP-016.
```
sras, line 891: > CID

What is this? Wonder if we should include this as an optional argument to storage creation `make` command to optionally include it in storage?

README.md
```diff
@@ -65,6 +65,21 @@ Tests for this contract are implemented in Haskell using the [cleveland framewor
...
+BaseDAO metadata should be compliant with [TZIP-16](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md).
+
+It can be generated via:
+
+```bash
+cd haskell
+stack exec -- baseDAO-ligo-meta print-metadata > metadata.json
+```
+
+This metadata should be stored via:
```
pasqu4le, line 79: `via` and the `As` (or `On`) of the following points sounds incorrect.

Also here we are giving two suggestions, but there are other valid methods.

I suggest rephrasing this as:
```suggestion
This metadata should be stored either:
```

README.md
```diff
@@ -65,6 +65,21 @@ Tests for this contract are implemented in Haskell using the [cleveland framewor
...
+- As a separate contract: [The `tezos-storage` URI Scheme](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#the-tezos-storage-uri-scheme)
+- On IPFS file storage: [Get started with IPFS](https://ipfs.io/)
```
pasqu4le, line 81: Given that this is the main point of this issue/PR, I would put this as the first option.

Also `FS` in `IPFS` stand for `file storage`, so I think this can be rephrased to avoid this repetition.

README.md
```diff
@@ -65,6 +65,21 @@ Tests for this contract are implemented in Haskell using the [cleveland framewor
...
+
```
pasqu4le, line 82: I think one thing is missing here: the step, which is how to use this stored metadata

I think we should point at the right section of the TZIP-16 spec and perhaps even quote the main part:
> the encoding of the values must be the direct stream of bytes of the data being stored.
> For instance, an URI starting with `http:` will start with the 5 bytes `0x687474703a` (`h` is `0x68`, `t` is `0x74`, etc.). 

and after this conversion mention that the `metadata_map` can be used (point at the `building` doc).

README.md
```diff
@@ -65,6 +65,21 @@ Tests for this contract are implemented in Haskell using the [cleveland framewor
 For more detail about the test suite and how to execute it, check out its
 [README](./haskell/test/README.md).
 
+## Contract Metadata
```
pasqu4le, line 68: It's a bit odd that this is here BTW, I think it should be put perhaps before the `Deploying/Originating` section (here it's too late IMO).

Also, this should be linked to in the `building` doc, I think, given that the metadata needs to be provided when building the storage.

In fact, maybe it makes sense to move this section to that document entirely? What do you think?

README.md
```diff
@@ -65,6 +65,21 @@ Tests for this contract are implemented in Haskell using the [cleveland framewor
...
+
+BaseDAO metadata should be compliant with [TZIP-16](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md).
+
+It can be generated via:
+
+```bash
+cd haskell
+stack exec -- baseDAO-ligo-meta print-metadata > metadata.json
+```
```
pasqu4le, line 77: It's not really part of the issue, but I think we should add a `Makefile` target for this and mention that it needs `stack` to be installed first, like tests.

docs/specification.md
```diff
@@ -873,6 +873,8 @@ Parameter (in Michelson):
 This contract implements [TZIP-016](https://gitlab.com/tzip/tzip/-/blob/21fb73fe01df8a744c9b03303e3d73b0f2265eb2/proposals/tzip-16/tzip-16.md).
 
 The DAO contract itself won't store the metadata, rather a dedicated contract will contain that.
+The metadata can also be uploaded to IPFS file storage.
+
 Motivation:
 * The baseDAO contract is very large and already over the Tezos hard limits.
```
pasqu4le, line 879: We forgot to update this, it seems, the contract is no longer over the limit.

docs/specification.md
pasqu4le, line 888: I think that this could be merged with the previous section, after all it's either way about deploying with metadata and the second step (include the reference) is the same in both cases (except for the different URI).

docs/specification.md
```diff
@@ -873,6 +873,8 @@ Parameter (in Michelson):
 This contract implements [TZIP-016](https://gitlab.com/tzip/tzip/-/blob/21fb73fe01df8a744c9b03303e3d73b0f2265eb2/proposals/tzip-16/tzip-16.md).
 
 The DAO contract itself won't store the metadata, rather a dedicated contract will contain that.
+The metadata can also be uploaded to IPFS file storage.
```
pasqu4le, line 876:
```suggestion
The DAO contract itself won't store the metadata, rather a file on IPFS (suggested), a dedicated contract or another external storage will contain that.
```

docs/specification.md
rinn7e: I believe IPFS URI is call CID: https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#the-tezos-storage-uri-scheme

We can already set this via the `metadata_map = Big_map.literal [("", ipfs://{cid})].

docs/specification.md

README.md
rinn7e: I decide to move this section completely to the `building.md`, also applies your suggestion over there. Let me know if this is alright.

docs/specification.md
rinn7e: Make sense, I've merged them now.

docs/specification.md

docs/specification.md
rinn7e: I've removed that part.

README.md
rinn7e: I've added this now, as part of the `building.md`.

docs/specification.md
sras: Thanks.

>We can already set this via the `metadata_map = Big_map.literal [("", ipfs://{cid})].

Does it make sense to include the handling of this in make file is what I am wondering. Basically, an optional `metadata_cid` argument, if provided, should be included as you shared here. 

Not sure if that is worth the trouble though. @pasqu4le ?

pasqu4le: > I believe IPFS URI is call CID:

That doesn't seem to be _quite_ right, AFAICT from the [documentation](https://docs.ipfs.io/how-to/address-ipfs-on-web/#native-urls) the URI is:
> ipfs://{immutable-root}/path/to/resourceA

CID may be a synonym for the `immutable-root` here, but there's also the rest of the path in the URI.

> We can already set this via the `metadata_map = Big_map.literal [("", ipfs://{cid})].

Ah there's a detail here to be careful with, the values for the `metadata_map` need to be `bytes`, so we need an encoding (which is not the same as the `Bytes,pack`/`PACK` conversion, so we cannot do it in LIGO), see [the spec for URI scheme](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#the-sha256-uri-scheme):
> the encoding of the values must be the direct stream of bytes of the data being stored.
> For instance, an URI starting with `http:` will start with the 5 bytes `0x687474703a` (`h` is `0x68`, `t` is `0x74`, etc.). 

> Basically, an optional metadata_cid argument, if provided, should be included as you shared here.

Well, since we still retain using an external contract as an option what we could do is only to make the encoding above.
So (if there is a common linux command for the conversion) we could get the `metadata_uri` string as an option and encode it in the map.

Makefile
```diff
@@ -238,6 +238,9 @@ $(OUT)/treasuryDAO_storage.tz: src/**
 	# See "$(OUT)/treasuryDAO_storage.tz" for compilation result #
 	#
 
+metadata: all
+	stack exec -- baseDAO-ligo-meta print-metadata
```
pasqu4le, line 242: I think this should use `$(MAKE) -C haskell` and not `stack` directly (see `test` below for example)

docs/building.md
```diff
@@ -39,6 +39,45 @@ You can use `ligo` to also generate a contract initial storage, using the
 Following are the intructions to generate the initial storages of the DAO configuration
 provided in this codebase, see the [README](../README.md).
```
pasqu4le, line 40: This seems off as there are no instructions "following" (they are after the `Generating a contract metadata` section now).

I suggest moving the new `Generating a contract metadata` section above this `Generating a contract storage` one instead.

docs/specification.md
```diff
@@ -872,13 +872,12 @@ Parameter (in Michelson):
...
-Motivation:
-* The baseDAO contract is very large and already over the Tezos hard limits.
-  * Providing the metadata during origination would increase the already excessive origination cost;
-  * Inserting the metadata after origination still requires mechanisms to manage the metadata, increasing the contract size even more.
+The DAO contract itself won't store the metadata, rather a file on IPFS (suggested), a dedicated contract or another external storage will contain that.
 
 ## Deployment with metadata
 
-The deployment of contract with metadata has an extra step: a dedicated contract for carrying metadata has to be originated first.
+The deployment of contract with metadata has an extra step either:
```
pasqu4le, line 879:
```suggestion
The deployment of contract with metadata has an extra step, either:
```

docs/building.md
```diff
@@ -39,6 +39,45 @@ You can use `ligo` to also generate a contract initial storage, using the
...
+
+This metadata should be stored either:
+- On IPFS: [Get started with IPFS](https://ipfs.io/)
+- As a separate contract: [The `tezos-storage` URI Scheme](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#the-tezos-storage-uri-scheme)
+
+
+To use this metadata, follow [TZIP-16 Contract storage](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#contract-storage)
+
+Specifically:
+> Unless otherwise-specified, the encoding of the values must be the direct stream
```
pasqu4le, line 63: I think we can remove the "Unless otherwise-specified" as we don't specify a different format here (now we should), so IMO we can just cut the quote to:
```suggestion
> the encoding of the values must be the direct stream
```

rinn7e: Updated.

docs/specification.md
rinn7e: Fixed.

docs/building.md
rinn7e: Make sense, I've updated this.

Makefile
rinn7e: I had to modify `haskell/Makefile` to do this, I pushed the update now.

docs/building.md
```diff
@@ -39,6 +74,9 @@ You can use `ligo` to also generate a contract initial storage, using the
 Following are the intructions to generate the initial storages of the DAO configuration
 provided in this codebase, see the [README](../README.md).
 
+
+### Storing
```
pasqu4le, line 78: I think this is unnecessary now, probably left after the previous change.

docs/building.md
```diff
@@ -31,6 +31,41 @@ make all
...
+
+Specifically:
+> The encoding of the values must be the direct stream
+of bytes of the data being stored. For instance, an URI starting with `http:`
+will start with the 5 bytes `0x687474703a` (`h` is `0x68`, `t` is `0x74`,
+etc.). There is no implicit conversion to Michelson's binary format (`PACK`) nor
+quoting mechanism.
+
+After the conversion, the bytes could be use in the argument `metadata_map` of the `make <storage>` command
+(See [Storing](#storing) section for more detail.)
```
pasqu4le, line 62: Oh I see now that you mention the storing section here, however with the reordering you can point at the `Generating a contract storage` section or even just say "the next section", given that it immediately follows.

docs/building.md
```diff
@@ -31,6 +31,41 @@ make all
 to generate the contract as well as all the example DAOs' initial storage
 defined below.
 
+## Generating a contract metadata
+
+BaseDAO metadata should be compliant with [TZIP-16](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md).
+
+
+To print the metadata you'll need the `stack` tool, see
```
pasqu4le, line 39:
```suggestion
You can generate the known metadata, but you'll need the `stack` tool installed, see
```

docs/building.md
```diff
@@ -31,6 +31,41 @@ make all
...
+[The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/README/) tutorial
+for instructions on how to obtain it.
+
+```bash
+make metadata > metadata.json
+```
+
```
pasqu4le, line 46:
```suggestion

You can then modify the produced `metadata.json` if you want to add more information.

```
I remembered now that we previously had a similar explanation as we only produce **known** parts of the metadata (hence the previous comment as well.

docs/specification.md
pasqu4le: I think we should do that, given that it fits both the "suggestions" that we give for metadata.

To be clear: I don't think we should modify the behavior of `make metadata` as one might want to add info to the produced JSON (see comments below), but I think the 3 storage generation target should do the encoding instead.

docs/building.md
rinn7e: I've push an update on this, as well as handle other changes. Have to update makefile a bit since its output was not as expected.

docs/specification.md
rinn7e: I'm thinking of creating a command (e.g. `make metadata-encoding ipfs_path=..`) when then output `Big_map.literal [("", <some-vytes>]`. Is this what you have in mind?

pasqu4le: :thinking: that would mean that the user will need to copy the result of that call to then be used in the storage `make` call.

I was thinking more of replacing the `metadata_map` with a string argument, e.g. for trivialDAO (the default would be an empty string):
```make
$(OUT)/trivialDAO_storage.tz : metadata_uri = ""
```
and convert it for LIGO directly:
```
          ; metadata_map = (Big_map.literal [(\"\", $(call encode_metadata_uri,$(metadata_uri)))] : metadata_map)\
```
where `encode_metadata_uri` is the encoding function that needs to be added.

pasqu4le (approved):

PR #265, opened 2021-05-25
[#237] Use levels instead of Timstamp for voting period duration

### LINKED ISSUE #237: Stop using time/timestamps and use blocks instead

We have several types based on `timestamp`s that are usually used to track the passing of `period`/`stage`/`cycle` for proposals' life-cycle.

Dealing with these is however not as convenient (arguably also not as precise) as it would be to handle `nat` amounts of blocks, which is now possible thanks to the `LEVEL` instruction.

If we do this switch it will be likely much easier to do it before #171 

# Acceptance criteria

The contract no longer bases his logic on length of time, but amount of blocks instead.

sras:
## Description

Changes the time durations like voting period to be represented as a number of levels. Updates some configuration varaibles, documents and tests.

## Related issue(s)

Resolves #237 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ ifeq ($(OPTIMIZE), true)
 endif
 	#
 
-$(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
+$(OUT)/trivialDAO_storage.tz : current_level = 100n
 $(OUT)/trivialDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/trivialDAO_storage.tz : ledger = []
@@ -64,12 +64,12 @@ $(OUT)/trivialDAO_storage.tz : quorum_threshold = 10n
 $(OUT)/trivialDAO_storage.tz : min_quorum = 1n
 $(OUT)/trivialDAO_storage.tz : max_quorum = 99n
 $(OUT)/trivialDAO_storage.tz : max_votes = 1000n
-$(OUT)/trivialDAO_storage.tz : period = 950400n # 11 days
+$(OUT)/trivialDAO_storage.tz : period = 15840n # 11 days
 $(OUT)/trivialDAO_storage.tz : quorum_change = 5n
 $(OUT)/trivialDAO_storage.tz : max_quorum_change = 19n
 $(OUT)/trivialDAO_storage.tz : governance_total_supply = 1000n
-$(OUT)/trivialDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
-$(OUT)/trivialDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
+$(OUT)/trivialDAO_storage.tz : proposal_flush_level = 36000n # 22 days
+$(OUT)/trivialDAO_storage.tz : proposal_expired_level = 47520n # 33 days
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
@@ -82,7 +82,7 @@ $(OUT)/trivialDAO_storage.tz: src/**
             { address = (\"$(call require_defined,governance_token_address)\" : address) \
             ; token_id = ($(call require_defined,governance_token_id) : nat) \
             } \
-          ; now_val = ($(now_val) : timestamp)  \
+          ; current_level = {blocks = $(current_level)} \
           ; metadata_map = ($(metadata_map) : metadata_map) \
           ; ledger_lst = ($(ledger) : ledger_list) \
           } \
@@ -90,9 +90,9 @@ $(OUT)/trivialDAO_storage.tz: src/**
           { max_quorum = { numerator = (($(max_quorum) : nat) * quorum_denominator)/100n } \
           ; min_quorum = { numerator = (($(min_quorum) : nat) * quorum_denominator)/100n } \
           ; max_votes = ($(max_votes) : nat) \
-          ; period = { length = ($(period) : nat) } \
-          ; proposal_flush_time = ($(proposal_flush_time)  : nat) \
-          ; proposal_expired_time = ($(proposal_expired_time) : nat) \
+          ; period = { blocks = ($(period) : nat) } \
+          ; proposal_flush_level = { blocks = ($(proposal_flush_level) : nat) } \
+          ; proposal_expired_level = { blocks = ($(proposal_expired_level) : nat) }\
           ; fixed_proposal_fee_in_token = ($(fixed_proposal_fee_in_token) : nat) \
           ; quorum_threshold = { numerator = (($(quorum_threshold) : nat) * quorum_denominator)/100n } \
           ; quorum_change = { numerator = (($(quorum_change) : nat) * quorum_denominator)/100n } \
@@ -111,7 +111,7 @@ $(OUT)/registryDAO_storage.tz : slash_scale_value = 1n
 $(OUT)/registryDAO_storage.tz : slash_division_value = 0n
 $(OUT)/registryDAO_storage.tz : min_xtz_amount = 0mutez
 $(OUT)/registryDAO_storage.tz : max_xtz_amount = 100mutez
-$(OUT)/registryDAO_storage.tz : now_val = `date +"%s"`
+$(OUT)/registryDAO_storage.tz : current_level = 100n
 $(OUT)/registryDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/registryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/registryDAO_storage.tz : ledger = []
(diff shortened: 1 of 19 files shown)


src/types.mligo
```diff
@@ -117,10 +117,8 @@ type voter =
   }
 
 // Amount of `seconds` used for lengths of time
```
pasqu4le, line 119: This comment seems outdated.

src/types.mligo
```diff
@@ -117,10 +117,8 @@ type voter =
...
-type seconds = nat
-
-// Length of a 'stage'
-type period = { length : seconds }
+type levels = { levels : nat }
```
pasqu4le, line 120: Is `level` a reserved word?

`levels` sounds incorrect in some uses of the type, maybe we should use `blocks`?

src/proposal.mligo
```diff
@@ -243,32 +246,32 @@ let do_total_vote_meet_quorum_threshold (proposal, store : proposal * storage):
 // Delete a proposal from 'sProposalKeyListSortByDate'
```
pasqu4le, line 246: This comment also should be updated.

src/proposal.mligo
```diff
@@ -221,8 +224,8 @@ let unfreeze_proposer_and_voter_token
 
 
 [@inline]
-let is_time_reached (proposal, sec : proposal * seconds): bool =
-  Tezos.now > proposal.start_date + int(sec)
+let is_level_reached (proposal, target : proposal * levels): bool =
+  Tezos.level > proposal.start_level.levels + target.levels
```
pasqu4le, line 228: I wonder if we should consider using `>=` now :thinking:

haskell/test/Test/Ligo/TreasuryDAO.hs
```diff
@@ -184,10 +183,10 @@ flushXtzTransfer = uncapsNettest $ withFrozenCallStack $ do
         }
 
   -- Advance one voting period to a voting stage.
-  advanceTime (sec 10)
+  advanceLevel 10
   withSender dodOwner2 $ call dodDao (Call @"Vote") [upvote]
   -- Advance one voting period to a proposing stage.
-  advanceTime (sec 10)
+  advanceLevel 10
```
pasqu4le, line 189: Since you are going through the trouble of updating all these, I wonder if we can make an `advanceStage` that can read the `period` length from the `DaoOriginateData` and move as many levels as necessary.

Makefile
```diff
@@ -61,19 +61,19 @@ $(OUT)/trivialDAO_storage.tz : admin_address = tz1QozfhaUW4wLnohDo6yiBUmh7cPCSXE
 $(OUT)/trivialDAO_storage.tz : guardian_address = KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh
 $(OUT)/trivialDAO_storage.tz : governance_token_address = KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5
 $(OUT)/trivialDAO_storage.tz : governance_token_id = 0n
-$(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
+$(OUT)/trivialDAO_storage.tz : now_level = 100n
```
pasqu4le, line 64: maybe `current_level` would be a better name?

Also, remember to update the `building` doc accordingly as well as the spec.

haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -577,7 +577,7 @@ mkStorage admin extra metadata now tokenAddress qt =
         }
     , sTotalSupply = M.fromList [(frozenTokenId, 0)]
     , sFreezeHistory = mempty
-    , sStartTime = arg #now now
+    , sStartLevel = arg #now now
```
pasqu4le, line 580: Same here, with these changes `now` seems a bit out of place.

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -203,7 +203,7 @@ originateLigoDaoWithBalance extra config balFunc = do
         , (#owner .! owner2, #operator .! operator2, #token_id .! unfrozenTokens)
         ]
 
-  now <- getNow
+  now <- getLevel
```
pasqu4le, line 206: Likewise in these uses of the word `now`

src/proposal.mligo
sras: If a proposal is raised in 100th level, and say proposal exipiry is 5 levels, then I think the expectation is that it should be unexpired for 5 levels. So I think it should expire on 106th level. So shouldn't `>` is the right one itself?

haskell/test/Test/Ligo/TreasuryDAO.hs
sras: It seems that voting period is not part of `DaoOriginateData`. Or your suggestion also include adding it in `DaoOriginateData`?

Makefile
sras: Have changed. Updated `building` doc as well.

haskell/src/Ligo/BaseDAO/Types.hs
sras: Changed

haskell/test/Test/Ligo/BaseDAO/Common.hs
sras: Changed

src/types.mligo
sras: Changed

src/types.mligo
sras: Updated

src/proposal/quorum_threshold.mligo
```diff
@@ -49,7 +49,7 @@ let bound_qt (qt, min_qt, max_qt : quorum_fraction * quorum_fraction * quorum_fr
   else qt
 
 [@inline]
-let period_to_cycle (p: nat): nat = (p + 1n) / 2n
+let stage_to_cycle (p: nat): nat = (p + 1n) / 2n
```
pasqu4le, line 52: Oh somehow I got confused here and didn't update it in my PR.

I think we should add a comment and make it more clear:
```suggestion
// Returns the 'cycle' number of which the given 'stage' number is part of.
let stage_to_cycle (p: nat): nat = (p + 1n) / 2n
```

src/types.mligo
```diff
@@ -145,7 +143,7 @@ type proposal =
   // ^ total amount of votes in favor
   ; downvotes : nat
   // ^ total amount of votes against
-  ; start_date : timestamp
+  ; start_level : blocks
   // ^ time of submission, used to order proposals
```
pasqu4le, line 147:
```suggestion
  // ^ block level of submission, used to order proposals
```

src/proposal.mligo
pasqu4le: :thinking: the way we define it in the config:
> Determine the minimum number of levels after the proposal is proposed to be considered as expired.

I think you are right, as this reads to me as "_after_ 5 level from the one it was created in, this is expired" :ok_hand:

haskell/test/Test/Ligo/TreasuryDAO.hs
pasqu4le: Ah yeah, I didn't check that, but I think `DaoOriginateData` should indeed contain the `period` IMO.

In such case these lines could be changed to:
```suggestion
  advanceLevel dodPeriod
```
instead.

Makefile
sras: Oh. It was missing more updates. Have included it now.

src/proposal.mligo
sras: @pasqu4le,  After giving it more thought, it appears that you maybe correct as well. The question really come down to whether we want to count blocks like we count seconds. If it is like seconds, then we should use `>`. If it is counted as discreete entities (which I think we should), then it should be `>=`.

Either way, I think we need to clarify this in the docs to guard against the possibility of off-by-one errors.

haskell/test/Test/Ligo/TreasuryDAO.hs
sras: Done.

docs/building.md
```diff
@@ -48,19 +48,19 @@ make out/trivialDAO_storage.tz \
...
-  now_val=`date +"%s"` \
+  current_level=100n \
   metadata_map=(Big_map.empty : metadata_map) \
   fixed_proposal_fee_in_token=0n \
   ledger=([] : ledger_list) \
   quorum_threshold=10n \
   min_quorum=1n \
   max_quorum=99n \
-  period=950400n \
+  period={ blocks = 15840n } \
```
pasqu4le, line 58: It looks like this is incorrect, the `Makefile` takes the `nat` as an argument (as it should, IMO)

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -190,7 +192,7 @@ test_RegistryDAO =
             setExtra @Natural [mt|slash_scale_value|] slash_scale_value $
             setExtra @Natural [mt|slash_division_value|] slash_division_value $ initialStorageWithExplictRegistryDAOConfig admin [wallet1]) $
 
-          \(admin: wallet1: _) _ baseDao _ -> let
+          \(admin: wallet1: _) (toVp -> vp) baseDao _ -> let
```
pasqu4le, line 195: same for this `vp` variable here

docs/specification.md
```diff
@@ -152,8 +152,8 @@ type proposal =
   // ^ total amount of votes in favor
   ; downvotes : nat
   // ^ total amount of votes against
-  ; start_date : timestamp
-  // ^ time of submission, used to order proposals
+  ; start_level : blocks
+  // ^ level of submission, used to order proposals
```
pasqu4le, line 156: "level of submission" sounds a bit... wrong? I suggest to use `block level of submission` instead, as in the code comment.

pasqu4le (commented):
Minor points. but the rest LGTM

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -112,22 +111,25 @@ withOriginated addrCount storageFn tests = do
     }
   tests addresses storage (TAddress baseDao) dodTokenContract
 
+toVp :: FullStorage -> Natural
```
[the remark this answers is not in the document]
sras: Changed

haskell/test/Test/Ligo/RegistryDAO.hs
sras: Changed

docs/building.md
sras: Fixed.

docs/building.md
```diff
@@ -48,19 +48,19 @@ make out/trivialDAO_storage.tz \
...
   quorum_change=5n \
   max_quorum_change=19n \
   governance_total_supply=100n \
-  proposal_flush_time=1900801n \
-  proposal_expired_time=2851200n
+  proposal_flush_level={ blocks = 356000n } \
```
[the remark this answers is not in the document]
sras: Fixed.

pasqu4le (approved):

PR #266, opened 2021-05-25
[#250] Rename `rejected_proposal_return_value` and remove cleveland's dependency 

### LINKED ISSUE #250: Consider renaming `rejected_proposal_return_value`

We currently have a `rejected_proposal_return_value` lambda to decide what happens to tokens after a proposal is rejected.

However, the names suggest that the returned value of the function is the amount of tokens to keep, whereas it's actually the amount to slash, so probably `rejected_proposal_slash_value` would be more appropriate.

# Acceptance criteria

- consider renaming `rejected_proposal_return_value` to reflect the slashing
- the documentation correctly reflects how the value returned by that lambda is used

rinn7e:
## Description

Problem: We currently have a `rejected_proposal_return_value` lambda to
decide what happens to tokens after a proposal is rejected.
However, the names suggest that the returned value of the function is the
amount of tokens to keep, whereas it's actually the amount to slash.

Solution: Rename `rejected_proposal_return_value` to
`rejected_proposal_slash_value`.

## Related issue(s)

Resolves #250

- Tests
        silently reappearing again.

- Documentation
    - Haddock


sras (approved):
LGTM

docs/specification.md
```diff
@@ -763,7 +763,7 @@ Parameter (in Michelson):
 - Frozen tokens from voters and proposal submitter associated with those proposals
   are returned in the  form of tokens in governance token contract:
   - If proposal got rejected due to the quorum was not met or the quorum was met but upvotes are less then downvotes:
-    - The return amount for the proposer is equal to or less than the slashed amount based on `rejectedProposalReturnValue`.
+    - The return amount for the proposer is equal to or less than the slashed amount based on `rejectedProposalSlashValue`.
```
pasqu4le, line 766:
```suggestion
    - The return amount for the proposer is equal to or less than the slashed amount based on `rejected_proposal_slash_value`.
```
let's use the name of the `lambda` in LIGO, after all the haskell name is only interesting for us and should not really be part of the spec

pasqu4le (approved):
Only one minor suggestion

PR #267, opened 2021-05-25
[#254] Move tests to submodules

### LINKED ISSUE #254: Move Proposal and Management tests into sub-modules

Both the `Test.Ligo.BaseDAO.Proposal` and `Test.Ligo.BaseDAO.Management` test modules contain _some_ of the definition of the scenarios to run as well as the `TestTree`.

This however results in long compilation times that can get to be quite inconvenient, especially in case of small changes.

Note: some sub-modules already exist, e.g. `Test.Ligo.BaseDAO.Proposal.Vote` for an entrypoint, others might need to be created (possible for a behavior).

Additionally, we should consider renaming `Test.Ligo.BaseDAO.Proposal.Proposal` to `Test.Ligo.BaseDAO.Proposal.Propose`.

# Acceptance criteria

- Scenarios for both proposals and management logic are defined in different sub-modules (existing or new).
- If there are no reason against it `Test.Ligo.BaseDAO.Proposal.Proposal` is renamed to `Test.Ligo.BaseDAO.Proposal.Propose`

sras:
## Description

Includes the following changes.

1. `Test.Ligo.BaseDAO.Proposal.Proposal` is renamed to `Test.Ligo.BaseDAO.Proposal.Propose`
2. Scenarios embedded in `Test.Ligo.BaseDAO.Proposal` is moved to `Test.Ligo.BaseDAO.Proposal.Propose`
3. Scenarios embedded in `Test.Ligo.BaseDAO.Management` is moved to `Test.Ligo.BaseDAO.Management.TransferOwnership`

## Related issue(s)

Resolves #254 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
```diff
@@ -0,0 +1,1119 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Contains tests on @propose@ entrypoin logic for testing the Ligo contract.
```
pasqu4le, line 4: For `Management` I think the change is fine, but regarding `Proposal` and this module, isn't this now the big one with the same very long compilation time?

I think the scenarios here could be moved to different submodules, e.g. at first glance these could be created:
- `Test.Ligo.BaseDAO.Proposal.Quorum` for tests only about the quorum to `propose`
- `Test.Ligo.BaseDAO.Proposal.Tokens` for tests only about the frozen tokens conditions to respect
- `Test.Ligo.BaseDAO.Proposal.Vote` for tests on `vote` only
- etc.

I think this would make it easier to also check that everything is tested and find tests.
e.g. if a condition for `vote` is changed I know that I should add/update a test in `Test.Ligo.BaseDAO.Proposal.Vote`

What do you think?

sras: I have split them into more modules  and have added new modules`Flush`, `Quorum`, `Token` and moved some tests from `Propose` to `Vote` module.

haskell/test/Test/Ligo/BaseDAO/Proposal/Quorum.hs
```diff
@@ -0,0 +1,121 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Contains tests on @propose@ entrypoin logic for testing the Ligo contract.
+module Test.Ligo.BaseDAO.Proposal.Quorum
+  ( proposalIsRejectedIfNoQuorum
+  , proposalSucceedsIfUpVotesGtDownvotesAndQuorum
+  )where
```
pasqu4le, line 8:
```suggestion
  ) where
```

haskell/test/Test/Ligo/BaseDAO/Proposal/Tokens.hs
```diff
@@ -0,0 +1,102 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Contains tests on @propose@ entrypoin logic for testing the Ligo contract.
```
pasqu4le, line 4: This is a copy of the `Propose` sibling module, perhaps it could contain a more detailed description?

P.s. it has the same `entrypoin` typo

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
```diff
@@ -0,0 +1,405 @@
...
```
pasqu4le, line 4:
```suggestion
-- | Contains tests on @propose@ entrypoint logic for testing the Ligo contract.
```

haskell/test/Test/Ligo/BaseDAO/Proposal/Quorum.hs
```diff
@@ -0,0 +1,121 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+-- | Contains tests on @propose@ entrypoin logic for testing the Ligo contract.
```
pasqu4le, line 4: Same here, IMO the description should mention the quorum

pasqu4le (approved):
LGTM now, I only have some minor comment.

Please update the PR description however to reflect the changes.

pasqu4le (approved):

PR #268, opened 2021-05-25
[#257] Prevent empty transactions in transfer proposals

### LINKED ISSUE #257: Prevent empty transactions in transfer proposals

Both `treasuryDAO` and `registryDAO` have proposals that can support xtz transfers.

Both implement `handle_transfer` for `Xtz_transfer_type` in a way that can fail: empty transactions (of `unit` parameter an `0` amount) will be rejected by the network.

We should prevent this by avoiding such transactions before they get executed, in the `proposal_check` lambda of both contract.

Note that in both DAO's `proposal_check` we have a validation of `Xtz_transfer_type` that checks that `min_xtz_amount <= xt.amount`, which makes this unlikely,

However it's possible for `min_xtz_amount` to be `0` in which case the proposal will be accepted and fail during `flush`.

# Acceptance criteria

- Proposal of `Xtz_transfer_type` are rejected in when the `amount` is `0`, even when `min_xtz_amount` is `0`.
- A test is implemented to check on this behavior.

rinn7e:
## Description

Problem: Both `treasuryDAO` and `registryDAO` have proposals
that can support xtz transfers.
Both implement `handle_transfer` for `Xtz_transfer_type` in a way
that can fail: empty transactions (of `unit` parameter an `0` amount)
will be rejected by the network.

Solution: `proposal_check` now reject proposal that contains
`Xtz_transfer_type` with 0 mutez, add tests to cover this.

## Related issue(s)

Resolves #257 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Ligo/RegistryDAO.hs
+++ b/haskell/test/Test/Ligo/RegistryDAO.hs
@@ -146,6 +146,27 @@ test_RegistryDAO =
                baseDao (Call @"Propose") (ProposeParams proposalSize proposalMeta)
                & expectFailProposalCheck baseDao
 
+    , nettestScenarioCaps "proposal_check: fail when xtz transfer contains 0 mutez" $
+        withOriginated 2
+          (\(admin: wallet1:_) ->
+              initialStorageWithExplictRegistryDAOConfig admin [wallet1]
+                & setExtra @Natural [mt|min_xtz_amount|] 0
+          ) $
+          \(_:wallet1:_) _ baseDao _ -> do
+            let proposalMeta = lPackValueRaw @RegistryDaoProposalMetadata $
+                  Transfer_proposal $
+                    TransferProposal 1 [ xtzTransferType 0 wallet1 ] []
+            let proposalSize = metadataSize proposalMeta
+            withSender wallet1 $
+              call baseDao (Call @"Freeze") (#amount .! proposalSize)
+
+            -- Advance one voting period to a proposing stage.
+            advanceTime (sec $ 11)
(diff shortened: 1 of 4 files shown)


sras (approved):
LGTM

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -146,6 +146,24 @@ test_RegistryDAO =
                baseDao (Call @"Propose") (ProposeParams proposalSize proposalMeta)
                & expectFailProposalCheck baseDao
 
+    , nettestScenarioCaps "proposal_check: fail when xtz transfer contains 0 mutez" $
```
pasqu4le, line 149: AFAICT this test (and the same for TreasuryDAO) doesn't actually check that the issue is solved.

Because the `min_xtz_amount` is bigger than `0` here this test will pass even if you revert the changes made in the LIGO source code.

The config needs to be changed for this test to be valid.

I suggest splitting this PR in two commits (in this order):
1. that has the test that fails (proving the check is needed in the code)
2. the commit that updates the LIGO source to fix this

rinn7e: Hmm that's right, I totally miss this, I've split the test and implementation now, also make sure it throws error properly.

pasqu4le: Cool, works well now :ok_hand:

pasqu4le (approved):

PR #269, opened 2021-05-26
[#256] Allow more xtz entrypoints

### LINKED ISSUE #256: Modify entrypoints on which to apply the no-xtz check

We currently have a safety check for most entrypoints to disallow xtz tokens to be mistakenly sent to the contract and allow these funds to be stuck there.

We need to reconsider on which entrypoints we should keep this check and on which we don't need to (see #91).

As a rule of thumb, the distinction should be to keep disallowing these transfers for entrypoints commonly called by users, which may misunderstand the docs and attempt to send tez to.

Namely, they need to keep refusing xtz:
- Vote
- Flush
- Freeze
- Unfreeze 

Other entrypoints allow xtz for a specific reason and have to keep doing so:
- Propose, see #67 
- CallCustom (to allow having _some_ entrypoints to receive xtz if a DAO wants to have that option

The remaining ones can be reconsidered.

# Acceptance criteria

We should decide what to keep in `forbid_xtz_params` and `allow_xtz_params` parts of the overall parameter and then make the changes.

rinn7e:
## Description

Problem: We currently have a safety check for most entrypoints to
disallow xtz tokens to be mistakenly sent to the contract and allow
these funds to be stuck there.
We need to reconsider on which entrypoints we should keep this check
and on which we don't need to (see #91).

Solution: Move `accept_ownership` and `transfer_ownership` to
`allowing_xtz` entrypoints.

## Related issue(s)

Resolves #256 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/src/Ligo/BaseDAO/Types.hs
+++ b/haskell/src/Ligo/BaseDAO/Types.hs
@@ -456,12 +456,10 @@ type UnfreezeParam = ("amount" :! Natural)
 -- issue has been fixed:
 -- https://gitlab.com/morley-framework/morley/-/issues/527
 data ForbidXTZParam
-  = Accept_ownership ()
-  | Call_FA2 FA2.Parameter
+  = Call_FA2 FA2.Parameter
   | Drop_proposal ProposalKey
   | Flush Natural
   | Freeze FreezeParam
-  | Transfer_ownership TransferOwnershipParam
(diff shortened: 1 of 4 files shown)


pasqu4le (approved):
LGTM, my only nitpick is that I would replace `Add` with `Move` in the commit message `Solution`.

Issue #55, opened 2020-11-18
Restrict storage changes in decision lambda


Martoon-00:
# Clarification and motivation

Problem: currently `decisionLambda` can modify the storage arbitrarily, and this potentially can violate the invariants of the contract.

The best way to resolve this is to provide a limited `StorageC` constraint into the lambda (fortunately, it already operates on a polymorphic `store`, not the concrete `Storage`). 

For this, we need Morley to support `StoreHasField` and `StoreHasSubmap` in read-only modes.
Though I think this makes sense doing only if the existing instances will work as-is with the new interface, otherwise the resulting complexity does not worth it.

**Note:** whatever you decide to do, make sure your changes do not conflict with [this MR working on `StoreClass.hs`](https://gitlab.com/morley-framework/morley/-/merge_requests/640) in Morley.

# Acceptance criteria

* Morley is updated to support read-only mode in `StoreHasField` and `StoreHasSubmap`.
* `decisionLambda` provides a reduced `StorageC` where we provide access only to ledger, operators and contract extra.
* Methods like `transfer` and related can be invoked from within the decision lambda.

pasqu4le:
Note: the details of the issue above are specific to the Lorentz version, which will be soon dismissed in #169 

I'm keeping this open because the Ligo version is already more restrictive, but that may need to be reconsidered.

pasqu4le:
Update: we should probably avoid discussing this until after #263 

Currently the contract storage is divided in two major blocks, mostly due to supporting lambdas, which won't be required anymore after that.

If that's indeed the case I suspect that we can move `config` under the `storage` and make it so that what's in that (and only that) can be modified by proposals.

pasqu4le:
As expected, after #263 the section of the storage that can be affected by proposals is much more restricted, see [`decision_callback_output`](`https://github.com/tezos-commons/baseDAO/blob/d095368fc9c9fabd1f9a9bb5903efa50d36798a2/src/types.mligo#L253).

Given that this was made to be as small as possible, I'm closing this issue because no longer relevant.

PR #270, opened 2021-05-26
[#255] Convert integration tests

### LINKED ISSUE #255: Reimplement Proposal integrational tests as nettests

We currently have some integrational tests for quorum threshold and freeze history changes in one module: `Test.Ligo.BaseDAO.Integrational.Proposal`.

At the time these were implemented this was out of necessity, however now that we have the workaround of #214 we should be able to check the quorum threshold and freeze history in nettests as well.

However, due to the freeze history being a big_map, it will only possible to do so in emulated nettests until #31 is solved.

Note: it might be better to base this on the changes for #254 to avoid conflicts

# Acceptance criteria

There are no more integrational tests (or related modules), but nettests only.

sras:
## Description

Problem: Currently some of our tests are written in Integrational
interface, due to having to check storage etc. But we now have some
workaround for this and these tests can be moved to nettest interface.

Solution: Re-write those tests using the nettest interface.

## Related issue(s)

Resolves #255 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -46,7 +47,7 @@ import Ligo.BaseDAO.Types
 import Test.Ligo.BaseDAO.Common.StorageHelper as StorageHelper
 import Test.Ligo.BaseDAO.Proposal.Config (ConfigDesc, fillConfig)
 
-type OriginateFn m = m DaoOriginateData
+type OriginateFn m = QuorumThreshold -> m DaoOriginateData
```
sras, line 50: This was required because there is no `quorumThreshold` field in the `config`, and in tests we use `config` to pass initialization values to the origination function. Till now, we had used a hack by using`minQuorumThreshold` value in config to initialize `quorumThreshold` in storage during contract origination. But this caused some unexpected behavior that I encountered. This caused the quorum theshold update tests to fail, when I moved them to nettest interface, because both the initial quorum threhold and the `minQuorumThreshold` was set to the same value, and thus it could never go below the initial value during updates.

haskell/test/Test/Ligo/BaseDAO/Proposal/Quorum.hs
```diff
@@ -20,17 +25,216 @@ import Test.Ligo.BaseDAO.Proposal.Config
...
+    mulF (QuorumThreshold n1) (QuorumThreshold n2) = QuorumThreshold (div (n1 * n2) (fromIntegral fractionDenominator))
+    divF (QuorumThreshold n1) (QuorumThreshold n2) = QuorumThreshold (div (n1 * (fromIntegral fractionDenominator)) n2)
+    addF (QuorumThreshold n1) (QuorumThreshold n2) = QuorumThreshold (n1 + n2)
+    subF (QuorumThreshold n1) (QuorumThreshold n2) = QuorumThreshold (n1 - n2)
+    fractionToNumerator = mkQuorumThreshold
+
+checkQuorumThresholdDynamicUpdate
+  :: forall caps base m. (MonadEmulated caps base m, HasCallStack)
+  => (ConfigDesc Config -> OriginateFn m) -> m ()
+checkQuorumThresholdDynamicUpdate originateFn = do
```
pasqu4le, line 53: I'm taking this as an example, but it applies to all of the scenarios you converted.

So here you have restricted (by the `` constraint) the type of the scenario to only work on the emulated env and to fetch the value from the storage you do:
```haskell
  storage <- fromVal @FullStorage <$> getStorage' @(ToT FullStorage) dodDao
  let quorumThreshold= sQuorumThresholdAtCycle $ fsStorage storage
```

This works okay, but we have had to deal with this before and used a different solution, which is to define scenarios that are **not** restricted to be emulated by type, but instead accept a function to get the part of the storage that we need (depending if on emulated or network).

Take as example the `flushAcceptedProposals` scenario:
- it uses the more generic `MonadNettest` constraint
- it takes a `GetTotalSupplyFn` because it needs the total supply
- it's used with:
  ```haskell
  nettestScenarioOnEmulatorCaps <...> $ flushAcceptedProposals <...> getTotalSupplyEmulator
  ```

What are the advantages of this approach:
1. We can easily run the same scenario on the network as well (we don't because it takes too long)
   by providing a `GetTotalSupplyFn` that works on the network:
   ```haskell
   nettestScenarioOnNetworkCaps <...> $ flushAcceptedProposals <...> getTotalSupplyNetwork
   ```
2. we don't need to repeat the fetching code, the scenario just uses the `getTotalSupplyFn` it received and the actual implementation for the "get" is contained in the `Test.Ligo.BaseDAO.Common.StorageHelper` module.
   In fact, there you can see that [getFullStorage](https://github.com/tqtezos/baseDAO/blob/master/haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs#L26) has the same definition based on `getStorage'` that you use.

So I propose we do the same for the tests that you added here.

For example, with this scenario we'll need to:
1. add more helpers to `Test.Ligo.BaseDAO.Common.StorageHelper`:
   ```haskell
   ------------------------------------------------------------------------
   -- GetQuorumThresholdAtCycle
   ------------------------------------------------------------------------
   
   type GetQuorumThresholdAtCycleFn m = Address -> m QuorumThresholdAtCycle
   
   getQuorumThresholdAtCycleFnEmulator :: Address -> EmulatedT PureM QuorumThresholdAtCycle
   getQuorumThresholdAtCycleFnEmulator addr =
     sQuorumThresholdAtCycle . fsStorage <$> getFullStorage addr
   ```
2. modify the scenario type to take the helper as an argument:
   ```suggestion
   checkQuorumThresholdDynamicUpdate
     :: (MonadNettest caps base m, HasCallStack)
     => (ConfigDesc Config -> OriginateFn m)
     -> GetQuorumThresholdAtCycleFn m
     -> m ()
   checkQuorumThresholdDynamicUpdate originateFn getQuorumFn = do
   ```
3. use the given function to fetch the storage fieds, so replace:
   ```haskell
     storage <- fromVal @FullStorage <$> getStorage' @(ToT FullStorage) dodDao
     let quorumThresholdActual_ = sQuorumThresholdAtCycle $ fsStorage storage
   ```
   with
   ```haskell
     quorumThresholdActual_ <- getQuorumFn $ unTAddress dodDao
   ```
4. call the scenario with:
   ```haskell
         [ nettestScenarioOnEmulatorCaps "updates quorum-threshold correctly" $
             checkQuorumThresholdDynamicUpdate (originateLigoDaoWithConfigDesc dynRecUnsafe) 
               getQuorumThresholdAtCycleFnEmulator
   ```
What do you say? Does that sound reasonable to you?

sras: Have included this.

haskell/test/Test/Ligo/BaseDAO/Common.hs
pasqu4le: Perhaps we should stop using the config-based approach and instead make up a datatype for configuring in tests :thinking: 

I think we can leave this to be decided in #55

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -5,7 +5,13 @@ module Test.Ligo.BaseDAO.Common.StorageHelper
...
 
   , GetTotalSupplyFn
+  , GetProposalFn
+  , GetFreezeHistoryFn
+  , GetQuorumThresholdAtCycleFn
   , getTotalSupplyEmulator
+  , getQtAtCycleEmulator
+  , getProposalEmulator
+  , getFreezeHistoryEmulator
   -- , getTotalSupplyNetwork
```
pasqu4le, line 15: I think it would be nice to keep these separate by "group" like it was the case for `TotalSupply` and `GetVotePermitCounter` ones before, both here and in the code below.

haskell/test/Test/Ligo/BaseDAO/Proposal/Tokens.hs
```diff
@@ -38,11 +40,39 @@ freezeTokens originateFn = do
...
+checkFreezeHistoryTracking originateFn getFreezeHistory = do
+  DaoOriginateData{..} <- originateFn testConfig defaultQuorumThreshold
+  let frozen_scale_value = 2
+  let frozen_extra_value = 10
+  let proposalMeta1 = ""
+  let proposalSize1 = fromIntegral . BS.length $ proposalMeta1
+  let requiredFrozen = proposalSize1 * frozen_scale_value + frozen_extra_value
+
+  withSender dodOwner1 $ call dodDao (Call @"Freeze") (#amount .! requiredFrozen)
+  advanceLevel 11
```
pasqu4le, line 57:
```suggestion
  advanceLevel dodPeriod
```

haskell/test/Test/Ligo/BaseDAO/Proposal/Tokens.hs
```diff
@@ -38,11 +40,39 @@ freezeTokens originateFn = do
...
+  withSender dodOwner1 $ call dodDao (Call @"Propose") (ProposeParams requiredFrozen proposalMeta1)
+  advanceLevel 11
```
pasqu4le, line 59:
```suggestion
  advanceLevel dodPeriod
```

haskell/test/Test/Ligo/BaseDAO/Proposal/Quorum.hs
```diff
@@ -20,17 +24,219 @@ import Test.Ligo.BaseDAO.Proposal.Config
...
+      ) (mkQuorumThreshold 3 10)
+  let proposer = dodOwner1
+  let dao = dodDao
+  let requiredFrozen = (10 :: Natural)
+  let proposalMeta1 = ""
+
+  withSender proposer $
+    call dao (Call @"Freeze") (#amount .! (2 * requiredFrozen))
+
+  advanceLevel 11
```
pasqu4le, line 70: Can you please change all these in this module to
```suggestion
  advanceLevel dodPeriod
```
(or multiples of it)?

pasqu4le (approved):
Only nitpicks, but LGTM

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
sras: Have changed

haskell/test/Test/Ligo/BaseDAO/Proposal/Tokens.hs
sras: Done. Had to fix in a lot of places, I think these changes got lost during a rebase.

pasqu4le: IIRC it was mostly the opposite, when you opened this PR we hadn't added `dodPeriod` yet and there was no conflict during rebase, so they just didn't pop out.

I had noticed before, but forgot to comment about them :slightly_smiling_face:

PR #273, opened 2021-05-28
[#260] Throw better errors in `proposal_check`, add more tests

### LINKED ISSUE #260: Check all failure scenarios of proposal_check

Both `registryDAO` and `treasuryDAO` have a `proposal_check` lambda that is meant to check some conditions to approve/discard submitted proposals.

Tests on such checks however are a bit lacking and don't cover all the cases of possible failure and corner cases (including possible lack of failures).

# Acceptance criteria

All possible failure condition of the `proposal_check` for `registryDAO` as well as `treasuryDAO` are covered by a test.

rinn7e:
## Description

Problem: Both `registryDAO` and `treasuryDAO` have a `proposal_check`
lambda that is meant to check some conditions to approve/discard
submitted proposals.
Tests on such checks however are a bit lacking and don't cover all the
cases of possible failure and corner cases (including possible lack of failures).

Solution: Update `proposal_check` to throw clearer error messages, add
more tests to handle all possible cases.

## Related issue(s)

Resolves #260 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


src/treasuryDAO.mligo
```diff
@@ -23,22 +23,35 @@ let treasury_DAO_proposal_check (params, extras : propose_params * contract_extr
...
-      List.fold is_all_transfers_valid pm.transfers true
-  else
-    false
+  let is_all_transfers_valid (acc, transfer_type: unit * transfer_type) =
+    match transfer_type with
+    | Token_transfer_type _tt -> acc
+    | Xtz_transfer_type xt ->
+        if (xt.amount = 0mutez) then
+          ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
+            ("FAIL_PROPOSAL_CHECK", "Xtz transfer amount cannot be 0") : unit)
```
pasqu4le, line 44: A couple of comments about this (in general, it applies to all of these new errors):

1. We should still use short strings for these errors, there is either way a cost associated with them.
2. This change to `FAIL_PROPOSAL_CHECK` needs to be documented in the `Errors` section of the spec document, same for all these new "options" for the right side of the pair (perhaps it's worth making a separate table?)
3. These can be abstracted to a function, which would be more convenient:
   ```ocaml
   let fail_proposal_check (msg : string) : unit =
     ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
       ("FAIL_PROPOSAL_CHECK", msg) : unit)
   ```

src/registryDAO.mligo
```diff
@@ -54,34 +54,49 @@ let apply_diff_affected
...
+
+  let _ : unit =
+    if (params.frozen_token <> required_token_lock) then
+      ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
+        ("FAIL_PROPOSAL_CHECK", "Incorrect token amounts locked") : unit)
+    else if (proposal_size >= max_proposal_size) then
+       ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
+        ("FAIL_PROPOSAL_CHECK", "Proposal size is bigger than 'max_proposal_size'") : unit)
+    else
+      unit in
```
pasqu4le, line 74: One more comment about these common errors too is that they can be put in a common source file IMO, e.g.:
```ocaml
let err_incorrect_locked_amt = "WRONG_LOCKED"
```

src/registryDAO.mligo
```diff
@@ -54,34 +54,49 @@ let apply_diff_affected
...
+            else if (xt.amount < min_xtz_amount) then
+              ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
+                ("FAIL_PROPOSAL_CHECK", "Xtz transfer amount cannot be smaller than 'min_xtz_amount'") : unit)
+            else if (xt.amount > max_xtz_amount) then
+              ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
+                ("FAIL_PROPOSAL_CHECK", "Xtz transfer amount cannot be bigger than 'max_xtz_amount'") : unit)
+            else
+              acc
+      in
+        List.fold is_all_transfers_valid tp.transfers unit
```
pasqu4le, line 96: I think we should now use `List.iter` instead

src/treasuryDAO.mligo
rinn7e: I've added point 1 and 3 now. 

Related to point 2, since these errors are only related to registry and treasury DAO, do we need to add it to the spec? Or maybe we should add it to `registry.md` and `treasury.md` instead?

src/registryDAO.mligo
rinn7e: Updated

src/registryDAO.mligo
rinn7e: Updated.

src/treasuryDAO.mligo
pasqu4le: Yeah about that, I was thinking of separating them somehow:
> perhaps it's worth making a separate table?

I was thinking of making a list of "known errors", but your idea to use `registry.md` and `treasury.md` is better, let's go with that.

pasqu4le: Yeah about that, I was thinking of separating them somehow:
> perhaps it's worth making a separate table?

I was thinking of making a list of "known errors", but your idea to use `registry.md` and `treasury.md` is better, let's go with that.

src/registryDAO.mligo
```diff
@@ -54,34 +55,44 @@ let apply_diff_affected
...
+        | Token_transfer_type _tt -> unit
+        | Xtz_transfer_type xt ->
+            if (xt.amount = 0mutez) then
+              fail_proposal_check (zero_mutez_err_msg)
+            else if (xt.amount < min_xtz_amount) then
+              fail_proposal_check (too_small_xtz_err_msg)
+            else if (xt.amount > max_xtz_amount) then
+              fail_proposal_check (too_large_xtz_err_msg)
+            else
+              unit
```
pasqu4le, line 90: One more suggestion about these multi-branch if: this implementation is now a bit too expensive as it's based on multiple `fail_proposal_check` this means that there are 2 `PUSH` , 1 `PAIR` and a `FAILWITH` for every branch.

What I suggest is to move the checking to another function that returns a `string option`, e.g. in this case:
```ocaml
let check_xtz_transfer (xt, min_xtz_amount, max_xtz_amount : xtz_transfer * tez * tez) : string option =
  if (xt.amount = 0mutez) then Some zero_mutez_err_msg
  else if (xt.amount < min_xtz_amount) then Some too_small_xtz_err_msg
  else if (xt.amount > max_xtz_amount) then Some too_large_xtz_err_msg
  else None
```
and in turn only call `fail_proposal_check` in case this returns `Some` error message, here:
```suggestion
        | Xtz_transfer_type xt ->
            begin
              match check_xtz_transfer(xt, min_xtz_amount, max_xtz_amount) with
              | Some err_msg -> fail_proposal_check(err_msg)
              | None -> unit
            end
```

One more thing is: in this case you may notice that the check itself is common between `registry` nad `treasury`, so `check_xtz_transfer` could also be moved in the common module, I think.

src/treasuryDAO.mligo
rinn7e: I've added the errors to them now.

src/registryDAO.mligo
rinn7e: I've add `check_xtz_transfer` to a common file, and update the proposal_check now.

src/proposal.mligo
```diff
@@ -59,9 +59,8 @@ let ensure_proposal_is_unique (propose_params, store : propose_params * storage)
 
 [@inline]
 let check_is_proposal_valid (config, propose_params, store : config * propose_params * storage): storage =
```
pasqu4le, line 61: There is a single use of `check_is_proposal_valid` now, in `propose`:
```ocaml
  let store = check_is_proposal_valid (config, param, store) in
```
I would remove this function's declaration and instead replace its use with:
```ocaml
  let _ : unit = config.proposal_check (propose_params, store.extra)
```

src/registryDAO.mligo
```diff
@@ -54,34 +55,41 @@ let apply_diff_affected
...
-    false
+  let pm = unpack_proposal_metadata(params.proposal_metadata) in
+
+  let _ : unit =
+    if (params.frozen_token <> required_token_lock) then
+      fail_proposal_check (incorrect_token_amount_err_msg)
+    else if (proposal_size >= max_proposal_size) then
+      fail_proposal_check (too_large_proposal_err_msg)
+    else
+      unit in
```
pasqu4le, line 73: Just like for `check_xtz_transfer` this too can be moved to a common function in order to avoid calling `fail_proposal_check` twice (as well as not repeat the code for the checks).

haskell/test/Test/Ligo/TreasuryDAO.hs
```diff
@@ -168,11 +166,11 @@ flushXtzTransfer = uncapsNettest $ withFrozenCallStack $ do
   withSender dodOwner1 $ do
   -- due to smaller than min_xtz_amount
     call dodDao (Call @"Propose") (proposeParams 1)
-      & expectCustomErrorNoArg #fAIL_PROPOSAL_CHECK dodDao
+      & expectCustomError #fAIL_PROPOSAL_CHECK dodDao tooSmallXtzErrMsg
```
pasqu4le, line 169: I think a test to check this same error is missing for `RegistryDAO`.

src/common/errors.mligo
```diff
@@ -0,0 +1,35 @@
...
+
+let fail_proposal_check (msg : string) : unit =
+  ([%Michelson ({| { FAILWITH } |} : (string * string) -> unit)]
+    ("FAIL_PROPOSAL_CHECK", msg) : unit)
+
+// Xtz transfer amount cannot be 0
+let zero_mutez_err_msg = "ZERO_MUTEZ"
+
+// Xtz transfer amount cannot be smaller than 'min_xtz_amount'
+let too_small_xtz_err_msg = "TOO_SMALL_XTZ_AMOUNT"
```
pasqu4le, line 17:
```suggestion
let too_small_xtz_err_msg = "LOW_XTZ"
```
just to keep it as short as possible.

src/common/errors.mligo
```diff
@@ -0,0 +1,35 @@
...
+
+// Xtz transfer amount cannot be bigger than 'max_xtz_amount'
+let too_large_xtz_err_msg = "TOO_LARGE_XTZ_AMOUNT"
```
pasqu4le, line 20:
```suggestion
let too_large_xtz_err_msg = "HIGH_XTZ"
```
same here

src/common/errors.mligo
```diff
@@ -0,0 +1,35 @@
...
+
+// Incorrect token amounts locked
+let incorrect_token_amount_err_msg = "INCORRECT_TOKEN_AMOUNT"
+
+// Proposal size is bigger than 'max_proposal_size'
+let too_large_proposal_err_msg = "TOO_LARGE_PROPOSAL"
```
pasqu4le, line 26:
```suggestion
let too_large_proposal_err_msg = "LARGE_PROPOSAL"
```

src/common/errors.mligo
```diff
@@ -0,0 +1,35 @@
...
+let zero_mutez_err_msg = "ZERO_MUTEZ"
+
+// Xtz transfer amount cannot be smaller than 'min_xtz_amount'
+let too_small_xtz_err_msg = "TOO_SMALL_XTZ_AMOUNT"
+
+// Xtz transfer amount cannot be bigger than 'max_xtz_amount'
+let too_large_xtz_err_msg = "TOO_LARGE_XTZ_AMOUNT"
+
+// Incorrect token amounts locked
+let incorrect_token_amount_err_msg = "INCORRECT_TOKEN_AMOUNT"
```
pasqu4le, line 23:
```suggestion
let incorrect_token_amount_err_msg = "WRONG_TOKEN_AMOUNT"
```
P.s. if you make any of these changes remember to update the `registry` and `treasury` `.md` docs

docs/registry.md
```diff
@@ -71,6 +71,18 @@ It simply applies all updates from the accepted proposal one by one.
...
+
+| Error                                | Description                                                                                                 |
+|--------------------------------------|-------------------------------------------------------------------------------------------------------------|
+| `ZERO_MUTEZ`                         | Xtz transfer amount cannot be 0                                                                             |
+| `TOO_SMALL_XTZ_AMOUNT`               | Xtz transfer amount cannot be smaller than 'min_xtz_amount'                                                 |
+| `TOO_LARGE_XTZ_AMOUNT`               | Xtz transfer amount cannot be bigger than 'max_xtz_amount'                                                  |
+| `INCORRECT_TOKEN_AMOUNT`             | Incorrect token amounts locked                                                                              |
+| `TOO_LARGE_PROPOSAL`                 | Proposal size is bigger than 'max_proposal_size'                                                            |
+
 ## Custom entrypoints
```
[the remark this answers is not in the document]
rinn7e: Make sense, I've updated this now.

pasqu4le (approved):
LGTM now

PR #275, opened 2021-05-28
[#262] Remove FA2 supports 

### LINKED ISSUE #262: Consider removing FA2 compliance from the contract

After the BYO changes, now the DAO relies on an external FA2 governance token, whereas in itself it only stores the froze token and possibly custom additional ones.

However the custom additional tokens are not used by the two main implementations (both `registryDAO` and `treasuryDAO`) and are likely not very useful to most custom DAOs (if necessary, one can always use other external/existing FA2 contracts, as for the governance one).

Moreover, taking in consideration that frozen tokens:
- cannot be `transfer`ed
- have to be stored in a `ledger` for `balance_of` and `update_operator` to work (for the FA2 compliance)
- have much custom logic that don't apply to other tokens (e.g. they have to remember in which stage the tokes were frozen to know in which they can be used)

this causes a lot of inefficiency, as the storage requires "normal" token updates for FA2 as well as "special" info for the custom logic.

Given that both `transfer` and `update_operator` (as a consequence) are not really usable, this seems a big and unnecessary inefficiency, which could perhaps be removed.

# Acceptance criteria

1. consider and discuss dropping FA2 support in the DAO itself
2. if an agreement is reached, make the necessary changes.

rinn7e:
## Description

Problem: After the BYO changes, now the DAO relies on an external
FA2 governance token, whereas in itself it only stores the froze
token and possibly custom additional ones.
However the custom additional tokens are not used by the two main
implementations (both `registryDAO` and `treasuryDAO`) and are likely
not very useful to most custom DAOs (if necessary, one can always use
other external/existing FA2 contracts, as for the governance one).

Solution: Remove FA2 entrypoints, remove `ledger` and `total_supply`,
handle all the affected changes, update the tests.

## Related issue(s)

Resolves #262 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

Problem: When a proposal is rejected, we burn its value in the ledger
and total_supply, but not in freeze_history.

Solution: Update `unstake_tk` to take `burn_amount` into account.


src/proposal.mligo
```diff
@@ -223,16 +214,11 @@ let unfreeze_proposer_and_voter_token
...
-  match Map.find_opt store.frozen_token_id store.total_supply with
-  | Some v -> v
-  | None -> ((failwith "BAD_STATE") : nat)
 
 [@inline]
-let do_total_vote_meet_quorum_threshold (proposal, store : proposal * storage): bool =
+let do_total_vote_meet_quorum_threshold (proposal, config: proposal * config): bool =
   let votes_placed = proposal.upvotes + proposal.downvotes in
-  let total_supply = frozen_total_supply(store) in
+  let total_supply = config.governance_total_supply in
```
rinn7e, line 221: @pasqu4le Took me a while to update the tests, since there are a few strange cases coming up. But they should be all fine now.

One note is that, since we remove `total_supply`, I have to used `governance_total_supply` here. Not sure if this is to be expected.

This PR is pretty much ready for review, only the specification is left. I'll work on that next.

sras: @rinn7e I am a bit doubtful of this change that uses `governance_total_supply` for `frozen_total_supply`.

AFAIU, the governance total supply is the meant to be the total supply of the unfrozen tokens, some of which can be converted to frozen tokens. And the amount of frozen tokens for an account is considered as a measure of how much a voter wants to invest in the DAO.

Also, it does not make sense to use the governace_total_supply as a denominator for calculating the quorum, because AFAIU the denominator should be the maximum possible votes, which is related to the number of frozen tokens, aka, `frozen_total_supply`, and not related to the governance total supply.

@pasqu4le should probably  confirm.

rinn7e: I see, if this is the case, then we should not remove `total_supply`. Let me know if we should put this back in.

src/types.mligo
```diff
@@ -222,13 +174,15 @@ type transfer_ownership_param = address
 type custom_ep_param = (string * bytes)
 
 type propose_params =
-  { frozen_token : nat
+  [@layout:comb]
+  { from_ : address
```
sras, line 178: @rinn7e I didn't get why we need to add the `from_` field here and for `vote_param`, can you please clarify?

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -75,13 +65,25 @@ getProposalEmulator addr pKey =
...
+checkBalanceEmulator :: Address -> Address -> Natural -> EmulatedT PureM ()
+checkBalanceEmulator addr owner bal = do
+  fh <- getFreezeHistoryEmulator addr owner
+  (sumAddressFreezeHistory <$> fh) @== Just bal
+
+
+checkBalanceNetwork :: (Monad m) => Address -> Address -> Natural -> NettestT m ()
+checkBalanceNetwork _addr _owner _bal = do
+  -- We can't check BigMap in network
+  True @== True
```
sras, line 83: I get that we cannot check balance in network, but how does this help?

src/types.mligo
rinn7e: This is related to the new role `delegate` which is similar to FA2 `operator` but instead can call `propose` and `vote` entrypoints. Since we can't use `sender` anymore, we need a new field. More discussion here: https://serokell.slack.com/archives/C019PCGKTK2/p1622222486020100?thread_ts=1621943302.012700&cid=C019PCGKTK2

src/proposal/freeze_history.mligo
```diff
@@ -18,12 +18,10 @@ let stake_frozen_fh (amt, fh : nat * address_freeze_history): address_freeze_his
...
-let unstake_frozen_fh (amt, fh : nat * address_freeze_history): address_freeze_history =
-  match is_nat(fh.staked - amt) with
-  | Some new_amt ->
-      // Adding to past_unstaked should be fine since as of now, the staked
-      // tokens have to be from past stages.
-      { fh with staked = new_amt; past_unstaked = fh.past_unstaked + amt }
+let unstake_frozen_fh (amt_to_unstake, amt_to_burn, fh : nat * nat * address_freeze_history): address_freeze_history =
+  match is_nat(fh.staked - (amt_to_unstake + amt_to_burn)) with
+  | Some new_staked_amt ->
+      { fh with staked = new_staked_amt; past_unstaked = fh.past_unstaked + amt_to_unstake }
```
sras, line 24: May be we should add a test for this.

rinn7e: With the new `checkBalance` function, it already uses `freeze_history` to check the balance, which means it covers `unstake_frozen_fh` as well. https://github.com/tqtezos/baseDAO/pull/275/files#diff-d519aedaf6d7160f71739847f8cdfe0224fdee4cd758cd76e5d56df3e62215f3R86

Or maybe you have some other tests in mind, could suggest to me as well.

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
rinn7e: This allows using the same test case, and run it in both network and emulator:

https://github.com/tqtezos/baseDAO/pull/275/files#diff-bce87f85bb8daa411eca6841102ce815e6bccc0b9977a22c827f936f12559810R124-R129

I could also just use `(\ _ _ _ -> pure ())`, but it is mainly for consistency purpose.

sras: @rinn7e But AFAIU it does not check anything, so wouldn't it make the test useless?

src/proposal/freeze_history.mligo
sras: I see. AFAIU, it now check the freeze balance by summing all the field in `address_freeze_history`, so cannot detect if two field values are swapped (because the sum will remain the same). May be it is good enough. I ll leave it up to you.

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
rinn7e: Using this test as example: https://github.com/tqtezos/baseDAO/pull/275/files#diff-8ce82adc395e63130a569731f81e84d545f091f9038d4bb597c84aed02e0c221R28-R38

Basically, this test does a bunch of things including checkBalance. When running in network, it just skip the checkBalance, but it still run the rest of the actions.

sras: Ok, may be we should rename the function to something like "checkBalanceNetworkDummy", or use 

>  (\ _ _ _ -> pure ())

inline.

src/proposal.mligo
pasqu4le: Yeah, @sras is correct, we cannot use the `governance_total_supply` for that we need to store the `total_supply` indeed.

However keep in mind that we only need to store that for frozen tokens, so it can be a single `nat` in the storage, I suggest naming it `frozen_total_supply`.

src/proposal.mligo
```diff
@@ -47,12 +46,40 @@ let ensure_proposing_stage(stage_num, store : nat * storage): storage =
...
   if Map.mem proposal_key store.proposals
     then (failwith("PROPOSAL_NOT_UNIQUE") : proposal_key)
     else proposal_key
 
+// -----------------------------------------------------------------
+// Delegate
+// -----------------------------------------------------------------
+
+[@inline]
+let check_delegate (from_, author, store : address * address * storage): address =
```
pasqu4le, line 60: A comment above this function could be useful to tell what it does, for example without looking at the code it's not clear what address is returned.

src/proposal.mligo
```diff
@@ -47,12 +46,40 @@ let ensure_proposing_stage(stage_num, store : nat * storage): storage =
   else (failwith("NOT_PROPOSING_STAGE") : storage)
 
 [@inline]
-let ensure_proposal_is_unique (propose_params, store : propose_params * storage): proposal_key =
-  let proposal_key = to_proposal_key(propose_params, Tezos.sender) in
+let ensure_proposal_is_unique (from_, propose_params, store : address * propose_params * storage): proposal_key =
+  let proposal_key = to_proposal_key(propose_params, from_) in
```
pasqu4le, line 50: I think that now that `propose_params` includes the source address we don't need to provide it separately, AFAIU we could change `to_proposal_key` to only have one argument.

src/proposal.mligo
```diff
@@ -69,17 +96,13 @@ let check_proposal_limit_reached (config, store : config * storage): storage =
   then (failwith("MAX_PROPOSALS_REACHED") : storage)
   else store
 
-let freeze_on_ledger (tokens, addr, ledger, total_supply, frozen_token_id, governance_token : nat * address * ledger * total_supply * token_id * governance_token)
-    : (operation * ledger * total_supply) =
+let freeze_on_ledger (tokens, addr, governance_token : nat * address * governance_token)
```
pasqu4le, line 93: 2 suggestions:

1. rename this to `lock_governance_tokens`, since there is no longer a ledger to freeze on and all that this function does now is indeed lock governance tokens into the DAO's account.
2. nitpick: return `operation list` instead (with the only element), it will make this cleaner to use in its only call site

src/proposal.mligo
```diff
@@ -97,19 +120,15 @@ let stake_tk(token_amount, addr, period, store : nat * address * period * storag
   in { store with freeze_history = new_freeze_history; quorum_threshold_at_cycle = {store.quorum_threshold_at_cycle with staked = new_cycle_staked } }
 
 [@inline]
-let unfreeze_on_ledger (tokens, addr, ledger, total_supply, frozen_token_id, governance_token : nat * address * ledger * total_supply * token_id * governance_token): (operation * ledger * total_supply) =
+let unfreeze_on_ledger (tokens, addr, governance_token : nat * address * governance_token): operation =
```
pasqu4le, line 117: Same two suggestions here, except that I would name this `release_governance_tokens`, or maybe `unlock_governance_tokens`.

src/proposal.mligo
```diff
@@ -97,19 +120,15 @@ let stake_tk(token_amount, addr, period, store : nat * address * period * storag
...
+  operation
 
-  // Once this contract is credited on the token contract, we can burn frozen tokens
-  // to debit the `addr` address here.
-  let (ledger, total_supply) = debit_from (tokens, addr, frozen_token_id, ledger, total_supply) in
-  (operation, ledger, total_supply)
-
-let add_proposal (propose_params, period, store : propose_params * period * storage): storage =
-  let proposal_key = ensure_proposal_is_unique (propose_params, store) in
+let add_proposal (from_, propose_params, period, store : address * propose_params * period * storage): storage =
```
pasqu4le, line 124: Why is `from_` needed? Can't we use `propose_param.from_`?

src/proposal.mligo
```diff
@@ -193,28 +208,27 @@ let unfreeze_proposer_and_voter_token
   (rejected_proposal_slash, is_accepted, proposal, period, fixed_fee, store :
     (proposal * contract_extra -> nat) * bool * proposal * period * nat * storage): storage =
   // unfreeze_proposer_token
-  let (tokens, store) =
+  let (tokens, burn_amount, store) =
```
pasqu4le, line 205: `store` is no longer needed here, since it doesn't get updated anymore.

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
rinn7e: I've updated this now.

Makefile
```diff
@@ -58,8 +58,8 @@ endif
 
 $(OUT)/trivialDAO_storage.tz : current_level = 100n
 $(OUT)/trivialDAO_storage.tz : metadata_map = Big_map.empty
+$(OUT)/trivialDAO_storage.tz : freeze_history = Big_map.empty
```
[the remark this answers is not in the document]
rinn7e: Make sense, I've updated this makefile.

src/types.mligo
```diff
@@ -222,13 +174,15 @@ type transfer_ownership_param = address
...
+  ; frozen_token : nat
   ; proposal_metadata : proposal_metadata
   }
 
 type vote_param =
-  [@layout:comb]
-  { proposal_key : proposal_key
+  { from_ : address
```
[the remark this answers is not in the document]
rinn7e: Updated.

src/proposal/freeze_history.mligo
```diff
@@ -18,12 +18,10 @@ let stake_frozen_fh (amt, fh : nat * address_freeze_history): address_freeze_his
...
   | None ->
       ([%Michelson ({| { FAILWITH } |} : (string * unit) -> address_freeze_history)]
         ("NOT_ENOUGH_STAKED_TOKENS", ()) : address_freeze_history)
```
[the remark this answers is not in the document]
rinn7e: Updated.

src/proposal.mligo
```diff
@@ -47,12 +46,40 @@ let ensure_proposing_stage(stage_num, store : nat * storage): storage =
...
+      { owner = Tezos.sender
+      ; delegate = param.delegate
+      } in
+  let updated_delegates = Big_map.update key delegate_update store.delegates
+  in  { store with
+        delegates = updated_delegates
+      }
+
+let update_delegates (params, config, store : update_delegate_params * config * storage): return =
+  (nil_op, List.fold update_delegate params store)
```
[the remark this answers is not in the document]
rinn7e: Make sense, I've updated this.

src/proposal.mligo
rinn7e: Added a comment on this.

src/proposal.mligo
rinn7e: I've updated this now.

src/proposal.mligo
rinn7e: Renamed this as well as the `unlock_governance_tokens`

src/proposal.mligo
rinn7e: Not really needed, I've removed this now.

src/proposal.mligo
rinn7e: Updated.

src/proposal.mligo
rinn7e: I have added `frozen_total_supply` back now.

Makefile
```diff
@@ -58,8 +58,8 @@ endif
```
pasqu4le, line 58: @rinn7e the metadata generation need to be updated in the Haskell code.

Running `make metadata` now I see several things to remove:
1. `TZIP-12` is still listed as a supported interface
2. The `all_tokens` view, which is FA2-specific
3. The `token_metadata` view, which is also FA2-specific

docs/specification.md
```diff
@@ -15,22 +15,15 @@ SPDX-License-Identifier: LicenseRef-MIT-TQ
...
-These two parts are coupled into one smart contract to reduce interactions
-between smart contracts, which is expensive and hard to get right in Tezos.
-However, an existing separate FA2 contract is used for governance.
+An existing separate FA2 contract is used for governance.
 
 # General Requirements
 
-- The contract must be FA2 compatible.
-
 - The contract must store frozen tokens (`token_id` is 0).
```
pasqu4le, line 25: There is no `token_id` anymore.

docs/specification.md
```diff
@@ -197,14 +190,13 @@ These values are:
...
+- The contract maintains a map of addresses and its balance (frozen tokens).
 - The contract maintains the address of an a FA2 contract and a `token_id`,
   to use in the governance process.
-- The contract manages two special roles, the `admin` and `guardian`.
+- The contract manages three special roles, the `admin`, `guardian`, and `delegate`.
 - The contract stores a list of proposals that can be in one of the states:
-  "ongoing", "rejected", "accepted" or "expired".
-- The contract forbids transferring XTZ to the contract,
-  because they might be locked forever.
+  "pending", "ongoing", and "expired".
```
pasqu4le, line 198: AFAIK the proposal states have not changed, what is `pending`?

docs/specification.md
```diff
@@ -219,12 +211,11 @@ These roles apply to the whole contract (hence "global"):
   - Can drop any proposal at any time.
   - Cannot be an implicit address, in other words it must be a contract.
   - There always must be exactly one `guardian`.
+* **delegate**
```
pasqu4le, line 214: This shouldn't be in this list of "global" role, it's explanation I think should be more similar to the removed one for the `operator`

haskell/test/Test/Ligo/BaseDAO/Proposal/Delegate.hs
```diff
@@ -0,0 +1,69 @@
...
+
+  withSender dodOwner1 $
+    call dodDao (Call @"Update_delegate") [param False]
+
+  withSender dodOperator1 $ call dodDao (Call @"Vote") [vote]
+    & expectCustomErrorNoArg #nOT_DELEGATE dodDao
+  advanceLevel dodPeriod
+  withSender dodOperator1 $ call dodDao (Call @"Propose") (params 2)
+    & expectCustomErrorNoArg #nOT_DELEGATE dodDao
+
```
pasqu4le, line 69: A couple more things that would be nice to check:
1. that one can update multiple delegates
2. that (since it is a list) one can add and remove a delegate in the same call without an error
3. that one can remove a delegate even if it was not set before (since we are permissive about that)

docs/specification.md
```diff
@@ -15,22 +15,15 @@ SPDX-License-Identifier: LicenseRef-MIT-TQ
 
 # Overview
 
-The contract described here consists of two parts:
-- Token functionality ([FA2-based][FA2]).
+The contract described here consists ofs:
```
sras, line 18: >ofs

Possible typo

docs/specification.md
```diff
@@ -219,12 +211,11 @@ These roles apply to the whole contract (hence "global"):
   - Can drop any proposal at any time.
   - Cannot be an implicit address, in other words it must be a contract.
   - There always must be exactly one `guardian`.
+* **delegate**
+  - Must be assigned to a partuclar owner address.
```
sras, line 215: >partuclar

Possible typo

docs/specification.md
rinn7e: Seems like the previous proposal state are out dated. I'm using `pending` here to refer to "not yet voting period", probably have to reword this, hope to get your suggestion as well.

src/proposal.mligo
```diff
@@ -158,57 +185,59 @@ let check_vote_limit_reached
...
+        let new_total_supply =
+          match Michelson.is_nat (store.frozen_total_supply - burn_amount) with
+            Some new_total_supply -> new_total_supply
+          | None -> (failwith("BAD_STATE") : nat) in
+        { store with
+            freeze_history = new_freeze_history
+          ; frozen_total_supply = new_total_supply
+        }
     | None -> ([%Michelson ({| { FAILWITH } |} : (string * unit) -> storage)]
           ("NOT_ENOUGH_STAKED_TOKENS", ()) : storage)
```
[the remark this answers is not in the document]
rinn7e: Updated the code as well as the doc.

src/proposal.mligo
```diff
@@ -48,11 +47,41 @@ let ensure_proposing_stage(stage_num, store : nat * storage): storage =
...
+  let delegate_update =
+    if param.enable
+    then (Some unit)
+    else (None : unit option) in
+  let key: delegate =
+      { owner = Tezos.sender
+      ; delegate = param.delegate
+      } in
+  let updated_delegates = Big_map.update key delegate_update delegates
+  in  updated_delegates
```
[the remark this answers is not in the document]
rinn7e: Fixed.

src/proposal.mligo
```diff
@@ -91,16 +116,18 @@ let stake_tk(token_amount, addr, period, store : nat * address * period * storag
...
-  // Once this contract is credited on the token contract, we can burn frozen tokens
-  // to debit the `addr` address here.
-  let (ledger, total_supply) = debit_from (tokens, addr, frozen_token_id, ledger, total_supply) in
-  (operation, ledger, total_supply)
+  let new_total_supply =
+    match Michelson.is_nat (frozen_total_supply - tokens) with
+      Some new_total_supply -> new_total_supply
+    | None ->
+        ([%Michelson ({| { FAILWITH } |} : string * unit -> nat)]
+          ("BAD_STATE", ())) in
```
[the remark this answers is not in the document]
rinn7e: Updated.

Makefile
rinn7e: I forgot about this, pushed an update now.

docs/specification.md
rinn7e: Removed the `token_id` part.

docs/specification.md
rinn7e: Make sense, I've updated this.

haskell/test/Test/Ligo/BaseDAO/Proposal/Delegate.hs
rinn7e: Added a new test that covers all this cases.

docs/specification.md
rinn7e: Fixed.

docs/specification.md
pasqu4le: I don't think they were really outdated, from their creation onward proposals go through these states:
1. "ongoing" when they have been proposed but can't be voted on yet (I guess we can rename it to "proposed")
2. "ongoing" when it's getting voted on
3. "ongoing" if it's no longer time to vote on it, but it hasn't been flushed yet (I guess this can be called "pending flush")
3. "accepted" or "rejected" when they are flushed, depending on the result
4. "expired" if they haven't been flushed in time

If you agree I think we can make this more clear here.

pasqu4le (approved):
needs a rebase , but otherwise LGTM now

docs/specification.md
rinn7e: I've updated it as the follow:
```
- The contract stores a list of proposals that can be in one of the states:
  "proposed", "ongoing", "pending flush", "accepted, "rejected", or "expired".
```

Let me know if this is good enough. I also rebased and squashed the commits now.

pasqu4le: Yep, sounds good to me, except that you forgot a double quote after `accepted`. it should be:
```
"proposed", "ongoing", "pending flush", "accepted", "rejected", or "expired".
```

The rebase also looks good to me :+1:

Issue #271, opened 2021-05-26
Verify the limit on `max_votes` and what to suggest for its default value


pasqu4le:
# Clarification and motivation

Sibling issue to #68 (or rather its solution, but for `max_votes`.

We currently suggest a number of `max_votes` of `1000`, but in most cases that might be too small.

However, a limit should be put, mainly because the number of votes affects the size of the `voters` field in a `proposal`, which has to be de-serialized for `flush`, which means it can get to be too large to fit in a single operation's gas cost and hit a constant failure.

While this problem is not entirely catastrophic (such a proposal could be dropped) we should still verify that `1000` is a sensible value to suggest and if possible perhaps suggest an even higher number (and limit it in storage creation).

# Acceptance criteria

1. Some evidence is given that either:
    - no limit is actually needed on `max_votes`
    - there is a number `x` that is a hard limit for `max_votes`
2. In case there is an `x`:
    - add a check in `defaults.mligo` to prohibit more than `x` (or even less, for safety) for `max_votes`
    - suggest a number of `max_votes` that's closer to `x/2`

sras:
It appears that changing the `voters` field in the proposal from a `list` to a `map` might help to reduce the size if there are multiple votes from an address. So we would like to do that as well in the context of this issue.

sras:
Since we have changed the `proposal` to track voters instead of votes, this issue should instead track the number of voters, instead of votes. 

It has been confirmed that a proposal that contain 1000 voters can continue to operate with out issues. After sending 1000 votes to the contract, each from a different voter address, It has been tested that the contract can successfully unpack the proposal in context (one with 1000 voters) and other operations like freeze.

To do this test, the contract was slightly modified temporarly to extend the potentioal voting period indefinitely, instead of restricting it to a very specific period, and the test was done on this modified contract (or else the period would pass when sending the 1000 separate votes). 

The code for the test is preserved in branch sras/#271-stress-test-voters-do-not-merge

pasqu4le:
@sras to avoid changing the contract I think that here too we could instead use something more like [this](https://github.com/tezos-commons/baseDAO/pull/172#discussion_r599522411), which is to say to originate the contract with a large number of `voters` in its initial storage instead.

It should be just as effective and much less painful, don't you think?

pasqu4le:
I'm closing this because #311 removed the limit in question, making this issue obsolete.

PR #278, opened 2021-06-01
[#277] Use max voters instead of max votes

### LINKED ISSUE #277: Replace `max_vote` with `max_voters`

Currently we have a limit on the number of votes that could be made on a proposal.
Given that:
- when possible we'd want all the possible votes to be cast
- the number of total votes is limited by the governance token (because 1 vote per token)

the only reason for this limit seems to be to avoid the issue explained in #271

However, as that explains, the problem is the potential size of the `voters` field in a `proposal` and not really the number of cast votes.

So we should switch from checking for a `max_votes` to checking for a `max_voters`, aka we should limit the `voters` field to get larger than a set value and not the `upvotes + downvotes`.

# Acceptance criteria

The contract's proposals are no longer limited by number of `votes`, but rather by number of `voters`

sras:
## Description

Changes max_votes to max_voters. So the number of voters who has casted a vote on a proposal in counted instead
of the total number of votes. This is done to limit the size of the voters map (for which we will find a sensible default value in #271).

## Related issue(s)

Resolves #277 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -101,8 +101,9 @@ type config =
 
   ; max_proposals : nat
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
-  ; max_votes : nat
-  // ^ Determine the maximum number of votes associated with a proposal.
+  ; max_voters : nat
+  // ^ Determine the maximum number of voters that are allowed to vote on a proposal.
+  // Voters who votes on a proposal, both ways are counted twice.
   ; max_quorum_threshold : quorum_fraction
   // ^ Determine the maximum value of quorum threshold that is allowed.
   ; min_quorum_threshold : quorum_fraction
@@ -299,6 +300,11 @@ Moreover the proposal to vote on must have been submitted in the proposing `stag
 immediately preceding and the voter must have frozen his tokens in one of the
 preceding `stage`s.
 
+Each vote stakes one frozen token. Staked tokens cannot be unfreezed till they
+are unstaked when the associated proposal is flushed. The number of staked tokens
+only depend on the number of votes, and does not depend on whether the vote is in
+favor or against a proposal.
+
 It's possible to vote positively or negatively.
 After the voting ends, the contract is "flushed" by calling a dedicated entrypoint.
 
@@ -335,7 +341,7 @@ The list of errors may be inaccurate and incomplete, it will be updated during t
 | `QUORUM_NOT_MET`                     | A proposal is flushed, but there are not enough votes                                                                    |
 | `VOTING_STAGE_OVER`                  | Throws when trying to vote on a proposal that is already ended                                                           |
 | `MAX_PROPOSALS_REACHED`              | Throws when trying to propose a proposal when proposals max amount is already reached                                    |
-| `MAX_VOTES_REACHED`                  | Throws when trying to vote on a proposal when the votes max amount of that proposal is already reached                   |
+| `MAX_VOTERS_REACHED`                 | Throws when trying to vote on a proposal when the max voter count of that proposal is already reached                   |
 | `FORBIDDEN_XTZ`                      | Throws when some XTZ was received as part of the contract call                                                           |
 | `PROPOSER_NOT_EXIST_IN_LEDGER`       | Expect a proposer address to exist in Ledger but it is not found                                                         |
 | `PROPOSAL_NOT_UNIQUE`                | Trying to propose a proposal that is already existed in the Storage.                                                     |
@@ -739,7 +745,7 @@ Parameter (in Michelson):
   from past stages that is not staked is less than specified `vote_amount` .
 - Fails with `PROPOSAL_NOT_EXIST` if the proposal key is not associated with any ongoing proposals.
 - Fails with `VOTING_STAGE_OVER` if the voting `stage` for the proposal has already ended.
-- Fails with `MAX_VOTES_REACHED` if the amount of votes of the associated proposal
+- Fails with `MAX_VOTERS_REACHED` if the voter count of the associated proposal
   is already at the max value set by the configuration.
 - Fails with `MISSIGNED` if permit is incorrect with respect to the provided vote parameter and contract state.
 - The entrypoint accepts a list of vote params. As a result, it is possible to
--- a/src/defaults.mligo
+++ b/src/defaults.mligo
@@ -11,12 +11,6 @@ let validate_proposal_flush_expired_level (data : initial_config_data) : unit =
     failwith("proposal_flush_level needs to be more than twice the voting_period length")
   else unit
 
-let validate_max_votes (data : initial_config_data) : unit =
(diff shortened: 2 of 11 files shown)


docs/specification.md
```diff
@@ -101,8 +101,8 @@ type config =
 
   ; max_proposals : nat
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
-  ; max_votes : nat
-  // ^ Determine the maximum number of votes associated with a proposal.
+  ; max_voters : nat
+  // ^ Determine the maximum number of voters that are allowed to vote in the context of a proposal.
```
pasqu4le, line 105:
```suggestion
  // ^ Determine the maximum number of voters that are allowed to vote on a proposal.
```
Nitpick, but this seems simpler.

docs/specification.md
```diff
@@ -335,7 +335,7 @@ The list of errors may be inaccurate and incomplete, it will be updated during t
 | `QUORUM_NOT_MET`                     | A proposal is flushed, but there are not enough votes                                                       |
 | `VOTING_STAGE_OVER`                  | Throws when trying to vote on a proposal that is already ended                                              |
 | `MAX_PROPOSALS_REACHED`              | Throws when trying to propose a proposal when proposals max amount is already reached                       |
-| `MAX_VOTES_REACHED`                  | Throws when trying to vote on a proposal when the votes max amount of that proposal is already reached      |
+| `MAX_VOTERS_REACHED`                 | Throws when trying to vote on a proposal when the votes max amount of that proposal is already reached      |
```
pasqu4le, line 338: The description was not updated

docs/specification.md
```diff
@@ -738,7 +738,7 @@ Parameter (in Michelson):
   from past stages that is not staked is less than specified `vote_amount` .
 - Fails with `PROPOSAL_NOT_EXIST` if the proposal key is not associated with any ongoing proposals.
 - Fails with `VOTING_STAGE_OVER` if the voting `stage` for the proposal has already ended.
-- Fails with `MAX_VOTES_REACHED` if the amount of votes of the associated proposal
+- Fails with `MAX_VOTERS_REACHED` if the amount of votes of the associated proposal
```
pasqu4le, line 741:
```suggestion
- Fails with `MAX_VOTERS_REACHED` if the amount of voters of the associated proposal
```

src/defaults.mligo
```diff
@@ -27,15 +21,14 @@ let validate_quorum_threshold_bound (data : initial_config_data) : unit =
 
 let default_config (data : initial_config_data) : config =
   let _ : unit = validate_proposal_flush_expired_level(data) in
-  let _ : unit = validate_max_votes(data) in
```
pasqu4le, line 30: I see you (correctly) removed this check, however can you please still leave a TODO for `#271`, since the problem remains (we need to find safe values for this option)?

docs/specification.md
pasqu4le: Also I think that here it's worth adding a note that specifies that voters that votes on the same proposal in both ways are counted twice, what do you think?

src/types.mligo
```diff
@@ -110,12 +110,12 @@ type nonce = nat
...
 // Voter info for a proposal
-type voter =
-  { voter_address : address
-  ; vote_amount : nat
-  ; vote_type : vote_type
+type vote_item =
+  { vote_amount : int
   }
 
+type voter_map = (address, vote_item) map
```
pasqu4le, line 117: I suppose this precedes our discussion and it should actually be like:
```ocaml
type vote_amount = nat
type voter =
  { voter_address : address 
  ; vote_type : vote_type
  }

type voter_map = (voter, vote_amount) map 
```

docs/specification.md
sras: Have updated.

docs/specification.md
sras: Oh, I think that is restriction will appear to be a bit arbitrary. Can we count them only once? It would make `vote` operation more costly. May be we can limit the voter count to `2 * max_voters`, allowing every voter to vote both ways in they choose.

What do you think?

sras: Right now I have modifed the spec to say the voters are counted twice. Will change it if you think otherwise.

src/types.mligo
sras: Have changed this and have added a test to check that the tokens are unstaked correctly if there are multiple votes from a voter.

src/proposal.mligo
```diff
@@ -136,29 +136,25 @@ let add_proposal (propose_params, period, store : propose_params * period * stor
...
-      voters = voter :: proposal.voters
-    } in
+        let map_key = (author, vote_param.vote_type) in
+        let vote_item = match Map.find_opt map_key proposal.voters with
+          | Some vote_item -> { vote_item with vote_amount = vote_item.vote_amount + vote_param.vote_amount }
+          | None -> { vote_amount = vote_param.vote_amount } in
+        let voters = Map.add map_key vote_item proposal.voters
+        in match vote_param.vote_type with
+          | true -> { proposal with upvotes = vote_item.vote_amount; voters = voters  }
+          | false -> { proposal with downvotes = vote_item.vote_amount; voters = voters  } in
```
pasqu4le, line 147: A couple of things about this:
1. setting both `voters` and `upvotes`/`downvotes` at the same time is actually more expensive (the code to set `voters` is repeated twice.
2. having this all indented under the `let proposal =` is less readable IMO
3. possibly `vote_item` would be easier to use if it wasn't a record (I don't see the advantage of it being one here, but I'll let you decide).
4. the most important part is that `vote_amount` needs to be added to `upvotes`/`downvotes` not replace it!
    It looks like we don't have a test for this, please add one

This is my suggestion here:
```ocaml
  let voter_map_key = (author, vote_param.vote_type) in
  let vote_item = match Map.find_opt voter_map_key proposal.voters with
    | Some vote_item -> { vote_item with vote_amount = vote_item.vote_amount + vote_param.vote_amount }
    | None -> { vote_amount = vote_param.vote_amount } in
  let voters = Map.add voter_map_key vote_item proposal.voters in
  let proposal = { proposal with voters = voters } in
  let proposal =
    if vote_param.vote_type
    then { proposal with upvotes = proposal.upvotes + vote_param.vote_amount }
    else { proposal with downvotes = proposal.downvotes + vote_param.vote_amount }
  in
```

sras: Sorry about that mixup. 

I have made all the suggestions. It is not exactly as you suggested, because I have ditched `vote_item` and just use `nat`, as it does not made much sense to use the record (It might have added a bit more safety to the `do_unfreeze` call a little down, but it is only one place, so it is not probably a big deal).

sras: Oh, have updated it btw.

pasqu4le (approved):
LGTM now

PR #279, opened 2021-06-01
[#276] throw an error to avoid empty flush calls

### LINKED ISSUE #276: Flush should fail when no proposals are flushed

The conditions under which a proposal may be flushed are complex and users may attempt to flush optimistically.
Currently, flushing will complete successfully and cost gas when flushing zero proposals.

# Acceptance criteria

The `flush` entrypoint should fail with an error when zero proposals are successfully flushed.

pasqu4le:
## Description

Problem: the conditions for a successful flush are complex and user may
try to optimistically call the entrypoint when unnecessary.
This will create no issue, but will result in gas costs that could be
avoided.

Solution: throw an error when flush is called without any available
proposals.

## Related issue(s)

Resolves #276

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e (approved):
Looks good to me.

Issue #258, opened 2021-05-25
Remove cleveland's dependency from the haskell library


pasqu4le:
# Clarification and motivation

Currently the `baseDAO-ligo-meta` `library` has a dependency on `cleveland`.

This is odd (as only the `test-suite` should have it) and the reason this exist is due to the use of `Michelson.Test.Import`, which was the only way to read contract and storage from the files produced by LIGO.

However since [morley/565](https://gitlab.com/morley-framework/morley/-/issues/565) has been solved it's now possible to remove such dependency and only depend on the `morley` library instead.

# Acceptance criteria

The `library` for `baseDAO-ligo-meta` no longer depends on `cleveland`

PR #280, opened 2021-06-07
[#258] Remove cleveland from library

rinn7e:
## Description

Problem: Currently the `baseDAO-ligo-meta` `library` has a dependency on `cleveland`.
This is odd (as only the `test-suite` should have it) and the reason this exist is due
to the use of `Michelson.Test.Import`, which was the only way to read contract and storage
from the files produced by LIGO.
However since [morley/565](gitlab.com/morley-framework/morley/-/issues/565) has
been solved it's now possible to remove such dependency and only depend on the `morley` library instead.

Solution: Bump morley version, remove `cleveland` dependency from library.

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e:
This took quite some times, because I was using the latest commit from morley: https://gitlab.com/morley-framework/morley/-/commit/398d409b3de9cf7f683edfd2e745cc548364f5fe, which results in a more refactoring and some strange behaviour:
- One of it was that, compiling registry dao, take a really long time. The changes can be check in this branch: https://github.com/tqtezos/baseDAO/commit/8a5cc5a313b0ef9b8b227e838cb6b204981e243a

As a result, I just bump morley version slightly (a few commits above the previous), which seems to work fine now.

haskell/src/Ligo/Util.hs
```diff
@@ -20,7 +20,7 @@ import System.Environment (lookupEnv)
 import System.FilePath ((</>))
 
 import Michelson.Parser
-import Michelson.Test.Import -- TODO 258: remove cleveland dependency
+import Michelson.Runtime.Import (readContract)
```
pasqu4le, line 23: I see that you now import `readContract` here, but one more change that we can do is to use `readValue` as well, which was added in the same [morley MR](https://gitlab.com/morley-framework/morley/-/merge_requests/820) and allows us to not need to implement `readTypedValue_` in this module.

Can you please replace that as well?

pasqu4le (commented):
Only one request.

Also can you please remove the marksown link from the commit message? That's not going to be rendered and it makes it much less readable.

haskell/src/Ligo/Util.hs
rinn7e: I've updated the commit message, as well as using the `readValue` instead now.

pasqu4le (approved):
LGTM now

PR #288, opened 2021-06-21
[#284] Restore use of blocks to denote intervals

### LINKED ISSUE #284: Temporarily revert `LEVEL` changes

In #273 we switched the contract from `timestamp`-based to `LEVEL`/ block-based.

This change however is a complication for applications using BaseDAO, as they need to adapt to this change in logic and types.

On the other hand, there are other changes, especially the FA2 removal of #262 which are easy to adapt to and bring a bigger advantage, in the form of a much smaller contract.

This was merged after the `LEVEL` change, so there was never a point in which the contract had no FA2 and was using timestamps at the same time.

For this reason we want to make such a version now.

# Acceptance criteria

Two PRs need to be opened for this issue:
1. which reverts the contract to a time-based logic, but keeps all the other changes
2. which reverts the first PR and goes back to level-based logic once again

sras:
Problem: We had temporarly reverted to using Timestamps for denoting
time intervals. Now we can restore the old behavior.

Solution: Revert commit 9e0c236b9759d4312904dd3af089cd709c20a8d0,
reversing changes made to a21886ed9422ba2e95a95ee23a9aeca95830a58f.

## Related issue(s)

Resolves #284 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,6 @@ ifeq ($(OPTIMIZE), true)
 endif
 	#
 
-$(OUT)/trivialDAO_storage.tz : current_level = 100n
 $(OUT)/trivialDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/trivialDAO_storage.tz : freeze_history = []
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
@@ -64,12 +63,12 @@ $(OUT)/trivialDAO_storage.tz : quorum_threshold = 10n
 $(OUT)/trivialDAO_storage.tz : min_quorum = 1n
 $(OUT)/trivialDAO_storage.tz : max_quorum = 99n
 $(OUT)/trivialDAO_storage.tz : max_voters = 1000n
-$(OUT)/trivialDAO_storage.tz : period = 15840n # 11 days
+$(OUT)/trivialDAO_storage.tz : period = 15840n
 $(OUT)/trivialDAO_storage.tz : quorum_change = 5n
 $(OUT)/trivialDAO_storage.tz : max_quorum_change = 19n
 $(OUT)/trivialDAO_storage.tz : governance_total_supply = 1000n
-$(OUT)/trivialDAO_storage.tz : proposal_flush_level = 36000n # 22 days
-$(OUT)/trivialDAO_storage.tz : proposal_expired_level = 47520n # 33 days
+$(OUT)/trivialDAO_storage.tz : proposal_flush_level = 36000n
+$(OUT)/trivialDAO_storage.tz : proposal_expired_level = 47520n
 $(OUT)/trivialDAO_storage.tz: src/**
 	# ============== Compiling TrivialDAO storage ============== #
 	mkdir -p $(OUT)
@@ -82,7 +81,7 @@ $(OUT)/trivialDAO_storage.tz: src/**
             { address = (\"$(call require_defined,governance_token_address)\" : address) \
             ; token_id = ($(call require_defined,governance_token_id) : nat) \
             } \
-          ; current_level = {blocks = $(current_level)} \
+          ; start_level = {blocks = $(call require_defined,start_level)} \
           ; metadata_map = ($(call escape_double_quote,$(metadata_map)) : metadata_map) \
           ; freeze_history = ($(call escape_double_quote,$(freeze_history)) : freeze_history_list) \
           } \
@@ -111,7 +110,6 @@ $(OUT)/registryDAO_storage.tz : slash_scale_value = 1n
 $(OUT)/registryDAO_storage.tz : slash_division_value = 0n
 $(OUT)/registryDAO_storage.tz : min_xtz_amount = 0mutez
 $(OUT)/registryDAO_storage.tz : max_xtz_amount = 100mutez
-$(OUT)/registryDAO_storage.tz : current_level = 100n
 $(OUT)/registryDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/registryDAO_storage.tz : freeze_history = []
 $(OUT)/registryDAO_storage.tz : fixed_proposal_fee_in_token = 0n
@@ -119,11 +117,11 @@ $(OUT)/registryDAO_storage.tz : quorum_threshold = 10n
 $(OUT)/registryDAO_storage.tz : min_quorum = 1n
 $(OUT)/registryDAO_storage.tz : max_quorum = 99n
 $(OUT)/registryDAO_storage.tz : max_voters = 1000n
-$(OUT)/registryDAO_storage.tz : period = 15840n # 11 days
+$(OUT)/registryDAO_storage.tz : period = 15840n
 $(OUT)/registryDAO_storage.tz : quorum_change = 5n
 $(OUT)/registryDAO_storage.tz : max_quorum_change = 19n
-$(OUT)/registryDAO_storage.tz : proposal_flush_level = 36000n # 22 days
-$(OUT)/registryDAO_storage.tz : proposal_expired_level = 47520n # 33 days
+$(OUT)/registryDAO_storage.tz : proposal_flush_level = 36000n
+$(OUT)/registryDAO_storage.tz : proposal_expired_level = 47520n
 $(OUT)/registryDAO_storage.tz : governance_total_supply = 1000n
 $(OUT)/registryDAO_storage.tz: src/**
 	# ============== Compiling RegistryDAO storage ============== #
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -174,19 +174,19 @@ $(OUT)/treasuryDAO_storage.tz : slash_scale_value = 0n
...
 $(OUT)/treasuryDAO_storage.tz : max_quorum = 99n
 $(OUT)/treasuryDAO_storage.tz : max_voters = 1000n
-$(OUT)/treasuryDAO_storage.tz : period = 950400n # 11 days
+$(OUT)/treasuryDAO_storage.tz : period = 15840n # 11 days
 $(OUT)/treasuryDAO_storage.tz : quorum_change = 5n
 $(OUT)/treasuryDAO_storage.tz : max_quorum_change = 19n
-$(OUT)/treasuryDAO_storage.tz : proposal_flush_time = 1900801n # 22 days 1 seconds
-$(OUT)/treasuryDAO_storage.tz : proposal_expired_time = 2851200n # 33 days
+$(OUT)/treasuryDAO_storage.tz : proposal_flush_level = 36000n # 22 days
+$(OUT)/treasuryDAO_storage.tz : proposal_expired_level = 47520n # 33 days
```
rinn7e, line 189: I think the comment here could be removed, since it is no longer in date, the rest looks good to me.

rinn7e (approved):

src/proposal.mligo
```diff
@@ -237,8 +236,8 @@ let unfreeze_proposer_and_voter_token
   Map.fold do_unfreeze proposal.voters store
 
 [@inline]
-let is_time_reached (proposal, sec : proposal * seconds): bool =
-  Tezos.now > proposal.start_date + int(sec)
+let is_level_reached (proposal, target : proposal * blocks): bool =
+  Tezos.level > proposal.start_level.blocks + target.blocks
```
pasqu4le, line 240: @sras I reconsidered the [previous discussion](https://github.com/tqtezos/baseDAO/pull/265#discussion_r639372733) and I too think we should count level discretely and not as we did with seconds/time.

So I suggest we change this `>` to `>=` here.

In fact I think there's another thing to correct, which is that in `validate_proposal_flush_expired_level` we currently enforce that the `proposal_flush_level` is has to be `> period * 2`, but I think that we should change it to be `>=
period * 2`.

Here's why: when we were dealing with time we introduced that `proposal_flush_time` needed to be more than twice the period length so that the `voting` stage of a proposal had to be ended before it could be flushed.
In other words that limitation made sure that even if a proposal was `propose`d at he very beginning of a `proposing` stage, all that stage and the next one (the `voting`) had to pass before `flush` was possible.

So let's take the implementation after this PR and the most extreme case, in which the `period` has the length of `1` level (it works for bigger periods too tho), what we would like is that it should be possible to configure the contract so that a proposal can get flushed in the first level **after** its voting period ended,
However that's not possible, `validate_proposal_flush_expired_level` now imposes that `proposal_flush_level` cannot be `<=` then `period * 2`, so in our case it needs to be at least `3`.
Moreover, this `is_level_reached` requires the number of level to fully have passed (with `>`).

So, say that with a `period` of `1` I submit a proposal in stage `x`, which is proposing, then:
- I cannot flush it in `x + 1`, which is right because this is its voting period, so this should not be allowed
- I cannot flush it in `x + 2`, which should be possible to configure, but it isn't for `validate_proposal_flush_expired_level`
- I cannot flush it in `x + 3`, which should also be possible to configure, but it isn't for `is_level_reached`

If we make the two changes instead this would be fixed, so here I suggest:
```suggestion
let is_level_reached (proposal, target : proposal * blocks): bool =
  Tezos.level >= proposal.start_level.blocks + target.blocks
```
and ... (see next comment)

src/defaults.mligo
```diff
@@ -4,11 +4,11 @@
...
-let validate_proposal_flush_expired_time (data : initial_config_data) : unit =
-  if data.proposal_expired_time <= data.proposal_flush_time then
-    failwith("'proposal_expired_time' needs to be bigger than 'proposal_flush_time'.")
-  else if data.proposal_flush_time <= (data.period.length * 2n) then
-    failwith("'proposal_flush_time' needs to be more than twice the 'period' length.")
+let validate_proposal_flush_expired_level (data : initial_config_data) : unit =
+  if data.proposal_expired_level.blocks <= data.proposal_flush_level.blocks then
+    failwith("proposal_expired_level needs to be bigger than proposal_flush_level")
+  else if data.proposal_flush_level.blocks <= (data.period.blocks * 2n) then
+    failwith("proposal_flush_level needs to be more than twice the voting_period length")
```
pasqu4le, line 11: (see previous comment) ... here I suggest:
```suggestion
  else if data.proposal_flush_level.blocks < (data.period.blocks * 2n) then
    failwith("proposal_flush_level needs to be more than twice the 'period' length.")
```

pasqu4le (changes requested):
Note: needs a rebase as well

Makefile
```diff
@@ -56,20 +56,20 @@ ifeq ($(OPTIMIZE), true)
 endif
 	#
 
-$(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
+$(OUT)/trivialDAO_storage.tz : current_level = 100n
```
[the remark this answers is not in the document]
sras: I think we should not have a default for it.

I ll change it so and rename the argument/field. Let me know if you think otherwise.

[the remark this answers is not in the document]
pasqu4le: :+1:  sounds good to me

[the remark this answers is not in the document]
sras: Done

src/defaults.mligo
sras: Changed.

Makefile
sras: I actually set the block count so that it would be the same duration. But I have removed the comments now.

docs/building.md
```diff
@@ -75,20 +75,20 @@ make out/trivialDAO_storage.tz \
   guardian_address=KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh \
   governance_token_address=KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5 \
   governance_token_id=0n \
-  now_val=`date +"%s"` \
+  current_level=100n \
```
pasqu4le, line 78: Looks like this document needs an update.

src/defaults.mligo
pasqu4le: @sras can you also please check that there's no ambiguity in the documentation?

pasqu4le: @sras can you also please check that there's no ambiguity in the documentation?

sras: Checked, and it seems that the current description can have some ambiguity, because it says something like "x levels after which a propsal is raised". Because it does not make it clear if the block at which the proposal is raised is counted as one.

So I have now introduced a concept of a 'proposal age', and included a section in the spec doc that describe how we count blocks, and how age for proposal is calcuated. As a related change, I have also renamed `is_level_reached` function as `is_proposal_age`, to make it clear that it is checking the proposal age.

So in the comments is specification and types.ligo associated with config type, I have used this proposal age to describe the behavior. I think this might help to reduce ambiguity.

docs/building.md
sras: :facepalm: Updated.

docs/building.md
```diff
@@ -75,20 +75,20 @@ make out/trivialDAO_storage.tz \
...
   quorum_change=5n \
   max_quorum_change=19n \
   governance_total_supply=1000n \
-  proposal_flush_time=1900801n \
-  proposal_expired_time=2851200n
+  proposal_flush_level=36000n  \
+  proposal_expired_level=47520n \
 ```
 
 The `admin_address`, `guardian_address`, `governance_token_address`, and `governance_token_id`
```
pasqu4le, line 94: This should list the `start_level` to, since it's now required (same for the two variants below)

Makefile
```diff
@@ -56,20 +56,19 @@ ifeq ($(OPTIMIZE), true)
...
-$(OUT)/trivialDAO_storage.tz : now_val = `date +"%s"`
 $(OUT)/trivialDAO_storage.tz : metadata_map = Big_map.empty
 $(OUT)/trivialDAO_storage.tz : freeze_history = []
 $(OUT)/trivialDAO_storage.tz : fixed_proposal_fee_in_token = 0n
 $(OUT)/trivialDAO_storage.tz : quorum_threshold = 10n
 $(OUT)/trivialDAO_storage.tz : min_quorum = 1n
 $(OUT)/trivialDAO_storage.tz : max_quorum = 99n
 $(OUT)/trivialDAO_storage.tz : max_voters = 1000n
-$(OUT)/trivialDAO_storage.tz : period = 950400n # 11 days
+$(OUT)/trivialDAO_storage.tz : period = 15840n # 11 days
```
pasqu4le, line 66: I think we should remove this comment too (and same below), levels can have different duration and this may be incorrect (or become so).

pasqu4le (approved):
Last two small things, but LGTM, feel free to squash these changes and merge

Makefile
sras: Removed.

docs/building.md
sras: Yes. added.

PR #293, opened 2021-07-05
[#47] Refine tests

### LINKED ISSUE #47: Refine time-sensitive tests

sras:
## Related issue(s)

Resolves #47 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/baseDAO-ligo-meta.cabal
+++ b/haskell/baseDAO-ligo-meta.cabal
@@ -1,6 +1,6 @@
 cabal-version: 2.2
 
--- This file has been generated from package.yaml by hpack version 0.34.3.
+-- This file has been generated from package.yaml by hpack version 0.34.4.
 --
 -- see: https://github.com/sol/hpack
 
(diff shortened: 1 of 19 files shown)


pasqu4le:
Closing because the issue was already solved elsewhere (see comments there)

PR #297, opened 2021-08-19
[#68] Add a stress test for the number of proposals

### LINKED ISSUE #68: Allow the option of unlimited number of proposals in a cycle

sras:
## Description

Adds a stress test to check that the contract can handle a thousand proposals without getting locked.

## Related issue(s)

Resolves #68 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
@@ -17,6 +17,7 @@ module Test.Ligo.BaseDAO.Proposal.Propose
   , proposalBoundedValue
   , proposerIsReturnedFeeAfterSucceeding
   , proposalStressTest
+  , proposalStressTest_
   , rejectProposal
   , validProposal
   , validProposalWithFixedFee
@@ -26,10 +27,14 @@ module Test.Ligo.BaseDAO.Proposal.Propose
 import Universum
 
 import Lorentz hiding (assert, (>>))
+import Michelson.Printer
+import Michelson.Typed.Convert (convertContract, untypeValue)
 import Morley.Nettest
 import Util.Named
 
+import Ligo.BaseDAO.Contract
 import Ligo.BaseDAO.Types
+import Ligo.Util
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Test.Ligo.BaseDAO.Common
 import Test.Ligo.BaseDAO.Proposal.Config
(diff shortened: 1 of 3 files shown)


rinn7e:
@sras Correct me if I'm wrong, but the original issue is about "Allow the option of unlimited number of proposals in a cycle #68". Since it is already the case (there is no limit), we simply want to add a test for this. Is this correct?

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
```diff
@@ -521,3 +522,44 @@ proposalBoundedValue originateFn = do
...
+      traverse_ (\c ->
+        let params = ProposeParams
+              { ppFrozenToken = 10
+              , ppProposalMetadata = lPackValueRaw c
+              , ppFrom = dodOwner1
+              }
+        in call dodDao (Call @"Propose") params) range
+
+  withSender dodOwner1 $
+    call dodDao (Call @"Freeze") (#amount .! 1)
```
rinn7e, line 565: Should we try to vote/drop_proposal on some of these proposals to make sure it is fine?

rinn7e (changes requested):

sras:
>Since it is already the case (there is no limit), we simply want to add a test for this. Is this correct?

Yes. But I think it is more about testing that the contract is not locked (because of the storage cost of decoding the storage is too big) even after adding a lot of proposals, rather than checking the logic.

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
sras: I think it does not make much of a difference since the storage needs to be decoded no matter which entrypoint we call.

rinn7e: I see, since the concern is about the storage decoding, I'm fine with this.

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
```diff
@@ -521,3 +522,44 @@ proposalBoundedValue originateFn = do
...
+  let ensureProposalPeriod = do
+        nowLevel <- getLevel
+        let periodNum = div ((nowLevel + 10) - originLevel) dodPeriod
+        when (even periodNum) $ do
+          let waitTill = originLevel + (periodNum + 1)*dodPeriod
+          advanceToLevel waitTill
+
+  -- Create a thousand proposals in batchs of 100
+  forM_ [0..(99 :: Natural)] $ \count -> do
+    let multiplier = 10
```
rinn7e, line 551: I'm wondering what would be the maximum proposals we can handle. Did you try let say 1 million and see if this causes error or not?

pasqu4le:
@sras since this additional test seems to create some problem and additionally adds even more time to the execution of the tests, couldn't we go instead do something as suggested [here](https://github.com/tezos-commons/baseDAO/pull/172#discussion_r599522411)?

Instead of calling many times the `propose` entrypoint we could originate the contract with a 1000 proposals in it's initial storage and verify with any call that it is still operating correctly. What do you think?

Note: if the contract with this full storage is too big (which may be the case or not) it should be possible for us to use the `originateLarge` functions instead

pasqu4le:
Closing this as it has been made unnecessary by #316

PR #298, opened 2021-08-20
[#139] Replace contract call checks with storage checks

### LINKED ISSUE #139: Reduce running time of nettests in local chain

Tests take quite a lot of time in local chain and we often have to wait for CI for quite long.

Currently waiting times are tolerable, but when we enable network tests for LIGO (see #124 ), they will increase and they will probably continue increasing as we add more features. So we'll soon need to reduce the required time.

Possible solutions: use batching  or replace usages of view entrypoints with direct storage checking (this connects this issue to #31 and consequently to morley/346.

# Acceptance criteria

The network tests time cost for CI is reduced (or kept) to being a tolerable waiting time for PRs.

sras:
## Description

Some checks in tests that currently uses additional contract calls were replaced with direct storage checks to reduce running time of tests.

## Related issue(s)

Resolves #139

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- [  ] My commits comply with [the following policy](https://www.notion.so/serokell/Commit-and-PR-policy-4cf98e1a910a415d86b5f2491d9af1af).

### DIFF
--- a/haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
@@ -72,6 +72,8 @@ changeToPendingAdmin withOriginatedFn initialStorage =
       withSender owner $ call baseDao (Call @"Transfer_ownership")
         (#newOwner .! wallet1)
       withSender wallet1 $ call baseDao (Call @"Accept_ownership") ()
+      administrator <- (sAdminRPC . fsStorageRPC) <$> (getStorageRPC baseDao)
+      assert (administrator == wallet1) "Administrator was not set from pending owner"
 
 noPendingAdmin
   :: MonadNettest caps base m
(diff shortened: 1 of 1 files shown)


haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
```diff
@@ -72,6 +72,7 @@ changeToPendingAdmin withOriginatedFn initialStorage =
       withSender owner $ call baseDao (Call @"Transfer_ownership")
         (#newOwner .! wallet1)
       withSender wallet1 $ call baseDao (Call @"Accept_ownership") ()
+      -- TODO@SRAS Add test to check new owner is set
```
rinn7e, line 75: Maybe we should create a new issue for this?

rinn7e (changes requested):

haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
sras: Ah, I missed to fill this. Have included the check now.

rinn7e (approved):

PR #300, opened 2021-09-08
[#259] Unsort Haskell constructors

### LINKED ISSUE #259: Stop sorting haskell's type constructors and let ligoLayout do it

Some of the constructors for Haskell types have been pre-sorted manually in order for `ligoLayout` to produce the representation that matches the one produced by LIGO.

See the `TODO` in `Ligo.BaseDAO.Types`.

This was however due to an issue, [morley/527](https://gitlab.com/morley-framework/morley/-/issues/527), which has since be solved.

We should check that everything now works correctly and stop manually sorting type constructors.

# Acceptance criteria

In the `haskell/` library type constructors are sorted by `ligoLayout` and not manually.

sras:
## Description

Problem: We have some comment that cautions agaist changing the ordering
of constructors of the Haskell contract parameter. But this is outdated
as the issue causing it have been fixed.

Solution: Change the sort order of constructors and check that tests
pass. Remove the respective comments and TODO.

## Related issue(s)

Resolves #259 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/src/Ligo/BaseDAO/Types.hs
+++ b/haskell/src/Ligo/BaseDAO/Types.hs
@@ -509,19 +509,13 @@ type CustomEntrypoint = (MText, ByteString)
 type FreezeParam = ("amount" :! Natural)
 type UnfreezeParam = ("amount" :! Natural)
 
--- NOTE: Constructors of the parameter types should remain sorted for the
---'ligoLayout' custom derivation to work.
---
--- TODO #259: The comment above should no longer be the case since the following
--- issue has been fixed:
--- https://gitlab.com/morley-framework/morley/-/issues/527
 data ForbidXTZParam
   = Drop_proposal ProposalKey
+  | Vote [PermitProtected VoteParam]
   | Flush Natural
   | Freeze FreezeParam
   | Unfreeze UnfreezeParam
   | Update_delegate [DelegateParam]
-  | Vote [PermitProtected VoteParam]
   deriving stock (Eq, Show)
 
 instance Buildable ForbidXTZParam where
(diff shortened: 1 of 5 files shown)


haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -509,19 +509,13 @@ type CustomEntrypoint = (MText, ByteString)
...
   = Drop_proposal ProposalKey
-  | Flush Natural
-  | Freeze FreezeParam
-  | Unfreeze UnfreezeParam
-  | Update_delegate [DelegateParam]
   | Vote [PermitProtected VoteParam]
+  | Update_delegate [DelegateParam]
+  | Unfreeze UnfreezeParam
+  | Freeze FreezeParam
+  | Flush Natural
```
pasqu4le, line 518: Instead of using another random sorting I think it would be more useful to have this match the order in which the contructors of the relative Ligo type are defined in `types.mligo`:
```suggestion
  | Vote [PermitProtected VoteParam]
  | Flush Natural
  | Freeze FreezeParam
  | Unfreeze UnfreezeParam
  | Update_delegate [DelegateParam]
```

pasqu4le: Related: it's possible that we did something similar in other types in this module, can you check the other places in which `ligoLayout` was used?

sras: I will do it. But I didn't think it mattered because the sorting of constructors does not matter at all now, so it was fine to leave the rest at what ever current ordering they are in. And by shuffling one type, we got proof that it works as expected.

pasqu4le: Yeah, that's a good point, reordering the other types won't prove much more.

I was thinking however that it wouldn't hurt either and generally it's easier to deal with changes on the two "sides" when things are lined up.

sras: Done.

pasqu4le (approved):

PR #301, opened 2021-09-09
[#233] clean up and make new release

### LINKED ISSUE #233: Stabilize the contract and make a new release

pasqu4le:
## Description

Several small adjustments and version number bump.
This follows the addition of better tests and the now stable contract logic.

See relative commit messages for an explanation on the changes.

## Related issue(s)

Resolves #233 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


src/base_DAO.mligo
```diff
@@ -40,8 +38,7 @@ let allowing_xtz (param, store, config : allow_xtz_params * storage * config) =
   | Accept_ownership           -> accept_ownership(store)
 
 (*
- * The actual DAO contract, which in this version is the same independently from
- * the DAO logic.
+ * The actual DAO contract, which is the same independently from the DAO logic.
```
sras, line 41: >  * The actual DAO contract, which is the same independently from the DAO logic.

May be we can further rephrase it as "which is the same, despite the specific DAO logic included", or something simlar.

src/treasuryDAO/types.mligo
```diff
@@ -19,18 +18,17 @@ type initial_treasuryDAO_storage =
   ; max_xtz_amount : tez
   }
 
-// Treasury dao `proposal_metadata` contains the type of proposal.
 type treasury_dao_transfer_proposal =
   { agora_post_id : nat
   ; transfers : transfer_type list
   }
 
+// Treasury dao `proposal_metadata`, defining the type its proposals.
```
sras, line 26: > defining the type its proposals.

'of' is probably missing here after 'type'.

sras (approved):
Reviewed and tested locally after updating ligo. Everything seems to work. So LGTM except for minor things.

PR #305, opened 2021-09-27
[#263] Build different variants instead of using lambdas from storage

### LINKED ISSUE #263: Stop using lambdas for instantiation and make compile-time choices instead

Currently the various instantiation of the DAO (`trivialDAO`, `registryDAO` and `treasuryDAO`, as well as custom DAOs) all have the same contract code and they only differ by stored lambdas, custom entrypoints and fields.

Such a setup however has a heavy cost, mainly due to:
- large Michelson types that have to be explicitly paid for in the storage (for lambdas, etc.)
- unpacking of data from `bytes` to allow for generality between the various versions

This also poorly reflects on the UX as packing is necessary for many entrypoint calls and that's inconvenient.

The alternative is to create different contract `code` (as well as `storage`) at compile time for the different instantiations.

Known LIGO tecniques based on `#if`/`#define`/`#include` can be used for this, however some consideration about import ordering might be necessary and non-trivial.

# Acceptance criteria

- the various contract instantiation are obtained at compile-time and produce different code
- there are no more lambdas and big_maps in the storage to contain custom functions/entrypoints/fields
- there is a way to create custom DAOs by means of a template-like module, whose usage is documented.

sras:
## Description

Changes the build to create different variants statically, and remove lambdas from storage.

More description in commit messages

## Related issue(s)

Resolves #263 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e:
@sras From the look of it, all seems fine for me. Do you still have thing lefts to be done, or the PR is already ready?

sras:
>Do you still have thing lefts to be done, or the PR is already ready?

@rinn7e I don't remember leaving anything out. Can you think of anything missing from the original requirement?

rinn7e:
> > Do you still have thing lefts to be done, or the PR is already ready?
> 
> @rinn7e I don't remember leaving anything out. Can you think of anything missing from the original requirement?

Couldn't really think of anything, I'll approve but we may want to wait for @pasqu4le to take a look as well.

rinn7e (approved):

pasqu4le:
@sras please do and sorry for the waiting in the review

docs/building.md
```diff
@@ -17,13 +17,59 @@ The latest working version tested is [0.24.0](https://gitlab.com/ligolang/ligo/-
 
 You can use the [`Makefile`](../Makefile) to build the LIGO contract by simply running:
 ```sh
-make out/baseDAO.tz
+make out/trivialDAO.tz
 ```
-which will use `ligo compile-contract` and save the result in `out/baseDAO.tz`.
+which will use `ligo compile-contract` and save the result in `out/trivialDAO.tz`.
```
pasqu4le, line 22:
```suggestion
which will use `ligo compile contract` and save the result in `out/trivialDAO.tz`.
```

docs/building.md
```diff
@@ -17,13 +17,59 @@ The latest working version tested is [0.24.0](https://gitlab.com/ligolang/ligo/-
 
 You can use the [`Makefile`](../Makefile) to build the LIGO contract by simply running:
 ```sh
-make out/baseDAO.tz
+make out/trivialDAO.tz
```
pasqu4le, line 20: This section is now somewhat odd, because it's about "Generation the contract code" but doesn't mention the fact that there are now multiple ways to do so...

IMO it should point this out and not only give the instructions for `trivialDAO`.

.buildkite/pipeline.yml
```diff
@@ -44,12 +44,13 @@ steps:
...
-    - nix run -f ci.nix morley -c morley optimize --contract ./ligo-out/baseDAO.tz --output baseDAO.tz
+    - nix run -f ci.nix morley -c morley optimize --contract ./ligo-out/trivialDAO.tz --output trivialDAO.tz
+    - nix run -f ci.nix morley -c morley optimize --contract ./ligo-out/registryDAO.tz --output registryDAO.tz
+    - nix run -f ci.nix morley -c morley optimize --contract ./ligo-out/treasuryDAO.tz --output treasuryDAO.tz
     - cp ./ligo-out/trivialDAO_storage.tz trivialDAO_storage.tz
     - cp ./ligo-out/registryDAO_storage.tz registryDAO_storage.tz
     - cp ./ligo-out/treasuryDAO_storage.tz treasuryDAO_storage.tz
     artifact_paths:
-      - baseDAO.tz
       - trivialDAO_storage.tz
```
pasqu4le, line 54: Why not give the compiled contracts as artifacts too?
```suggestion
      - trivialDAO.tz
      - registryDAO.tz
      - treasuryDAO.tz
      - trivialDAO_storage.tz
```

.buildkite/pipeline.yml
```diff
@@ -150,7 +151,6 @@ steps:
     depends_on: ligo-test
     commands:
       - mkdir assets
-      - buildkite-agent artifact download baseDAO.tz assets --step "build-ligo"
```
pasqu4le, line 153: Same here, why not add the compiled contracts from the artifacts to the auto pre-release?

Makefile
```diff
@@ -27,32 +27,126 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
+registryDAO: VARIANT = registry
+registryDAO: baseDAO
+
+treasuryDAO: VARIANT = treasury
+treasuryDAO: baseDAO
+
+testDAO: VARIANT = test
+testDAO: baseDAO
+
+all: $(OUT)/trivialDAO.tz $(OUT)/registryDAO.tz $(OUT)/treasuryDAO.tz $(OUT)/testDAO.tz
```
pasqu4le, line 51: by default `make` runs the top target that doesn't beging with a `.`

I recommend moving `all` again to the top because of this

Makefile
```diff
@@ -27,32 +27,126 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
+trivialDAO: baseDAO
+
+registryDAO: VARIANT = registry
+registryDAO: baseDAO
+
+treasuryDAO: VARIANT = treasury
+treasuryDAO: baseDAO
+
+testDAO: VARIANT = test
+testDAO: baseDAO
```
pasqu4le, line 49: Are these targets used? AFAIU these do he same thing as their `$(OUT)/xxx.tz` equivalents, right?

Are they meant to be more convenient?

Makefile
```diff
@@ -27,32 +27,126 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
+
+all: $(OUT)/trivialDAO.tz $(OUT)/registryDAO.tz $(OUT)/treasuryDAO.tz $(OUT)/testDAO.tz
 
 # Compile LIGO contract into its michelson representation.
-$(OUT)/baseDAO.tz: src/**
+baseDAO: src/**
```
pasqu4le, line 54: related to the previous comment as well: I see that there is now this target (used by ones above) but that its content is duplicated in the ones below, which seems to be doing the same thing :thinking: 

This was beforehand a single `$(OUT)/baseDAO.tz` target, couldn't these options all be re-united again?
Maybe in a single `$(OUT)/$(VARIANT)DAO.tz` target?

docs/specification.md
```diff
@@ -34,14 +34,13 @@ An existing separate FA2 contract is required and used for governance.
 # Configuration
 
 BaseDAO is a concrete smart contract, but also a framework to implement various DAOs.
```
pasqu4le, line 36: I'd say that `baseDAO` is no longer a concrete smart contract, given that only its variants can be built.

pasqu4le (commented):
One more nitpick: there is a `Soltion` typo in 24eb703357e198228fcd23e8c5ccf17e638f0d00

docs/specification.md
sras: Can you please suggest a good way to describe it at its current state?

Makefile
sras: Yes, it seems that it was left for convinence. 

Basically I initially thought that we can eliminate indvidual build section for each variant, by setting the `VARIANT` env variable. 

But it didn't work because we also needed file targets so that the dependency lookup will work for `make`.  I think it is not very useful to leave this out, so ll remove it. Let me know if you can find a reason to leave them in.

Makefile
sras: Done

.buildkite/pipeline.yml
sras: Have made a change to addres this.

Makefile
sras: Ah, yes.

sras: I just could't make it work. 

It appears that the wildcard does not match in more than one way, so when doing `make all`, make throws an error after building the variant for the first match.

Makefile
pasqu4le: None comes to mind, feel free to remove these

docs/specification.md
pasqu4le: Well, I would simply remove the "concrete smart contract" part and leave the fact that it is a framework.

I'd suggest:
```suggestion
BaseDAO is a framework to implement various DAO smart contracts.
```

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
 # Utility function to require a variable to be defined
 require_defined = $(if $(value $1),$(value $1),$(error "$1 needs to be defined"))
 
+# What implementation to build
+VARIANT ?= trivial
```
pasqu4le, line 31:
```suggestion
```

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
 # Where to put build files
 OUT ?= out
 # Where to put typescript files
 TS_OUT ?= typescript
 
 .PHONY: all clean test typescript
 
-all: \
-	$(OUT)/baseDAO.tz
+all: $(OUT)/trivialDAO.tz $(OUT)/registryDAO.tz $(OUT)/treasuryDAO.tz $(OUT)/testDAO.tz
```
pasqu4le, line 39: this can stay the same :slightly_smiling_face:

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
+treasuryDAO: VARIANT = treasury
+treasuryDAO: baseDAO
+
+testDAO: VARIANT = test
+testDAO: baseDAO
 
 # Compile LIGO contract into its michelson representation.
-$(OUT)/baseDAO.tz: src/**
+$(OUT)/$(VARIANT)DAO.tz: src/**
+	cp src/$(VARIANT)DAO.mligo src/implementation.mligo
```
pasqu4le, line 55:
```suggestion
$(OUT)/%DAO.tz: src/**
	cp src/$*DAO.mligo src/implementation.mligo
```
this can be changed to use a wildcard (`%`), which will match any non-empty string.

Note that patterns are expanded first, so unless `OUT` is changed from the default value this will only match if the destination is in `out`.
E.g. these would work:
```sh
make out/trivialDAO.tz
make different_dir/trivialDAO.tz OUT=different_dir
```
but this would not;
```sh
make different_dir/trivialDAO.tz
```

Here I also used the automatic `$*` variable that expands to what `%` matched on, see [the docs](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html).

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
 	mkdir -p $(OUT)
 	# ============== Compiling contract ============== #
-	$(BUILD) src/base_DAO.mligo base_DAO_contract --output-file $(OUT)/baseDAO.tz
+	$(BUILD) src/base_DAO.mligo -e base_DAO_contract --output-file $(OUT)/$(VARIANT)DAO.tz
```
pasqu4le, line 58: while we are at it, this could use another automatic variable, `$@`:
```suggestion
	$(BUILD) src/base_DAO.mligo -e base_DAO_contract --output-file $@
```

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
 	# ============== Compilation successful ============== #
-	# See "$(OUT)/baseDAO.tz" for compilation result #
+	# See "$(OUT)/$(VARIANT)DAO.tz" for compilation result #
```
pasqu4le, line 60:
```suggestion
	# See "$@" for compilation result #
```
same here

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
 
 	# strip the surrounding braces and indentation,
 	# note that dollar char is escaped as part of Makefile
-	sed -i '/^ *$$/d' $(OUT)/baseDAO.tz
-	sed -i 's/^[{ ] //g' $(OUT)/baseDAO.tz
-	sed -i '$$s/[}] *$$//' $(OUT)/baseDAO.tz
+	sed -i '/^ *$$/d' $(OUT)/$(VARIANT)DAO.tz
+	sed -i 's/^[{ ] //g' $(OUT)/$(VARIANT)DAO.tz
+	sed -i '$$s/[}] *$$//' $(OUT)/$(VARIANT)DAO.tz
```
pasqu4le, line 66:
```suggestion
	sed -i '/^ *$$/d' $@
	sed -i 's/^[{ ] //g' $@
	sed -i '$$s/[}] *$$//' $@
```
and here

Makefile
```diff
@@ -27,32 +27,46 @@ escape_double_quote = $(subst $\",$\\",$(1))
...
 ifeq ($(OPTIMIZE), true)
 	# ============== Optimizing contract ============== #
-	$(MORLEY) optimize --contract $(OUT)/baseDAO.tz --output $(OUT)/baseDAO.tz
+	$(MORLEY) optimize --contract $(OUT)/$(VARIANT)DAO.tz --output $(OUT)/($VARIANT)DAO.tz
```
pasqu4le, line 69:
```suggestion
	$(MORLEY) optimize --contract $@ --output $@
```
and finally here

pasqu4le (commented):
I made this small "review" to put all the suggestions needed for [the previous discussion](https://github.com/tezos-commons/baseDAO/pull/305#discussion_r778907454) in one place.

Makefile
pasqu4le: Ah, I see now, my suggestion isn't actually possible to use.

I looked at the docs and the patterns here aren't actually wildcards, they are variables.
However we can combine them with wildcards to get what we want.

See the comments that I left in the "review" below

Makefile
sras: Removed

Makefile
sras: Worked. Comments were very helpful. Thanks!

docs/building.md
sras: Have changed

sras:
> One more nitpick: there is a `Soltion` typo in [24eb703](https://github.com/tezos-commons/baseDAO/commit/24eb703357e198228fcd23e8c5ccf17e638f0d00)

Have corrected.

sras:
@pasqu4le Do you mind if I clean up and squash the commits?

pasqu4le:
@sras not at all, please rebase and feel free to squash the commits

sras:
> @sras not at all, please rebase and feel free to squash the commits

Rebased

sras:
@pasqu4le ping

scripts/generate_error_code.hs
```diff
@@ -96,8 +96,6 @@ invalidInputErrors = errorsEnumerate 100
     :? ("The sender has not been delegated the control of the required tokens.", NoArg)
   , "fail_decision_lambda"
     :? ("Executing the proposal's decision lambda results in failure.", NoArg)
-  , "entrypoint_not_found"
```
pasqu4le, line 99: Shouldn't `removedError` be used here instead?

src/template.mligo
```diff
@@ -0,0 +1,62 @@
...
+
+// Replace TEMPLATE_H with something unique corresponding to your
+// variant. For example, if your variant is called MyBaseDAOVariant you
+// can change TEMPLATE_H to MYBASEDAOVARIANT_H so this section reads
+//
+// #if !MYBASEDAOVARIANT_H
+// #define MYBASEDAOVARIANT_H
+
+#if !TEMPLATE_H
+#define TEMPLATE_H
```
pasqu4le, line 12: Couldn't this be fixed instead and used by all the implementations (including the ones already existing here)?

Something like:
```suggestion
#if !IMPLEMENTATION
#define IMPLEMENTATION
```
or
```suggestion
#if !VARIANT
#define VARIANT
```
or something like that?

src/template.mligo
```diff
@@ -0,0 +1,62 @@
...
+
+//
+
+#include "types.mligo"
+
```
pasqu4le, line 17: :thinking: how come `contract_extra` is not variant-specific as well?

AFAIK that field has a different content depending on the contract, but one that will be fixed throughout its life.

This would save a lot of unnecessary `pack`/`unpack` calls and be overall easier to use, I think.

IIRC this was sort of what we did back when there was a `lorentz` version of baseDAO too BTW.

src/template.mligo
```diff
@@ -0,0 +1,62 @@
...
+// -----------------------------------------------------------------
+// Decision_lambda
+//
+// The decision lambda is executed based on a successful proposal.  It has
+// access to the proposal, can modify `contractExtra` and perform arbitrary
+// operations.
+// -----------------------------------------------------------------
+
+let decision_lambda (input : decision_lambda_input)
```
pasqu4le, line 38: It seems a bit odd that this is still called `decision_lambda` now, given that it's no longer technically a lambda.

What do you think about renaming it?

src/parameter.mligo
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !PARAMETER_H
+#define PARAMETER_H
+#include "types.mligo"
+#include "implementation.mligo"
```
pasqu4le, line 7: This seems to be working, but taking into account my previous comment about `contract_extra` I suppose that it would result in a need for another `storage.mligo` file as a result.

I recommend doing something else instead: move all the types except `storage` and `parameter` in a base module, e.g. `types/base.mligo` and then then have `types.mligo` instead only contain:
1. the import of `types/base.mligo`
2. the import of `implementation.mligo`
3. the definition of `parameter` and `storage`.

In this way everything except the implementation module can import `types.mligo` and find everything there...
(The implementation modules can't, but that's ok as they shouldn't be allowed to access the full storage nor parameter anyway).

Note: the `decision_lambda_input` and `decision_lambda_output` wouldn't fit well in this, but I think they could be removed.

src/registryDAO/types.mligo
```diff
@@ -47,6 +47,10 @@ type lookup_registry_param =
 
 type lookup_registry_view = (registry_key * (registry_value option)) contract
 
+type custom_ep_param =
+  | Lookup_registry of lookup_registry_param
+  | RegistryCepDummy of unit
```
pasqu4le, line 52: What is this for? :thinking:

src/testDAO.mligo
```diff
@@ -0,0 +1,44 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !TEST_H
```
pasqu4le, line 4: Could you please clarify why is this `testDAO` needed?

pasqu4le (changes requested):

src/registryDAO/types.mligo
sras: It is so that we can annotate the "Lookup_registry" entrypoint. Or is there another way to annotate this entrypoint from this parameter nesting level?

src/testDAO.mligo
sras: I thought I had left a comment explaining it somewhere...

This was required to use in test where we loaded some test implementations for decision lambda. Now that we have concrete variants without lambdas in storage, this variant is one that concrete entrypoints, which just call lambdas from storage, emulating earlier functionality.

src/template.mligo
sras: Ah, yes. Good catch! Let me try.

src/parameter.mligo
sras: >In this way everything except the implementation module can import types.mligo and find everything there...

But implementation module import modules that inturn depend on `types.mligo`. I thought it would somehow work, but it does not seem to.

pasqu4le: :thinking: I think that implementation modules only import helpers, common modules and such, AFAICT these could also import only a `types/base.mligo` module alone.

It takes away a bit of the convenience, but it might still be worth it.

src/registryDAO/types.mligo
pasqu4le: Ah, I see, I don't think there are other way to annotate this with Ligo syntax.

Possibly there's some way to do this with some `#if` clauses, but it doesn't seem straightforward, so I think we can avoid this not and perhaps get back to it later.

Please add a comment to explain this tho, I think it should have a TODO too.

src/testDAO.mligo
pasqu4le: > I thought I had left a comment explaining it somewhere...

Maybe you were thinking of this in the commit message that introduces it:
> Some tests used some predefined behavior of the lambdas. But now that
the lambdas were removed, we need a way to accomodate these tests. For
this a fourth variant, that fetched test lambdas from contract extra was
added. And in tests that checked for the core behavior, this variant was
used.

I don't see another explanation in the code.

> this variant is one that have concrete entrypoints, but just wraps calls to lambdas from storage, emulating earlier functionality for tests.

Wait, doesn't this invalidate the tests tho?
I mean the contract that is being tested isn't actually any of variants that can be compiled, right?

IMO tests that check variant-specific code should run on both `registryDAO` and `treasuryDAO` (separately), whereas the ones that don't interact with the "variable" parts can run using any of them, `trivialDAO` would be my pick.

sras: >Maybe you were thinking of this in the commit message that introduces it:

Yes.

>I mean the contract that is being tested isn't actually any of variants that can be compiled, right?

The registry and treasury tests uses the respective concrete contracts (not testDAO). You can see [here](https://github.com/tezos-commons/baseDAO/blob/1d5c5356181cbf0dccdbf90328cc1c05c5b833ac/haskell/test/Test/Ligo/RegistryDAO.hs#L61) for example. 

The remaining tests use testDAO, as these tests does not call any specific variant behavior, except those that test the behavior of the rest of the contract for a very specific variant code behavior. And for this lambdas loaded to testDAO is used to implement this specific behavior.

I am not sure, but it seem it is safer to use `trivialDAO` for the tests that does not need this behavior. So ll try to switch them to trivialDAO.

pasqu4le: I agree, `trivialDAO` seems like it would be better used for this :+1:

src/parameter.mligo
pasqu4le: :thinking: AFAICT some of them do indeed, I missed that.

Not sure if we could change something about those, but I'd leave this for a separate issue.

Never mind this comment(s) then :ok_hand:

src/template.mligo
sras: I changed the LIGO contract to have differnet storages, but I am not able to find a way to make this change on the Haskell side. Specifically I cannot find a way to deriving `StorageRPC` instance for different variants of storage, without completely duplicating the common fields.

I mean, how can we accomodate the Storage variants here?

```
deriveRPCWithStrategy "Storage" ligoLayout
```

Can you think of a solution for this?

pasqu4le: IIUC you now have a single `Storage` type with an argument (for `contract_extra`) and a concrete datatype for each variant, right?

If that's the case, have you tried doing something similar to what [`morley-ledgers` does](https://gitlab.com/morley-framework/morley-ledgers/-/blob/3a7d14cff8ea011e79ceec0bcdc1668bf987f7e9/code/morley-ledgers/src/Lorentz/Contracts/ManagedLedger/Types.hs#L110-L114)?

In this case you would have something like:
```haskell
data Storage contract_extra = -- <definition>

data TrivialExtra = -- <definition>
type instance AsRPC TrivialExtra = TrivialExtra

data RegistryExtra = -- <definition>
type instance AsRPC RegistryExtra = RegistryExtra

data TreasuryExtra = -- <definition>
type instance AsRPC TreasuryExtra = TreasuryExtra


deriveRPC "Storage"

type TrivialStorage = Storage TrivialExtra
type TrivialStorageRPC = StorageRPC TrivialExtra

type RegistryStorage = Storage RegistryExtra
type RegistryStorageRPC = StorageRPC RegistryExtra

type TreasuryStorage = Storage TreasuryExtra
type TreasuryStorageRPC = StorageRPC TreasuryExtra
```

pasqu4le: :thinking: that sounds like there might be a mistake in the definition of `deriveRPC` itself.

Can you try to reproduce the issue with a minimal example?

pasqu4le: It indeed looks like we bumped in a bug here.

I tried to look into it a bit and I couldn't quickly find the root cause, so I opened [`morley/#754`](https://gitlab.com/morley-framework/morley/-/issues/754) with an even smaller reproducible example.

However it might take a while for that to be solved and for an update to be usable here.
In the meanwhile I think we should remove the use of `customGeneric`/`deriveRPCWithStrategy` and `ligoLayout` and instead sort the fields manually and use `deriving stock Generic`/`deriveRPC`, undoing (at least for this type) the work done in #259

pasqu4le: Did it cause complex changes?

I would have thought that such an update would have mostly required to rearrange fields and their uses :thinking:

sras: Oh, it was not about using stock derivation of Generic instance, but about the bigger task of switching to polymorphic storage on Haskell side.

scripts/generate_error_code.hs
sras: Yes, good catch. Changed.

src/template.mligo
sras: Have changed.

src/template.mligo
sras: Done now.

src/template.mligo
sras: Have renamed it to `decision_callback`.

src/registryDAO/types.mligo
sras: Have added a comment and a TODO.

src/testDAO.mligo
sras: Have switched to `trivialDAO` for most tests, and added a new variant `testDynamic` to implement tests that require specific behaviour of decision_callback etc, loaded as part of the test.

scripts/generate_error_code.hs
```diff
@@ -93,10 +93,9 @@ invalidInputErrors = errorsEnumerate 100
     :? ("There are no available proposals to flush.", NoArg)
   , "not_delegate"
     :? ("The sender has not been delegated the control of the required tokens.", NoArg)
-  , "fail_decision_lambda"
+  , "fail_decision_callback"
     :? ("Executing the proposal's decision lambda results in failure.", NoArg)
```
pasqu4le, line 97: I suppose that "lambda" should be removed from the description as well.

src/types.mligo
```diff
@@ -296,22 +281,10 @@ type initial_data =
...
 
 type config =
-  { proposal_check : propose_params * contract_extra -> unit
-  // ^ A lambda used to verify whether a proposal can be submitted.
-  // It checks 2 things: the proposal itself and the amount of tokens frozen upon submission.
-  // It allows the DAO to reject a proposal by arbitrary logic and captures bond requirements
-  ; rejected_proposal_slash_value : proposal * contract_extra -> nat
-  // ^ When a proposal is rejected, the value that the proposer gets back can be slashed.
-  // This lambda returns the amount to be slashed.
-  ; decision_lambda : decision_lambda
```
pasqu4le, line 309: _I'll mention this here just to pick a place to leave this comment_

I see that there are several mentions of this being a "lambda" yet:
- in documentation: `README.md`, `docs/specification.md`, `docs/registry.md`, etc.
- in Haskell modules, mostly in comments and some test names
- in LIGO code, again mostly in comments and some local variables

src/template_storage.mligo
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
```
pasqu4le, line 3: Much like `template.mligo` wouldn't it be best for this (and each implementation) to have a check that it wasn't imported already?
Like:
```
#if !VARIANT_STORAGE
#define VARIANT_STORAGE
```
or something like that

src/template_storage.mligo
```diff
@@ -0,0 +1,10 @@
...
+// This type should define the type of the contract-extra field in storage
+// for this variant.
+type contract_extra = (string, bytes) big_map
+
+// This should define a default value for the contract-extra field for the
+// variant.
+let default_extra : contract_extra = (Big_map.empty : contract_extra)
```
pasqu4le, line 10: Why define the standard as a `Big_map`, I think using `unit` (as in `trivialDAO`) would make this less of a burden in case someone is implementing a custom DAO and forgets to update these types (because they don't need any extra).

What do you think?

src/registryDAO_storage.mligo
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
```
pasqu4le, line 3: Not a big point, but since we have directories for variants (e.g. in this case `src/registryDAO`) wouldn't it be better for the storage definition to be moved in there?

I think that the same should be done for `template_storage.mligo` BTW

src/template.mligo
```diff
@@ -0,0 +1,55 @@
...
+// amount to be slashed.
+// -----------------------------------------------------------------
+
+let rejected_proposal_slash_value (_, _ : proposal * contract_extra) : nat = 0n
+
+// -----------------------------------------------------------------
+// Type that make up the custom entrypoint of the variant.
+//
+// -----------------------------------------------------------------
+type custom_ep_param = unit
```
pasqu4le, line 46: I think that during one of the recent changes the location of this got confused.

I see that it is defined here, but variants do different things:
- `testDAO` and `trivialDAO` define it in their "main" module
- `registryDAO` and `treasuryDAO` define it in their `types` module

I think that this should be consistent and that we should have a good default place.
Maybe there should be a `template/types.mligo`? Or this could be in the storage instead?

src/treasuryDAO_storage.mligo
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+type contract_extra =
+  { frozen_scale_value : nat option
+  ; frozen_extra_value : nat option
+  ; max_proposal_size : nat option
+  ; slash_scale_value : nat option
+  ; slash_division_value : nat option
+  ; min_xtz_amount : tez option
+  ; max_xtz_amount : tez option
```
pasqu4le, line 11: :thinking: I am confused: why are these `option`s?

If they are required in every case (as I think they are) then shouldn't they always have a defined value?

I took this as an example, but the same is the case for the others as well.

src/testDAO.mligo
pasqu4le: I haven't checked properly yet, but couldn't these test that require a specific behavior use one of the existing callbacks?

I would imaging that the 3 variants that we have would be able to cover all these tests :thinking:

typescript/baseDAO/src/api.ts
```diff
@@ -186,3 +181,22 @@ export class BaseDAOContract {
       contract => contract.methods.unfreeze(arg));
   }
 }
+
+export class RegistryDAOContract extends BaseDAOContract {
```
pasqu4le, line 185: Shouldn't there be one for `trivialDAO` too?

haskell/src/Ligo/BaseDAO/ErrorCodes.hs
```diff
@@ -96,8 +96,8 @@ notDelegate :: Natural
 notDelegate = 120
```
pasqu4le, line 97: This commit (bce9d640b063f475d64448746320f7948bb5937b) has no description in its message (even tho it will be a pretty simple one).

src/template_storage.mligo
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+// This type should define the type of the contract-extra field in storage
+// for this variant.
+type contract_extra = (string, bytes) big_map
+
```
pasqu4le, line 7: This commit (ddd9fa692a50961dbd11b12b8aaa819aace33c40) also doesn't have a description in its message and I believe that changes to this file are misplaced and actually belong to a different commit.

docs/building.md
```diff
@@ -64,6 +64,10 @@ type custom_ep_param =
   | MyCustomEpDummy of unit
 ```
 
+Next we should define the contract-extra field for the variant as well. This can be done by copying
+and the `src/template_storage.mligo` to `src/acmeDAO_storage.mligo` and modifying it to contain the
+actual type and a default value for the contract-extra field for this variant.
+
```
pasqu4le, line 70: Aside from some of the comments I left above, I think that even more updates are needed to many of the docs in the `docs` dir.

AFAICT a lot of details that were explained there are no longer correct.
Would you please check them carefully and update them accordingly?

src/treasuryDAO_storage.mligo
sras: Good catch. I think I made a blunder here.

src/registryDAO_storage.mligo
sras: I tried to do this but it didn't look good, because it would make the registry/treasury storage location to differ from what is said in the documentation, on how to define a variant. We can change the documentation and make it a little more complicated so that the user have to create a directory for the variant, and put the storage files there.

>I think that the same should be done for template_storage.mligo BTW

I didn't get this. Did you mean `template_storage.mligo` should be moved to `template` directory?

sras: I have new created a `variants` folder in `src` and have put `callbacks.mligo`, `storage.mligo` files in it. `callback.mligo` will be the `implementation.mligo` and `storage.mligo` will contact contract extra definitions. Similarly `templates` folder has been populated with template `callbacks.mligo` and `storage.mligo`. 

Does this look good?

src/base.mligo
```diff
@@ -0,0 +1,300 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
```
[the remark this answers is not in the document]
sras: It appears that this module was indeed reundant. Have removed it. I ll check if documenting other modules could help and do it if yes.

src/template_storage.mligo
sras: Changed to unit.

src/template_storage.mligo
sras: Yes, done.

src/template.mligo
```diff
@@ -0,0 +1,55 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !VARIANT
+#define VARIANT
+
+//
```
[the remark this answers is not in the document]
sras: Yes, removed.

src/testDAO.mligo
sras: >that require a specific behavior use one of the existing callbacks?

May be, but I think it is better to explicitly configure the behavior, than use some behavior that just incidently happen to be what we need.

src/template.mligo
sras: Have moved this type to the new `callbacks.mligo` module. `template/types.mligo` might sound very general and make `template/storage.mligo` redundant, and there are not many types to justify a separate module. One thing could be we can rename `callbacks.mligo` to `implementation.mligo` which make it ok to inlcude some type definitions in it.

src/template_storage.mligo
sras: Have expanded the commit message.

haskell/src/Ligo/BaseDAO/ErrorCodes.hs
sras: Have expanded the commit message.

src/base.mligo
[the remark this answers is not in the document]
sras: The rest of the modules does not appear to need any extra documentation.

docs/error-codes.md
```diff
@@ -38,7 +38,6 @@ Here is a summary of all the error codes thrown by the contract.
 | 119 | `empty_flush` | There are no available proposals to flush. |
 | 120 | `not_delegate` | The sender has not been delegated the control of the required tokens. |
 | 121 | `fail_decision_lambda` | Executing the proposal's decision lambda results in failure. |
```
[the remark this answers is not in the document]
sras: Yes, it is generated. Have changed it in the source now.

scripts/generate_error_code.hs
sras: Sorry about this. Have renamed lambda everywhere now.

src/types.mligo
sras: Have renamed everything now.

docs/building.md
sras: Have have checked all the docs in `docs` dir and updated where it seemed required.

src/registryDAO_storage.mligo
pasqu4le: What I was suggesting was to move this file (`src/registryDAO_storage.mligo`) to `src/registryDAO/storage.mligo` given that we already had an `src/registryDAO` directory.

> it would make the registry/treasury storage location to differ from what is said in the documentation, on how to define a variant.

Yeah, that's why I suggested to do the same for `template_storage.mligo`.

AFAIU each variant needs 2 files now, I think we should put these in separate directories so that:
- all types and code related to a variant can be easily found and are organized
- anyone that wants to create their own custom DAO can copy the `template` directory and fill in the two files as they wish

what do you think? is that possible?

I see that you have already done some work in that direction in a sense, except for a few things that are out of place IMO:
1. there is a `src/registryDAO/types.mligo` which is a bit odd given that this is the only file in `src/registryDAO` and everything else about registry DAO seems to be under `src/variants/registry`. Can't the two directory be merged? Perhaps the content of the `types.mligo` can also be moved to `callbacks` or `storage`
2. same for `treasuryDAO`
3. I see that `template` now is out of `src`. Given that it's LIGO code I'd move it in `src/template` or maybe even `src/variants/template`. 

I think that changing these would improve the codebase structure, which LGTM for everything else.

src/template.mligo
pasqu4le: > One thing could be we can rename callbacks.mligo to implementation.mligo which make it ok to inlcude some type definitions in it.

Sound like a good idea to me, especially since AFAIK the variant the `callbacks` of the variant that gets picked is copied to the `implementation.mligo` file.

This is also how `types.mligo` can be moved in the same directory which [I think should be done](https://github.com/tezos-commons/baseDAO/pull/305#discussion_r819023570).

src/testDAO.mligo
pasqu4le: Yeah, but I mean... do we need tests on `decision callback`s behavior that **is not** the one of a supported variant?

I think all of our tests are supposed to check the behavior of one between `registry`, `treasury` and the `trivial` implementation, right?

If so, we don't need this `testDAO` IMO, because we can rely on each specific implementation that we actually want to cover. I think that this would be a better idea, as it would give us coverage of what we are actually interested in testing, rather than something that's supposed to behave in the same way.

README.md
```diff
@@ -2,7 +2,7 @@
 
 BaseDAO is a generic smart contract on Tezos that enables a community to collectively govern resources, registries, or rules.
 
-The contract enables the creator to customize their DAO based on a number of attributes, and uses a ‘decision lambda’ to specify arbitrary code that can be governed by a DAO’s proposals.
+The contract enables the creator to customize their DAO based on a number of attributes, and uses a ‘decision callback’ to specify arbitrary code that can be governed by a DAO’s proposals.
 BaseDAO also includes Permit (TZIP-17) to enable off-chain voting.
 
 ## Documentation and configurations
```
pasqu4le, line 8: In this section there is a paragraph starting with:
```
BaseDAO uses a runtime configuration
```
which I believe is no longer correct

docs/registry.md
```diff
@@ -18,7 +18,7 @@ Extra storage data:
 We store these two `big_map`s separately because we want to include `proposalId` for deleted keys as well.
```
pasqu4le, line 18: I don't think that any of this paragraph nor the info above here still holds true. 

AFAICT there aren't any `big_map`s anymore, are there?

src/testDAO.mligo
sras: >do we need tests on decision callbacks behavior that is not the one of a supported variant

To be clear, we are not testing decision callback's behavior. We are testing "Given that decision callback do X,  does the rest of the contract behave in a certain way Y". So we are testing Y, not X. 

Now you are right that we can take an existing variant, and load it with some extra values, so that its decision callback will do X. I think it makes tests a little obscure. Because the intendent behavior of the callback is not as explicit, and will have to be inferred from the extra values, some proposal parameters and variant logic.  Also having the ability to pack arbitrary behavior in these callback could be useful in writing future tests.

Having said all that, I think what you suggest is not all that problematic, so I ll implement it for the sake of avoiding a drawn out discussion for a comparitively minor thing.

sras: Couple of problems I encounterd while trying this...

1. The decision lambda on the testDynamic contract is very liberal and does not check for the exact frozen tokens required as computed by frozen scale/extra values. It seems that without this, the 'createSampleProposal` function cannot create proposals, because it always freezes 10 tokens. 

2.  `createSampleProposal` create dummy proposals (by serializing some numbers), and when we use variants, it will try to decode them, and fail during flush. `testContract` contains a no-op decision lambda.

Let me know if you still want to proceed with this.

src/registryDAO_storage.mligo
sras: I have 

1. Moved `registryDAO/types.mligo` to `variants/registry/types.mligo`
2. And same for `treasury`.
3. Moved `template` to `src/variants`.

Will update the docs if this looks good.

typescript/baseDAO/src/api.ts
sras: Should be? I think we didn't have one because `trivialDAO` was never meant for real use.

docs/registry.md
sras: Oh, missed it. Changed "big_map" to "map" now.

README.md
sras: Have change it.

src/testDAO.mligo
pasqu4le: > To be clear, we are not testing decision callback's behavior. We are testing "Given that decision callback do X, does the rest of the contract behave in a certain way Y". So we are testing Y, not X.

I see, but in this case, can't the straightforward implementation of `trivialDAO` be used for that?

AFAICT all it does is return the input `extra`, so it should be possible to:
- test the "Y" without being concerned by what the decision lambda does
- keep creating dummy proposals in `createSampleProposal` (I assume it has to be changed a bit, but it should still be possible to generate these quite randomly).
- predict the frozen tokens needed to be used, even if that's a bit more annoying than allowing any value (it probably a good thing to not be so liberal here too)

> Let me know if you still want to proceed with this.

Yep, I think we should.
Sorry for dragging this out so much, but IMO these tests were maybe only a little odd before, but quite a noticeable exception now.

I know this PR has been going for a long time, but since it's close to being completed now I think that it would be good to polish these tests while we are at it and avoid pushing this `testDAO` to `master` to begin with.

src/registryDAO_storage.mligo
pasqu4le: :+1: looks much better to me now, let's go with this

typescript/baseDAO/src/api.ts
pasqu4le: :thinking: fair enough, that's a good point.

never mind then

docs/registry.md
pasqu4le: Come to think of it, should these be changed to `big_map`s instead? I think that there is the possibility that either of these `map`s will grow too much, making the contract unusable, is that correct?

I see that this wasn't quire introduced here, so I'm fine with leaving this for a separate issue.

However I was also referring to the rest of the doc too, where I see that much of the explanation revolves around arbitrary `k` and `v` types that are supposed to be represented by `bytes`, but that's actually not the case (AFAICT they are `string`s and in fact have been so for a long time). Am I correct? Can you please fix that?

src/testDAO.mligo
sras: No problem. I have removed the test variant and replaced its use with TrivialDAO. I had to remove a couple of tests, which does not appear to test anything significant.

docs/building.md
```diff
@@ -11,19 +11,78 @@ in particular its CameLIGO dialect.
...
+  | MyCustomEp1 of (nat, nat)
+  | MyCustomEpDummy of unit
+```
+
+#### Building the variant
+
+After defining implementation.mligo and storage.mligo files for your variant, you can build it by using the make command:
+
+```
+export VARIANT=acme && make baseDAO
```
pasqu4le, line 83: shouldn't this be:
```suggestion
make out/acmeDAO.tz
```
now?

haskell/package.yaml
```diff
@@ -12,7 +12,11 @@ copyright:           2021 TQ Tezos
 extra-doc-files:
 - README.md
 extra-source-files:
-- resources/baseDAO.tz
+- resources/registryDAO.tz
+- resources/treasuryDAO.tz
+- resources/trivialDAO.tz
+- resources/testDAO.tz
```
pasqu4le, line 18:
```suggestion
```

haskell/baseDAO-ligo-meta.cabal
```diff
@@ -14,7 +14,11 @@ license:        MIT
 license-file:   LICENSE
 build-type:     Simple
 extra-source-files:
-    resources/baseDAO.tz
+    resources/registryDAO.tz
+    resources/treasuryDAO.tz
+    resources/trivialDAO.tz
+    resources/testDAO.tz
```
pasqu4le, line 20:
```suggestion
```

haskell/test/SMT/BaseDAO.hs
```diff
@@ -33,66 +29,30 @@ hprop_SMT =
...
-        # swap
-        # dip (push Nothing)
-        # constructStack @(DecisionLambdaOutput)
-      }
+  { fsStorage = (fsStorage fs) { sExtra = () }
   }
+  -- where
+    -- ce = fillConfig dummyDecisionCallback $
+    --   fillConfig (divideOnRejectionBy 2) $
+    --     fillConfig (proposalFrozenTokensMinBound 10) dynRecUnsafe
```
pasqu4le, line 53: a leftover?

pasqu4le (commented):
I only have a few minor comments left (and the open one above), but otherwise LGTM.

@sras could you please clean up the commit history once again as well?

docs/registry.md
pasqu4le: @sras not sure if you saw the comment just above

sras: I have included that in the built in registry, `k` and `v` are set as strings. Is that good enough? Or should we remove mentioning these as parameteric and use `string` everywhere?

docs/building.md
sras: :facepalm: Fixed.

haskell/test/SMT/BaseDAO.hs
sras: Yes. removed now.

docs/registry.md
pasqu4le: Yep, that work for me.

I still think that we should consider switching to use `big_map`, but I think that can be left for a separate issue/MR to avoid pushing this one even further.

pasqu4le (approved):
LGTM now :ok_hand:

docs/registry.md
sras: Have created this issue https://github.com/tezos-commons/baseDAO/issues/322

PR #307, opened 2021-11-18
[#306] Add proposal to update contract delegate for registry/treasury

### LINKED ISSUE #306: Add proposal type to delegate DAO's XTZs

We'vw been asked for the ability for both `registryDAO` and `treasuryDAO` to delegate the DAO XTZs to a baker.

In order to do this we'll need to use the [`SET_DELEGATE`](https://tezos.gitlab.io/michelson-reference/#instr-SET_DELEGATE) instructions and add a new proposal type that will allow updating (setting or removing) a delegate for the DAO.

# Acceptance criteria

Both `registryDAO` and `treasuryDAO` have a proposal type that uses `SET_DELEGATE` to update the DAO's delegate

sras:
## Description

We need to add a proposal type that updates the contract
delegate. In this PR we Update the proposal metadata type of both contracts with a new
branch that represents a proposal to update the contract delegate.

Tests for this is missing the final check. But I have tested this on localchain using the newly added nettest (after some small changes).

## Related issue(s)

Resolves #306

- Tests
        silently reappearing again.

- Documentation
    - Haddock


pasqu4le (commented):
Aside from the tests that are waiting for the dependency update, this overall LGTM,

One more missing thing however is that the `docs/registry.md` and `docs/treasury.md` need an update

haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -779,7 +782,7 @@ mkConfig customEps votingPeriod fixedProposalFee maxChangePercent changePercent
   }
 
 defaultConfig :: Config
-defaultConfig = mkConfig [] 10 0 19 5 500
+defaultConfig = mkConfig [] (Period 20) (FixedFee 0) 19 5 (GovernanceTotalSupply 500)
```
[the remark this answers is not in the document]
sras: This is how I got the CI tests to pass. I am not sure if it makes sense, but I doubled the period duration because the block time was halfed post granada protocol. So I thought may be it is causing the timing overflows, and it appear to have worked, at least for the timebeing...

docs/registry.md
```diff
@@ -43,6 +43,8 @@ registry. This includes:
    - `nat %agoraPostID` is used to refer to an Agora post explaining the proposed transfer and/or changes and motivation for them.
 4. Proposal to update the guardian address in the BaseDAO contract:
    - This proposal takes an address parameter and use it to update the guardian address in the storage.
+5. Proposal to update the contract delegate:
+   - This proposal takes a keyhash parameter and use it to update the delegate address of the contract.
```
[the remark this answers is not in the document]
sras: Done

docs/treasury.md
```diff
@@ -15,6 +15,8 @@ Its `proposal_metadata` contains proposal types:
    - `nat %agoraPostID` is used to refer to an Agora post explaining the proposed transfer and motivation for it.
 - Proposal to update the guardian address in the BaseDAO contract.
    - This proposal takes an address parameter and use it to update the guardian address in the storage.
+- Proposal to update the contract delegate.
+   - This proposal takes a keyhash parameter and use it to update the delegate address of the contract.
```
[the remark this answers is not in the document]
sras: Done

haskell/src/Ligo/BaseDAO/Types.hs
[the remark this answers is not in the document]
rinn7e: I keep getting the tests fail in localchain too, not sure why. The rests look good to me.

rinn7e (approved):
I keep getting the tests fail in localchain too,

haskell/test/Test/Ligo/RegistryDAO.hs
```diff
@@ -464,13 +464,13 @@ test_RegistryDAO =
                 , [ FA2.TransferItem { tiFrom = wallet1, tiTxs = [FA2.TransferDestination { tdTo = unTAddress baseDao, tdTokenId = FA2.theTokenId, tdAmount = proposalSize }] } ] -- governance token transfer for freeze
               ]
 
-    , nettestScenarioCaps "checks it can transfer with receive_xtz_entrypoint (#66)" $
+    , testScenario "checks it can transfer with receive_xtz_entrypoint (#66)" $ scenario $
```
pasqu4le, line 467: the test name seems outdated, given that `receive_xtz` was removed

haskell/test/SMT/RegistryDAO.hs
```diff
@@ -47,8 +47,7 @@ hprop_RegistryDaoSMT =
       , soRejectedProposalSlashValue = registryDaoRejectedProposalSlashValue
       , soDecisionLambda = registryDaoDecisionLambda
       , soCustomEps = Map.fromList
-          [ ( unsafeMkMText "receive_xtz", \_ -> pure () )
```
pasqu4le, line 50: I see that you have removed `receive_xtz` from the custom entrypoints in the tests, but not in the ligo code. Was this intentional?

I would imagine that with `default` we no longer need it at all.

Related: both in `docs/registry.md` and `docs/treasury.md` there is an explanation of the `receive_xtz` entrypoint left behind.

haskell/src/Ligo/BaseDAO/Types.hs
```diff
@@ -527,6 +527,7 @@ data AllowXTZParam
   | Transfer_contract_tokens TransferContractTokensParam
   | Transfer_ownership TransferOwnershipParam
   | Accept_ownership ()
+  | Default ()
```
pasqu4le, line 532: This commit, e4149eccb8174f87ab4d8e5e97c2d0d452617349, lacks the `Problem`/`Solution` in the commit message

cabal.project
```diff
@@ -1,6 +1,6 @@
 -- Generated by stackage-to-hackage
 
-with-compiler: ghc-8.10.4
+with-compiler: ghc-8.10.7
```
pasqu4le, line 3: This commit, 8881558351ca85313a0746c4be5b40c727292ac0, also lacks `Problem` and `Solution` in its message as well as an issue tag (I think `[Chore]` should be used here).

haskell/src/Ligo/BaseDAO/Types.hs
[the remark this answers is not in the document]
pasqu4le: AFAIR the problem we had before with the period time was that, if blocks are not processed very quickly, then when we wait before making a new transfer too much time might pass by (because we waited for much of it in the first call).

It seems that local-chain on hangzhounet is suffering from these delays (and had some more trouble a few days back too) and it appears to me that we are now getting these consistently.

AFAIK (please check/correct me if I'm wrong), then all that this change is that tests will take longer, which is not too bad. So IMO we can get this merged and keep an eye on the local-chain efficiency.

haskell/test/SMT/RegistryDAO.hs
sras: Ah, good catch. I rmember thinking that custom LIGO entrypoints where dynamic (encoding as bytes) and thus does not require removing anything.

sras: I think we also need to mention the `default` entrypoint in the main spec.

sras: Have removed LIGO entrypoints, updated docs and mentioned `default` entrypoint in main spec.

haskell/test/Test/Ligo/RegistryDAO.hs
sras: Changed.

cabal.project
sras: Have included proper message in all commits.

haskell/src/Ligo/BaseDAO/Types.hs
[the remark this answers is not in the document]
pasqu4le: It looks like some tests on local-chain still failed on the last run :disappointed: 

I just restarted the job, if it succeeds I still think that we can merge this and keep an eye on it

pasqu4le (approved):

PR #308, opened 2021-11-22
[#235] Add TZIP-16 Numeric Errors

### LINKED ISSUE #235: TZIP-16 Numeric Errors

The [error messages](https://github.com/tqtezos/baseDAO/blob/master/docs/specification.md#errors) are long `string`'s that largely could be replaced with TZIP-16-style errors, i.e. numeric codes with a TZIP-16 view to lookup their longer-string expansion.

# Acceptance criteria

Aside from [TZIP-12 errors](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#error-handling), which don't currently support [TZIP-16-style errors](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#reserved-fields), all of the error `string`'s in the contract should be replaced with `nat` error codes and the associated TZIP-16 `errors` view should be added.

rinn7e:
## Description

Problem: The error messages are long string's that largely could be replaced
with TZIP-16-style errors, i.e. numeric codes.

Solution: Add TZIP-16 Numeric Errors, update tests.

## Related issue(s)

Resolves #235 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e:
@pasqu4le This PR is ready for review. I used the same generate script from SegCFMM. 

> with a TZIP-16 view to lookup their longer-string expansion.

The issue description includes this and I'm not sure exactly how to do this. From what I can guess, we need to make the generate script include lorentz instr to be used as a view, is this correct?

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,250 @@
...
+errorsEnumerate start =
+  map (\(eiCode, eiLabel :? eiDesc) -> ErrorItem{..}) .
+  filter ((/= removedError) . snd) .
+  zip [start..]
+
+
+invalidInputErrors :: [ErrorItem]
+invalidInputErrors = errorsEnumerate 100
+  -- New errors must be added to the _end_ of this list.
+  -- To remove an error, replace it with 'removedError'.
```
sras, line 41: May be I am missing something, but why can't we just use a `Map` from error code => (error_name, error_description) ?

AFAIU that will save us from putting these restrictions, that quite possibly could be missed when making an update?

rinn7e: I took this from SegCFMM contract: https://github.com/serokell/segmented-cfmm/pull/25

There were some issues about conflicting error codes (2 PRs containing news error code), this would help solving that problem. Also `removedError` helps when we need to delete an error without changing error code of other errors.

sras: > There were some issues about conflicting error codes 

Ah, yes. That makes sense. So in that case, does that project include manually added tests to ensure that certain error codes map to the expected errors?

rinn7e: Not sure why we want to test that. The script generate a haskell and ligo file to be used (example: https://github.com/tezos-commons/baseDAO/blob/rinn7e/%23235-numeric-error/haskell/src/Ligo/BaseDAO/ErrorCodes.hs)

We would want to use the constant from that file (Example: `notAdmin = 100`), instead of using `100` manually.

sras: > Not sure why we want to test that.

But what is preventing someone from not following the procedure and actually removing an error from the list (instead of replacing it with 'removed'?)

rinn7e: Nothing really prevent them, but we can see clearly if we review the code (the error code changes for the errors below the removed error). Maybe @Martoon-00 could help explaining better on this.

sras: > but we can see clearly if we review the code..

Well, if this mixup happens (re-ordering of lines) happen in the final rebase, after the reviews, just before the merge, will it get caught?

rinn7e: I'm not sure how this would happen but if 'removed` is re-order to a different line, we definitely can see the error codes change. I guess this require some attention to check before merging if the reviews are already done.

sras: > we definitely can see the error codes change

AFAIU you need to run a diff after running the generator to see it. But one can always just run the tests after running the generator, and proceed to commit and merge if everything passed in the tests, right? AFAIU it won't be caught in the CI as well.

rinn7e: @sras Sorry, I missed your comment. What is your suggestion to fix this? Should we add a test to ensure the errors are map to proper error codes?

sras: @rinn7e Yes, If we explicitly write the expected error code for each error in the test, I think that should be enough.

rinn7e: @sras I've added the tests, this should be good to go now.

sras (approved):

scripts/generate_error_code.hs.license
```diff
@@ -0,0 +1,2 @@
+SPDX-FileCopyrightText: 2021 Tocqueville Group
```
pasqu4le, line 1: nitpick, but I think that this might be included in the actual file (after the shebang)

docs/error-codes.md
```diff
@@ -0,0 +1,60 @@
...
+
+- SPDX-License-Identifier: LicenseRef-MIT-TQ
+-->
+
+<!--
+NOTE: This file should not be modified directly.
+Use `stack scripts/generate_error_code.hs` instead.
+-->
+
+## Error Codes
```
pasqu4le, line 12: Documenting it in markdown is nice, however the issue specifically asks for `TZIP-16` error codes, which should be included in the metadata, see the `"errors"` in the [reserved fields](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#reserved-fields) section.

I don't see this added in this PR.

pasqu4le (commented):
Note: this also needs a rebase

docs/error-codes.md
rinn7e: I missed the `errors` fields, I've added this now, thanks Pasquale.

Also, we probably want #305 to be merged first, since I rebase this on top of it due to some errors are removed.

scripts/generate_error_code.hs.license
rinn7e: Updated this.

docs/error-codes.md
rinn7e: @pasqu4le Since #305 is probably not merged any time soon, I rebase this PR on top of master instead and add error code for the removed errors for now. Hope to get your re-review on this.

rinn7e: @pasqu4le I've rebased this on the latest master, and fix the CI, let me know if this is good to be merged.

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+errorItemToTzipError :: ErrorItem -> Text
+errorItemToTzipError err@ErrorItem{eiDesc, eiCode} =
+  [i|
+  EStatic $ StaticError
+    { seError = ExpressionInt $ toInteger #{fnName}
+    , seExpansion = ExpressionString "#{eiDesc}"
+    , seLanguages = ["en"]
+    }|]
+  where
+    fnName = haskellFnName err
```
pasqu4le, line 241: @rinn7e does this suit all errors tho?

AFAIU the `error` field here should be the Michelson representations of exactly the error that will be thrown with `FAILWITH`, however here we have also errors that return not just a `nat` but also a `nat * unit`, `nat * string` and more, so this doesn't look correct to me, am I wrong?

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+
+invalidInputErrors :: [ErrorItem]
+invalidInputErrors = errorsEnumerate 100
+  -- New errors must be added to the _end_ of this list.
+  -- To remove an error, replace it with 'removedError'.
+
+  [ "not_admin"
+    :? "The sender is not the administrator"
+  , "not_pending_admin"
+    :? "Authorized sender is not the current pending administrator"
```
pasqu4le, line 49: While we are at it, I know you took these from the previous descriptions, but I have some suggestions about them.
```suggestion
    :? "The sender is not the current pending administrator."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "not_token_owner"
+    :? "Trying to configure operators for a different wallet which sender does not own"
```
pasqu4le, line 51: It looks like this is unused and so should be removed.

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "fail_proposal_check"
+    :? "Throws paired with an additional `string` message when trying to submit a proposal that does not pass `proposal_check`."
```
pasqu4le, line 53:
```suggestion
    :? "Thrown paired with a `string` error message when the proposal does not pass the `proposal_check`."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "frozen_token_not_transferable"
+    :? "Transfer entrypoint is called for frozen token"
```
pasqu4le, line 55: Also seems like it could be removed

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "proposal_not_exist"
+    :? "Throws when trying to vote on a proposal that does not exist or is no longer ongoing."
```
pasqu4le, line 57:
```suggestion
    :? "The proposal does not exist or is no longer ongoing."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "quorum_not_met"
+    :? "A proposal is flushed, but there are not enough votes"
```
pasqu4le, line 59: Also unused?

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "voting_stage_over"
+    :? "Throws when trying to vote on a proposal that is already ended"
```
pasqu4le, line 61:
```suggestion
    :? "The proposal voting stage has already ended."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "max_proposals_reached"
+    :? "Throws when trying to propose a proposal when proposals max amount is already reached"
```
pasqu4le, line 63:
```suggestion
    :? "The maximum amount of ongoing proposals has been reached."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "max_voters_reached"
+    :? "Throws when trying to vote on a proposal when the max voter count of that proposal is already reached"
```
pasqu4le, line 65:
```suggestion
    :? "The maximum amount of voters has been reached for the proposal."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "forbidden_xtz"
+    :? "Throws when some XTZ was received as part of the contract call"
```
pasqu4le, line 67:
```suggestion
    :? "Transfer of XTZ is forbidden on this entrypoint."

File: scripts/generate_error_code.hs [hunk 15]
```diff
@@ -0,0 +1,280 @@
...
+  , "proposer_not_exist_in_ledger"
+    :? "Expect a proposer address to exist in Ledger but it is not found"
```
pasqu4le, line 69: unused too

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "proposal_not_unique"
+    :? "Trying to propose a proposal that is already existed in the Storage."
```
pasqu4le, line 71:
```suggestion
    :? "The submitted proposal already exist."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "missigned"
+    :? "Parameter signature does not match the expected one - for permits."
+  , "unpacking_failed"
+    :? "Throw when unpacking of a packed value."
```
pasqu4le, line 75:
```suggestion
    :? "The unpacking of a submitted value failed."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "unpacking_proposal_metadata_failed"
+    :? "Throw when unpacking of a proposal metadata fails."
```
pasqu4le, line 77:
```suggestion
    :? "The unpacking of a proposal metadata failed."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "missing_value"
+    :? "Throw when trying to unpack a field that does not exist."
```
pasqu4le, line 79:
```suggestion
    :? "A required field value was not found."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "not_proposing_stage"
+    :? "Throw when `propose` call is made on a non-proposing period."
```
pasqu4le, line 81:
```suggestion
    :? "Proposals cannot be submitted in non-proposing stages."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "not_enough_frozen_tokens"
+    :? "Throw when there is not enough frozen tokens for the operation."
```
pasqu4le, line 83:
```suggestion
    :? "There aren't enough frozen tokens for the operation."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "bad_token_contract"
+    :? "Throw when the token contract is not of expected type."
```
pasqu4le, line 85:
```suggestion
    :? "The governance token contract does not have the expected entrypoints."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "bad_view_contract"
+    :? "Throw when the contract for a view entrypoint is not of the expected type."
+  , "drop_proposal_condition_not_met"
+    :? "Throw when calling `drop_proposal` when the sender is not proposer or guardian and proposal is not expired."
```
pasqu4le, line 89:
```suggestion
    :? "The conditions to drop this proposal are not met."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "expired_proposal"
+    :? "Throw when trying to `flush` expired proposals."
```
pasqu4le, line 91:
```suggestion
    :? "The proposal has expired and can no longer be flushed."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "empty_flush"
+    :? "Thrown when trying to `flush` with no available proposals."
```
pasqu4le, line 93:
```suggestion
    :? "There are no available proposals to flush."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "not_delegate"
+    :? "Throw when calling `propose` and `vote` with the sender that is not `from` or delegate of `from`."
```
pasqu4le, line 95:
```suggestion
    :? "The sender has not been delegated the control of the required tokens."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "bad_entrypoint_parameter"
+    :? "Throw when calling an entrypoint with invalid parameters."
```
pasqu4le, line 97: I think that this could be removed too (see next comment)

src/proposal.mligo
```diff
@@ -315,7 +316,7 @@ let handle_proposal_is_over
 // Flush all proposals that passed their voting stage.
 let flush(n, config, store : nat * config * storage): return =
   if n = 0n
-  then (failwith("BAD_ENTRYPOINT_PARAMETER") : return)
+  then (failwith bad_entrypoint_parameter : return)
```
pasqu4le, line 319: This is now the only use left of this error, but given that we also have `empty_flush` IMO we could throw that one here as well and remove the need for this `bad_entrypoint_parameter` error.

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "fail_decision_lambda"
+    :? "Executing decision lambda results in failure."
```
pasqu4le, line 99:
```suggestion
    :? "Executing the propossal's decision lambda results in failure."
```

scripts/generate_error_code.hs
```diff
@@ -0,0 +1,280 @@
...
+  , "entrypoint_not_found"
+    :? "Throw when `CallCustom` is called with a non-existing entrypoint."
```
pasqu4le, line 101:
```suggestion
    :? "The chosen custom entrypoint does not exist."
```

src/proposal/freeze_history.mligo
```diff
@@ -11,8 +12,8 @@ let sub_frozen_fh (amt, fh : nat * address_freeze_history): address_freeze_histo
   match is_nat(fh.past_unstaked - amt) with
   | Some new_amt -> { fh with past_unstaked = new_amt }
   | None ->
-      ([%Michelson ({| { FAILWITH } |} : (string * unit) -> address_freeze_history)]
-        ( "NOT_ENOUGH_FROZEN_TOKENS", ()) : address_freeze_history)
+      ([%Michelson ({| { FAILWITH } |} : (nat * unit) -> address_freeze_history)]
+        ( not_enough_frozen_tokens, ()) : address_freeze_history)
```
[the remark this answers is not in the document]
rinn7e: I've removed all the `(nat * unit)`, and replace it with `nat` and fixed the tests.

scripts/generate_error_code.hs
rinn7e: @pasqu4le  That make sense, I missed this.

I've added a condition in case of a pair:
```
errorItemToTzipError :: ErrorItem -> Text
errorItemToTzipError err@ErrorItem{eiDesc, eiCode, eiArg} =
  [i|
  EStatic $ StaticError
    { seError = #{error}
    , seExpansion = ExpressionString "#{eiDesc}"
    , seLanguages = ["en"]
    }|]
  where
    fnName = haskellFnName err
    (error :: Text) = case eiArg of
      NoArg ->
        [i|ExpressionInt $ toInteger #{fnName}|]
      ArgString val ->
        [i|ExpressionSeq [ ExpressionInt $ toInteger #{fnName}, ExpressionString "#{val}" ]|]
   ```
   
   Not sure if using `ExpressionSeq` for pair works here, hope to get your advice on this

File: scripts/generate_error_code.hs (hunk 5 above)
Comment (rinn7e) [reply scripts/generate_error_code.hs]: I've updated this and all other errors description as well.

File: scripts/generate_error_code.hs (hunk 6 above)
Comment (rinn7e) [reply scripts/generate_error_code.hs]: Removed.

File: src/proposal.mligo (hunk 28 above)
Comment (rinn7e) [reply src/proposal.mligo]: Removed and replace with `empty_flush`.

File: scripts/generate_error_code.hs (hunk 4 above)
Comment (pasqu4le) [reply scripts/generate_error_code.hs]: @rinn7e IIRC `ExpressionSeq` should work alright, although I think a `PrimExpr` for `Pair` woul be better.

However the thing is that we shouldn't really be doing that IMO, we have `toExpression` in `morley` to do this conversion.
This can be used on typed `Value`, which in turn can be obtained using `toVal` (from `IsoValue`).

Doing this instead should be the better option

File: src/management.mligo [hunk 32]
```diff
@@ -45,8 +46,8 @@ let call_custom(param, store, config : custom_ep_param * storage * config) : ret
   let packed_ep =
     match Map.find_opt ep_name config.custom_entrypoints with
     | Some (ep_code) -> ep_code
-    | None -> ([%Michelson ({| { FAILWITH } |} : string -> bytes)] "ENTRYPOINT_NOT_FOUND" : bytes)
+    | None -> ([%Michelson ({| { FAILWITH } |} : nat -> bytes)] entrypoint_not_found : bytes)
```
pasqu4le, line 49:
```suggestion
    | None -> (failwith entrypoint_not_found : bytes)
```

src/management.mligo
```diff
@@ -45,8 +46,8 @@ let call_custom(param, store, config : custom_ep_param * storage * config) : ret
...
   in
   match ((Bytes.unpack packed_ep) : (bytes * full_storage -> return) option) with
   | Some lambda -> lambda (packed_param, (store, config))
-  | None -> ([%Michelson ({| { FAILWITH } |} : string -> return)] "UNPACKING_FAILED" : return)
+  | None -> ([%Michelson ({| { FAILWITH } |} : nat -> return)] unpacking_failed : return)
```
pasqu4le, line 53:
```suggestion
  | None -> (failwith unpacking_failed : return)
```

scripts/generate_error_code.hs
rinn7e: I forgot about `toExpression`, pushed an update to use that instead now.

src/management.mligo
rinn7e: Updated.

src/management.mligo
rinn7e: Updated.

pasqu4le (approved):
LGTM now :ok_hand: don't forget to squash your commits

PR #310, opened 2021-12-01
[#240] Support TZIP-21 in the generated contract metadata

### LINKED ISSUE #240: Support TZIP-21 in the generated contract metadata

We currently generate contract metadata that is compliant to [TZIP-16](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md) but not [TZIP-21](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-21/tzip-21.md).

Since many wallets prefer/value support for TZIP-21 we should include it in baseDAO as well.

# Acceptance criteria

The generated contract metadata is TZIP-21 compliant.

rinn7e:
## Description

Problem: There is a new TZIP-21 which is an extension of TZIP-16
that we should support.

Solution: Add support for TZIP-21.

## Related issue(s)

Resolves #240

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -32,7 +32,11 @@ You can generate the known metadata, but you'll need the `stack` tool installed,
 [The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/README/) tutorial for instructions on how to obtain it.
 
 ```bash
-make metadata --output=metadata.json
+make metadata \
+    frozen_token_symbol=frozen_token \
+    frozen_token_name="BaseDAO Frozen Token" \
+    frozen_token_thumbnail_uri="ipfs://QmV3a1TAdCncfs84Gi9msDsDJVQBDt6Wb5gJRVuFRfrgtG" \
+    output=metadata.json
 ```
 You can then modify the produced `metadata.json` if you want to add more information.
 
--- a/haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
+++ b/haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
@@ -16,20 +16,24 @@ module Ligo.BaseDAO.TZIP16Metadata
 
 import qualified Universum as U
 
+import Data.Aeson.TH (deriveJSON)
 import Data.Version (showVersion)
 
 import Lorentz hiding (View)
 import Lorentz.Contracts.Spec.TZIP16Interface
-import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Morley.Metadata
+import Morley.Metadata.Util.Aeson (aesonOptions)
 
 import Ligo.BaseDAO.Types
 import qualified Paths_baseDAO_ligo_meta as Paths
 
 -- | Piece of metadata defined by user.
 data MetadataConfig = MetadataConfig
-  { mcFrozenTokenMetadata :: FA2.TokenMetadata
-  }
+  { mcName :: Text
+  , mcSymbol :: Text
+  , mcDecimals :: Integer
+  , mcThumbnailUri :: Maybe Text
+  } deriving stock (Eq, Show)
 
 -- | All the information for instantiating metadata.
 --
@@ -43,9 +47,12 @@ data MetadataSettings = MetadataSettings
 defaultMetadataConfig :: MetadataConfig
 defaultMetadataConfig =
   MetadataConfig
-    { mcFrozenTokenMetadata =
-        FA2.mkTokenMetadata "frozen_token" "Frozen Token" "8"
+    { mcName = "BaseDAO Frozen Token"
+    , mcSymbol = "frozen_token"
+    , mcDecimals = 0
+    , mcThumbnailUri = Nothing
     }
+
 
 -- | Construct MetadataSetting for the contract.
 mkMetadataSettings :: MetadataConfig -> MetadataSettings
@@ -56,13 +63,21 @@ mkMetadataSettings msConfig = MetadataSettings
 -- | Parts of metadata that can be filled just knowing the contract,
 -- without user's input.
 knownBaseDAOMetadata :: MetadataSettings -> Metadata (ToT Storage)
-knownBaseDAOMetadata settings = mconcat
-  [ version . fromString $ showVersion Paths.version
+knownBaseDAOMetadata settings@MetadataSettings{..} = mconcat
+  [ name "Base DAO"
+  , description "A generic smart contract on Tezos that enables a community to collectively govern resources, registries, or rules."
+  , version . fromString $ showVersion Paths.version
   , license $ License "MIT" Nothing
   , authors [Author "Serokell", Author "Tocqueville Group"]
   , homepage "https://github.com/tezos-commons/baseDAO"
-  , interfaces [tzip 17]
+  , interfaces
+      [ tzip 16
+      , tzip 21 -- 1728fcfe0ac90463ef15e6a994b6d6a15357e373
+      ]
   , views (baseDAOViews settings)
+  , keyValue @[MetadataConfig] "assets"
+      [ msConfig
+      ]
   ]
 
 ------------------------------------------------------------------------
@@ -107,3 +122,9 @@ permitsCounterView MetadataSettings{} = View
               stToField #sPermitsCounter
       ]
   }
+
+----------------------------------------------------------------------------
+-- JSON serializers/deserializers
+----------------------------------------------------------------------------
+
+deriveJSON aesonOptions ''MetadataConfig
--- a/Makefile
+++ b/Makefile
@@ -256,14 +256,20 @@ haskell-resources: test-storage all
 	mkdir -p haskell/resources
 	cp $(OUT)/* haskell/resources
 
+metadata : frozen_token_symbol = frozen_token
+metadata : frozen_token_name = Frozen Token
+metadata : frozen_token_thumbnail_uri =
 metadata : output = metadata.json
 metadata: haskell-resources
 	$(MAKE) -C haskell build PACKAGE=baseDAO-ligo-meta \
 		STACK_DEV_OPTIONS="--fast --ghc-options -Wwarn"
 
 	$(MAKE) -C haskell exec PACKAGE=baseDAO-ligo-meta \
-		EXEC_ARGUMENTS=print-metadata \
-		EXEC_OUTPUT=$(output)
+		EXEC_ARGUMENTS="print-metadata \
+		--frozen-token-symbol $(frozen_token_symbol) \
+		--frozen-token-name \"$(call escape_double_quote,$(frozen_token_name))\" \
+		--frozen-token-thumbnail-uri \"$(call escape_double_quote,$(frozen_token_thumbnail_uri))\" \
+		" EXEC_OUTPUT=$(output)
 
 test: haskell-resources
 	$(MAKE) -C haskell test PACKAGE=baseDAO-ligo-meta
--- a/haskell/app/Main.hs
+++ b/haskell/app/Main.hs
@@ -9,12 +9,10 @@ module Main
 import Universum
 
 import Data.Aeson.Encode.Pretty (encodePretty)
-import Fmt (pretty)
 import Data.Version (showVersion)
 import qualified Options.Applicative as Opt
 import Paths_baseDAO_ligo_meta (version)
 
-import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Util.Main
 
 import Ligo.BaseDAO.Types (Parameter)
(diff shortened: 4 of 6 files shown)


rinn7e:
@pasqu4le From what I understand TZIP-21 is an extension of TZIP-12 and TZIP-16. Basically it provides extra information about the tokens (call `asset` in the docs) in the contract.

Since our contract is TZIP-16 compliance only, I added a fields `assets` containing some info about the frozen token.
> If provided at the contract metadata level, to prevent pollution in the top level of the
metadata, content should be provided as a nested object under the key assets.

I only added `description` and `creators` field for now for the frozen token asset. Let me know if I should add more.

One confusion that I have is that, it seems like TZIP-21 targets contract that is either an FA2 token contract or a contract that manages FA2 tokens. Since we only have frozen token which is more like a special kind of token (cannot be `transfer` or `mint`), I'm not sure if it make sense to support TZIP-21. Probably the need for support would be on the governance contract instead. Hope to get your view on this.

pasqu4le:
@rinn7e AFAICT `TZIP-21` is a specification that extends `TZIP-16` for any token's metadata.

It has been clearly written with FA2/TZIP-12 in mind (which makes sense, given that this is a token standard), but it doesn't seem to require that and can be used for any token (in fact, it also mentions FA1.2 ones).

Here in baseDAO, the main tokens (contained withing the contract itself) are `frozen` tokens, so we want to have metadata for those.

Following the [Base Token Metadata Standard](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-21/tzip-21.md#base-token-metadata-standard) section we need to have fields for:
1. `name` or `symbol` or both (but at least one of them)
2. `decimals`

I suggest going with `name` for point number 1.

`frozen` tokens are also fungible, so following the [Fungible Token Recommendations](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-21/tzip-21.md#fungible-token-recommendations) section, we should also add, if possible:
- [shouldPreferSymbol](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-21/tzip-21.md#shouldprefersymbol-boolean-default-false)
- [thumbnailUri](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-21/tzip-21.md#thumbnailuri-string-format-uri-reference)

_Note: for `thumbnailUri` I think we should ask on the shared slack channel, as some graphic already exist and this may very well have been uploaded to a fixed location already._

AFAIK these fields should cover everything we need (I see you added more that seem incorrect to me as they are about NFTs or other kinds of tokens, but I may be wrong).

I noticed you added an `assets` sub-field too, but that doesn't seem to be needed, as per [documentation](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-21/tzip-21.md#asset-object):
> Properties of the `asset` object are designed to live at the root level of the token metadata, or as an object array under the key `assets`. at the contract metadata level.

we only have one "asset" here, so there is no need for an array of them and instead we can use the root of the JSON instead (as you did for some fields already).

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
```diff
@@ -57,12 +70,27 @@ mkMetadataSettings msConfig = MetadataSettings
...
+  , keyValue @[MetadataAsset] "assets"
+      [ MetadataAsset
+          { maName = "Frozen Token"
+          , maDecimals = 0
+          , maDescription = "Frozen tokens are tokens that are required to be able to propose and vote on proposals."
+          , maShouldPreferSymbol = False
+          , maThumbnailUri = ""
+          , maCreators = ["TQ Tezos"]
+          }
+      ]
```
rinn7e, line 93: @pasqu4le I've added more fields are you have suggested.

> for thumbnailUri I think we should ask on the shared slack channel, as some graphic already exist and this may very well have been uploaded to a fixed location already.

I couldn't find the thumbnail uri, so I leave it blank for now.

This is what the full metadata.json looks like, hope to get your review on this:

```
{
    "homepage": "https://github.com/tezos-commons/baseDAO",
    "interfaces": [
        "TZIP-16",
        "TZIP-21"
    ],
    "views": [
        {
            "implementations": [
                {
                    "michelsonStorageView": {
                        "returnType": {
                            "args": [
                                {
                                    "prim": "address",
                                    "annots": [
                                        "%address"
                                    ]
                                },
                                {
                                    "prim": "nat",
                                    "annots": [
                                        "%token_id"
                                    ]
                                }
                            ],
                            "prim": "pair"
                        },
                        "code": [
                            {
                                "prim": "CAR"
                            },
                            {
                                "prim": "CDR"
                            },
                            {
                                "prim": "CDR"
                            },
                            {
                                "prim": "CAR"
                            }
                        ]
                    }
                }
            ],
            "name": "governance_token",
            "pure": true,
            "description": "Return the address and token_id of the associated FA2 contract."
        },
        {
            "implementations": [
                {
                    "michelsonStorageView": {
                        "returnType": {
                            "prim": "nat"
                        },
                        "code": [
                            {
                                "prim": "CDR"
                            },
                            {
                                "prim": "CAR"
                            },
                            {
                                "prim": "CDR"
                            },
                            {
                                "prim": "CAR"
                            }
                        ]
                    }
                }
            ],
            "name": "GetCounter",
            "pure": true,
            "description": "Returns the next counter value with which a permit should be created."
        }
    ],
    "authors": [
        "Serokell",
        "Tocqueville Group"
    ],
    "name": "Base DAO",
    "version": "0.4.0.0",
    "license": {
        "name": "MIT"
    },
    "assets": [
        {
            "shouldPreferSymbol": false,
            "name": "Frozen Token",
            "creators": [
                "TQ Tezos"
            ],
            "thumbnailUri": "",
            "decimals": 0,
            "description": "Frozen tokens are tokens that are required to be able to propose and vote on proposals."
        }
    ],
    "description": "A generic smart contract on Tezos that enables a community to collectively govern resources, registries, or rules."
}
```

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
```diff
@@ -39,6 +42,16 @@ data MetadataSettings = MetadataSettings
   { msConfig :: MetadataConfig
   }
 
+-- | TZIP-21 assets field.
+data MetadataAsset = MetadataAsset
+  { maName :: Text
+  , maDecimals :: Integer
+  , maDescription :: Text
+  , maShouldPreferSymbol :: Bool
+  , maThumbnailUri :: Text
```
pasqu4le, line 51: since this is optional (and has no default value when unspecified) I believe it should be a
```suggestion
  , maThumbnailUri :: Maybe Text
```
and the field should not appear in the resulting JSON when `Nothing`

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
```diff
@@ -39,6 +42,16 @@ data MetadataSettings = MetadataSettings
...
+  , maCreators :: [Text]
```
pasqu4le, line 52: It seems that this and `mdDescription` are suggested in `TZIP-21` for [Semi-fungible tokens and NFTs](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-21/tzip-21.md#semi-fungible-and-nft-token-recommendations), given that frozen tokens are fungible I would remove both.

One reason is that their presence might be confusing to indexers and other tools, the other is that I'm not sure that `creators` should be the same as the contract authors (this field especially seems to be more suited for unique tokens on the blockchain).

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
```diff
@@ -57,12 +70,27 @@ mkMetadataSettings msConfig = MetadataSettings
...
 knownBaseDAOMetadata :: MetadataSettings -> Metadata (ToT Storage)
 knownBaseDAOMetadata settings = mconcat
-  [ version . fromString $ showVersion Paths.version
+  [ name "Base DAO"
+  , description "A generic smart contract on Tezos that enables a community to collectively govern resources, registries, or rules."
+  , version . fromString $ showVersion Paths.version
   , license $ License "MIT" Nothing
   , authors [Author "Serokell", Author "Tocqueville Group"]
   , homepage "https://github.com/tezos-commons/baseDAO"
-  , interfaces [tzip 17]
```
pasqu4le, line 64: :+1:

haskell/src/Ligo/Util.hs
```diff
@@ -40,3 +43,6 @@ fetchValue rawPath = do
         either (error . pretty) fromVal $ readValue "" valueLiteral
       |]
 
+
+aesonOptions :: A.Options
+aesonOptions = (A.aesonPrefix A.camelCase) { A.omitNothingFields = True}
```
pasqu4le, line 48: This seems exactly the same as the `aesonOptions` defined in `Morley.Metadata.Util.Aeson`.

Why not import and use that one?

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
pasqu4le: One more thing: I think that this should be given as an option for users to specify if they want to.

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
pasqu4le: Aside from the comments I left below it LGTM.

> I couldn't find the thumbnail uri, so I leave it blank for now.

It's ok, on second thought we probably don't want to use baseDAO's (or homebase's) logo for everyone :ok_hand:

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
rinn7e: I've make it configurable now, also updated the makefile as well.

haskell/src/Ligo/Util.hs
rinn7e: Make sense, I've removed this and use the one from `Moley.Metadata` instead.

pasqu4le (approved):

docs/building.md
```diff
@@ -78,7 +78,12 @@ You can generate the known metadata, but you'll need the `stack` tool installed,
 [The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/README/) tutorial for instructions on how to obtain it.
 
 ```bash
-make metadata --output=metadata.json
+make metadata \
+    frozen_token_symbol=frozen_token \
+    frozen_token_name="Frozen Token" \
+    frozen_token_decimals=0 \
```
sras, line 84: Does it makes sense to leave this configurable?

docs/building.md
```diff
@@ -78,7 +78,12 @@ You can generate the known metadata, but you'll need the `stack` tool installed,
 [The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/README/) tutorial for instructions on how to obtain it.
 
 ```bash
-make metadata --output=metadata.json
+make metadata \
+    frozen_token_symbol=frozen_token \
+    frozen_token_name="Frozen Token" \
```
sras, line 83: May be it is better to write down "BaseDAO Frozen Token" here, as a token name of "Frozen Token" might appear wierd without context, when it is displayed in a wallet or some other device

sras (approved):
LGTM apart from minor things.

docs/building.md
rinn7e: @sras Ideally the decimal would be the same as the governance token. AFAIU decimal is useful for UI rendering even though we don't really use it in contract logic. So I think it is fine leave it as configurable. @pasqu4le Hope to get your view on this as well.

docs/building.md
rinn7e: Updated this.

docs/building.md
sras: > @sras Ideally the decimal would be the same as the governance token.

Wait, do we have that kind of logic in our contract? So if the governance token have decimals of `2`, does our contract requires freezing 100 governance tokens to put down one vote?

I am not sure what the expectation of such a participant. So say I have 5 governance tokens. In my wallet, it is shown as `5.00`, and suppose I want to vote in a DAO proposal. So call `freeze` on the BaseDAO, and aftewords, my wallet shows the balance of my governance token as `4.99`.  But in the same wallet, in my BaseDAO account, the balance will be shown as `1.00`, (because the decimals is 2 in baseDAO as well), Right?

Does that make sense? Or am I wrong somewhere?

rinn7e: >  So say I have 5 governance tokens. In my wallet, it is shown as 5.00, and suppose I want to vote in a DAO proposal. 

AFAIU, it will be shown as `0.05` for governance token if decimal is purely for rendering purpose. So in BaseDAO it will be shown as `0.01`. Not sure if I explain this clearly.

sras: >AFAIU, it will be shown as 0.05 for governance token

Yes, I mean, 5 tokens, as displayed by the wallet. So it would be a value of `500` in the contact storage (if the decimals is 2)

pasqu4le: I agree with @sras, we should probably leave this at `0` and remove the option here from the makefile.

While the translation would work, there is still the problem that the `docs` are based on there being no decimals, for example, from the spec:
> Each vote stakes one frozen token

and
> - The voter MUST have frozen tokens equal to `vote_amount` or more (1 unstaked frozen token is needed for 1 vote) from past `stage`s.

Is "one frozen token" here equal to `1` or `1.0` (aka `100`) tokens? I think this might be confusing for people reading.

`homebase` as well is built with unitary tokens in mind AFAIK, so that would cause even more trouble IMO.

rinn7e: Got this, I've made this change now.

haskell/src/Ligo/BaseDAO/TZIP16Metadata.hs
rinn7e: Hmm, I replace this with tzip-16 by mistake. Should I add it back? cc @pasqu4le

pasqu4le: I noticed this and though it was intentional.

I checked and it seemed that indeed we don't properly support TZIP-17 (at least not the current spec), but rather a mechanism similar to it (please correct me if I'm wrong), so I guess it's ok to remove this.

pasqu4le (approved):
AFAIU this can be merged now then

PR #313, opened 2021-12-13
[#311] Remove voters limit and avoid returning tokens on flush 

### LINKED ISSUE #311: Remove voters limit and avoid returning tokens on flush

Currently as part of proposal `flush`ing the contract transfers (aka unstakes) back tokens used to vote on said proposal.
This is also the reason why there is a limit on the number of `voters` that can vote on any single proposal as, if they are too many, the call to `flush` would fail due to gas costs.

This behavior has been in the contract since basically its conception, and the logic itself relied on it in various places during its development, however currently that's no longer the case.
Moreover, this is actually problematic, because the cost of proposals handling increases linearly with the number of voters.

So, we want to to change this behavior to remove the voters limit and avoid returning tokens on `flush`.

# Acceptance criteria

1. on a `flush` call no token is returned to the `voters` (only to the proposer)
2. a new entrypoint exists that voters can call to get back the tokens staked on a proposal (after it was `flush`ed)
3. the limit on the number of voters per proposals is removed entirely
4. tests are updated accordingly

rinn7e:
## Description

Problem: The cost of proposals handling increases linearly with the
number of voters due to the fact that we unfreeze voters token on flush.

As a result, we want to to change this behavior to remove the voters
limit and avoid returning tokens on flush.

Solution: Remove returning voter tokens, remove voter limit, add
`Unfreeze_staked` entrypoints, and update the tests.

## Related issue(s)

Resolves #311

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -94,9 +94,6 @@ type config =
 
   ; max_proposals : nat
   // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
-  ; max_voters : nat
-  // ^ Determine the maximum number of voters that are allowed to vote on a proposal.
-  // Voters who votes on a proposal, both ways are counted twice.
   ; max_quorum_threshold : quorum_fraction
   // ^ Determine the maximum value of quorum threshold that is allowed.
   ; min_quorum_threshold : quorum_fraction
@@ -154,8 +151,6 @@ type proposal =
   // ^ address of the proposer
   ; proposer_frozen_token : nat
   // ^ amount of frozen tokens used by the proposer, exluding the fixed fee
-  ; voters : voter list
-  // ^ voter data
   ; quorum_threshold: quorum_threshold
   // ^ quorum threshold at the cycle in which proposal was raised
   }
@@ -304,10 +299,10 @@ Moreover the proposal to vote on must have been submitted in the proposing `stag
 immediately preceding and the voter must have frozen his tokens in one of the
 preceding `stage`s.
 
-Each vote stakes one frozen token. Staked tokens cannot be unfreezed till they
-are unstaked when the associated proposal is flushed. The number of staked tokens
-only depend on the number of votes, and does not depend on whether the vote is in
-favor or against a proposal.
+Each vote stakes one frozen token. Staked tokens cannot be unfreezed till the
+associated proposal is flushed or dropped. The tokens are unstaked by calling `unstake_vote`.
+The number of staked tokens only depend on the number of votes, and does not depend
+on whether the vote is in favor or against a proposal.
 
 It's possible to vote positively or negatively.
 After the voting ends, the contract is "flushed" by calling a dedicated entrypoint.
@@ -330,6 +325,7 @@ Full list:
 * [`drop_proposal`](#drop_proposal)
 * [`freeze`](#freeze)
 * [`unfreeze`](#unfreeze)
+* [`unstake_vote`](#unstake_vote)
 
 Format:
 ```
@@ -579,8 +575,6 @@ Parameter (in Michelson):
   from past stages that is not staked is less than specified `vote_amount` .
 - Fails with `PROPOSAL_NOT_EXIST` if the proposal key is not associated with any ongoing proposals.
 - Fails with `VOTING_STAGE_OVER` if the voting `stage` for the proposal has already ended.
-- Fails with `MAX_VOTERS_REACHED` if the voter count of the associated proposal
-  is already at the max value set by the configuration.
 - Fails with `MISSIGNED` if permit is incorrect with respect to the provided vote parameter and contract state.
 - The entrypoint accepts a list of vote params. As a result, it is possible to
   `vote` on multiple proposals (or the same proposal multiple time) in one entrypoint call.
@@ -610,6 +604,8 @@ Parameter (in Michelson):
   - If proposal got accepted:
     - The return amount for the proposer is equal to or less than the sum of the proposer frozen tokens and the fee paid for the proposal.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
+  - The token return to voters are not immediate. The voters should call `unstake_vote` with the proposal key to get their tokens
+  after `flush` is called.
 - If proposal is accepted, the decision lambda is called.
 - The `quorum_threshold` at the cycle in which the proposal was raised will be
   stored in the proposal, and this threshold will be used to check if the votes
@@ -680,6 +676,26 @@ Parameter (in Michelson):
   baseDAO contract to the address of the sender.
 - Fails with `NOT_ENOUGH_FROZEN_TOKENS` if the author does not have enough tokens
   that can be burned.
+
+
+### **unstake_vote**
+
+```ocaml
+type proposal_key = bytes
+type unstake_vote_param = [proposal_key]
+
+Unstake_vote of unstake_vote_param
+```
+
+Parameter (in Michelson):
+```
+(list %unstake_vote bytes)
+```
+
+- Unstake voter's tokens for proposals that are already flushed or dropped.
+- Fails with `unstake_invalid_proposal` error code if one of the proposals are not yet flushed or dropped.
+- Fails with `voter_does_not_exist` error code if the sender did not vote on the proposal or the sender already called this entrypoint before.
+
 
 ## Custom entrypoints
 
--- a/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Proposal.hs
@@ -114,10 +114,6 @@ test_BaseDAO_Proposal =
   , testGroup "Bounded Value"
       [ testScenario "bounded value on proposals" $ scenario $
           proposalBoundedValue (originateLigoDaoWithConfigDesc dynRecUnsafe)
-
-      , testScenario "bounded value on votes" $ scenario $
-          votesBoundedValue (originateLigoDaoWithConfigDesc dynRecUnsafe)
-
       ]
 
   , testGroup "Freeze-Unfreeze"
@@ -135,6 +131,9 @@ test_BaseDAO_Proposal =
 
       , testScenario "correctly track freeze history" $ scenario $
           checkFreezeHistoryTracking (originateLigoDaoWithConfigDesc dynRecUnsafe)
+
+      , testScenario "tokens are unstaked correctly, only when possible" $ scenario $
+          unstakeVote (originateLigoDaoWithConfigDesc dynRecUnsafe)
       ]
 
  , testGroup "LIGO-specific proposal tests:"
--- a/haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
@@ -19,6 +19,7 @@ module Test.Ligo.BaseDAO.Proposal.Propose
   , validProposal
   , validProposalWithFixedFee
   , unstakesTokensForMultipleVotes
+  , unstakeVote
   ) where
 
 import Universum
@@ -371,6 +372,8 @@ unstakesTokensForMultipleVotes = do
   proposalStart <- getProposalStartLevel dodDao key1
   advanceToLevel (proposalStart + 2*dodPeriod + 1)
   withSender dodAdmin $ call dodDao (Call @"Flush") 100
+  withSender voter $ call dodDao (Call @"Unstake_vote") [key1]
+
   fh_after_flush <- getFreezeHistory dodDao voter
   let expected_after_flush = Just (AddressFreezeHistory
         { fhCurrentUnstaked = 0
@@ -505,6 +508,7 @@ dropProposal originateFn = withFrozenCallStack $ do
   -- 30 tokens are frozen in total, but only 15 tokens are returned after drop_proposal
   checkBalance dodDao dodOwner1 15
 
+
 proposalBoundedValue
   :: (MonadCleveland caps base m, HasCallStack)
   => (ConfigDesc Config -> OriginateFn m) -> m ()
@@ -531,4 +535,98 @@ proposalBoundedValue originateFn = do
     call dodDao (Call @"Propose") params
     call dodDao (Call @"Propose") params
       & expectFailedWith maxProposalsReached
+
+
+unstakeVote
+  :: (MonadCleveland caps base m, HasCallStack)
+  => (ConfigDesc Config -> OriginateFn m)
+  -> m ()
+unstakeVote originateFn = do
+  let flushLevel = 20
+  DaoOriginateData{..}
+    <- originateFn (testConfig
+        >>- (ConfigDesc configConsts{ cmProposalFlushTime = Just flushLevel })
+        >>- (ConfigDesc configConsts{ cmProposalExpiredTime = Just 50 })
+        ) defaultQuorumThreshold
+
+  -- [Voting]
+  withSender dodOwner1 $
+    call dodDao (Call @"Freeze") (#amount :! 30)
+
+  withSender dodOwner2 $
+    call dodDao (Call @"Freeze") (#amount :! 10)
+
+  startLevel <- getOriginationLevel dodDao
+  advanceToLevel (startLevel + dodPeriod)
+
+  -- [Proposing]
+  key1 <- createSampleProposal 1 dodOwner1 dodDao
+  advanceLevel 1
+  key2 <- createSampleProposal 2 dodOwner1 dodDao
+
+  let vote' key = NoPermit VoteParam
+        { vFrom = dodOwner2
+        , vVoteType = True
+        , vVoteAmount = 5
+        , vProposalKey = key
+        }
+
+  -- Advance to a level where vote can be called
+  advanceToLevel (startLevel + 2 * dodPeriod)
+
+  -- [Voting]
+  withSender dodOwner2 . inBatch $ do
+      call dodDao (Call @"Vote") [vote' key1]
+      call dodDao (Call @"Vote") [vote' key2]
+      pure ()
+
+  -- Advance to a level where flush can be called
+  proposalStart <- getProposalStartLevel dodDao key1
+  advanceToLevel (proposalStart + (flushLevel + 1))
+
+  -- [Proposing]
+  withSender dodAdmin $ call dodDao (Call @"Flush") 1
+
+  -- Freeze history before `Unstake_vote`
+  fhOwner2 <- getFreezeHistory dodDao dodOwner2
+  (fhOwner2 <&> fhStaked) @== Just 10
+  (fhOwner2 <&> fhPastUnstaked) @== Just 0
+  (fhOwner2 <&> fhCurrentUnstaked) @== Just 0
+
+  -- Call succeeds.
+  withSender dodOwner2 $ call dodDao (Call @"Unstake_vote") [key1]
+
+  -- Freeze history after `Unstake_vote`
+  fhOwner2_ <- getFreezeHistory dodDao dodOwner2
+  (fhOwner2_ <&> fhStaked) @== Just 5 -- Still staked since `key2` is not yet flush.
+  (fhOwner2_ <&> fhPastUnstaked) @== Just 5
+  (fhOwner2_ <&> fhCurrentUnstaked) @== Just 0
+
+  -- Trigger error since the tokens are already unstaked.
+  withSender dodOwner2 $ call dodDao (Call @"Unstake_vote") [key1]
+    & expectFailedWith voterDoesNotExist
+
+  -- Proposal key2 is not yet flush.
+  withSender dodOwner2 $ call dodDao (Call @"Unstake_vote") [key2]
+    & expectFailedWith unstakeInvalidProposal
+
+  -- owner 1 does not vote on the proposal
+  withSender dodOwner1 $ call dodDao (Call @"Unstake_vote") [key1]
+    & expectFailedWith voterDoesNotExist
+
+  --------------------------------------
+  -- Unstake_vote on dropped proposal
+  --------------------------------------
+  withSender dodOwner1 $ do
+    call dodDao (Call @"Drop_proposal") key2
+
+  advanceToLevel (startLevel + 4 * dodPeriod)
+  -- Call succeeds.
+  withSender dodOwner2 $ call dodDao (Call @"Unstake_vote") [key2]
+
+  -- Freeze history after `Unstake_vote`
+  fhOwner2__ <- getFreezeHistory dodDao dodOwner2
+  (fhOwner2__ <&> fhStaked) @== Just 0
+  (fhOwner2__ <&> fhPastUnstaked) @== Just 10
+  (fhOwner2__ <&> fhCurrentUnstaked) @== Just 0
 
--- a/scripts/generate_error_code.hs
+++ b/scripts/generate_error_code.hs
@@ -65,8 +65,7 @@ invalidInputErrors = errorsEnumerate 100
     :? ("The proposal voting stage has already ended.", NoArg)
   , "max_proposals_reached"
     :? ("The maximum amount of ongoing proposals has been reached.", NoArg)
-  , "max_voters_reached"
-    :? ("The maximum amount of voters has been reached for the proposal.", NoArg)
+  , removedError
   , "forbidden_xtz"
     :? ("Transfer of XTZ is forbidden on this entrypoint.", NoArg)
   , "proposal_not_unique"
@@ -99,6 +98,10 @@ invalidInputErrors = errorsEnumerate 100
     :? ("Executing the proposal's decision lambda results in failure.", NoArg)
   , "entrypoint_not_found"
     :? ("The chosen custom entrypoint does not exist.", NoArg)
+  , "unstake_invalid_proposal"
+    :? ("Cannot call `unstake_vote` on the proposal that is not flushed or dropped.", NoArg)
+  , "voter_does_not_exist"
+    :? ("The sender did not vote on the proposal or already unstaked tokens from the proposal.", NoArg)
   ]
 
 invalidConfigErrors :: [ErrorItem]
--- a/src/base_DAO.mligo
+++ b/src/base_DAO.mligo
@@ -25,6 +25,7 @@ let requiring_no_xtz (param, store, config : forbid_xtz_params * storage * confi
     | Freeze p               -> freeze(p, config, store)
     | Unfreeze p             -> unfreeze(p, config, store)
     | Update_delegate p      -> update_delegates(p, store)
+    | Unstake_vote p         -> unstake_vote(p, config, store)
 
 (*
  * Entrypoints that allow xtz to be sent.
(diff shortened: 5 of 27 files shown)


src/proposal.mligo
```diff
@@ -88,6 +105,44 @@ let update_delegates (params, store : update_delegate_params * storage): return
...
+
+  // Remove the votes
+  let new_voter_proposal_map = Map.update proposal_key (None : (vote_status list) option) voter_proposal_map in
+
+  ( nil_op
+  , { store with
+      voters = Big_map.update Tezos.sender (Some new_voter_proposal_map) store.voters
+    }
+  )
+
```
rinn7e, line 145: @pasqu4le This PR is almost ready for review, only the tests are left to be done.

Before updating the tests, I want your advice on the current implementation. The new entrypoint is call `Unfreeze_unstaked`. Currently it only targets one proposal at a time, my questions are:
1. Do we plan to allow the user to unstake all their votes (on flushed proposals) all at once or not?
2. If the voter that call does not actually vote on that proposal, do we want to fail the call or let it through (by unfreeze nothing)? 
3. Do we keep their votes as history after `Unfreeze_unstaked` is called or removed them?

Hope to get your advice on this.

rinn7e: @pasqu4le I agree with all your points. I have push the final changes as well as updating the tests now, this should be ready for review. The only thing left is updating the docs, which will be what I'm working on next.

The new entrypoint is call `Unfreeze_staked`, let me know if the name is good enough as well.

pasqu4le: @rinn7e what do you think of `unfreeze_vote` as a name instead, given that it can only be used for tokens that were frozen for voting?

sras:
@rinn7e Since we still store voter_map in the proposal structure, doesn't cost of decoding a single proposal still increase linearly with the number of voters who have voted in that proposal?

rinn7e:
> @rinn7e Since we still store voter_map in the proposal structure, doesn't cost of decoding a single proposal still increase linearly with the number of voters who have voted in that proposal?

That's true, I'm thinking of storing the voters in storage as a big_map as well, but decide not to do it. If it is indeed a better choice to move the voter out of the proposal, I'll go with that approach.

src/types.mligo
```diff
@@ -166,6 +160,7 @@ type storage =
   ; extra : contract_extra
   ; proposals : (proposal_key, proposal) big_map
   ; proposal_key_list_sort_by_level : (blocks * proposal_key) set
+  ; voters : (address * proposal_key, vote_status) big_map
```
rinn7e, line 163: @sras I've made the changes according to your feedback, this should be good now.

src/types.mligo
```diff
@@ -223,6 +234,7 @@ type forbid_xtz_params =
   | Freeze of freeze_param
   | Unfreeze of unfreeze_param
   | Update_delegate of update_delegate_params
+  | Unfreeze_vote of unfreeze_vote_param
```
sras, line 237: Why is this called `Unfreeze` when what it is doing seems to be just unstaking staked frozen tokens back to unstaked frozen tokens. Is any 'Unfreez' ing happening here?

src/types.mligo
```diff
@@ -51,8 +51,16 @@ type nonce = nat
...
 
-// Stores all voter data for a proposal
-type voter_map = (address * vote_type, nat) map
+type vote_item =
+  { vote_amount : nat
+  ; vote_type : vote_type
+  }
+
+type vote_status =
+  { votes : vote_item list
```
sras, line 60: I see that we have replced `voter_map` with `vote_item list`, I suppose this is done to track the total number of staked tokens. If that is the case can't we just add a `nat` field to `vote_status` to keep track of staked tokens, and increment it during each `vote` operation for this user, and keep the `votes` field as `voter_map` itself?

src/types.mligo
```diff
@@ -94,10 +102,11 @@ type proposal =
   // ^ address of the proposer
   ; proposer_frozen_token : nat
   // ^ amount of frozen tokens used by the proposer, exluding the fixed fee
-  ; voters : voter_map
-  // ^ voter data
   ; quorum_threshold: quorum_threshold
   // ^ quorum threshold at the cycle in which proposal was raised
+  ; is_proposer_unfreezed: bool
```
sras, line 107: If this indicate that the proposal is either flushed or dropped, then I think it should be named accordingly.

src/types.mligo
rinn7e: That's true, the term `unfreeze` on `flush` was used a lot previously, but the correct term should be `unstake` on `flush`, since we have a separate `unfreeze` entrypoint now. 

I'll rename this to `unstake` instead.

src/types.mligo
rinn7e: `type voter_map = (address * vote_type, nat) map` is replaced by `(address * proposal_key, vote_status) big_map`, and is moved outside of the `proposal` to `storage`.

`vote_status` type is as followed:
```
type vote_status =
  { votes : vote_item list
  ; is_unfreezed : bool
  }

type vote_item =
  { vote_amount : nat
  ; vote_type : vote_type
  }

 ```
 
 I'm not sure exactly what your suggestion is, but I guess you mean that it is better to just use `nat` instead of a `list` there.
 
 ```
 type vote_status =
  { upvotes : nat
  ; downvotes: nat
  ; is_unfreezed : bool
  }
```

I could do it this way as well, but I think storing each upvote and downvote as a list could be useful in the future but I may be wrong.

src/types.mligo
rinn7e: I've pushed the changes, have to squash the commits due to fixing conflicts when rebasing.

src/types.mligo
sras:
```
type vote_status =
 { upvotes : nat
 ; downvotes: nat
 ; is_unfreezed : bool
 }
```
Ah, yes, that would work better.

>I think storing each upvote and downvote as a list could be useful in the future but I may be wrong.

But we didn't have that capability as of now, right?

rinn7e: > But we didn't have that capability as of now, right?
The current implementation already stores upvote and downvote as a list already. You mean we don't have the usage of it now right? 

@pasqu4le Since you suggest keeping the vote history, what do you think about @sras feedback? With the new feedback, we still keep track of the vote history but we won't know how many times the voter upvote/downvote a proposal (which may or may not be useful).

sras: >The current implementation already stores upvote and downvote as a list already.

@rinn7e Oh, by current implementation, I meant the version in master branch. AFAIU I that version, we don't keep the votes in a list.

rinn7e: I mean the implementation in this PR already store them in a list.

This is due to the suggestion 3 in here https://github.com/tezos-commons/baseDAO/pull/313#discussion_r767946169

docs/specification.md
```diff
@@ -681,6 +677,27 @@ Parameter (in Michelson):
...
+
+Unstake_vote of unstake_vote_param
+```
+
+Parameter (in Michelson):
+```
+(list %unstake_vote bytes)
+```
+
+- Unfreeze voter's tokens for proposals that are already flushed.
```
sras, line 695: >Unfreeze voter's tokens

Probably we should also say "Unstake" here as well..

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
```diff
@@ -532,3 +536,66 @@ proposalBoundedValue originateFn = do
...
+  withSender dodOwner2 $ call dodDao (Call @"Unstake_vote") [key1]
+    & expectFailedWith voterAlreadyUnstake
+
+  -- Proposal key2 is not yet flush.
+  withSender dodOwner2 $ call dodDao (Call @"Unstake_vote") [key2]
+    & expectFailedWith proposalIsNotFlushed
+
+  -- owner 1 does not vote on the proposal
+  withSender dodOwner1 $ call dodDao (Call @"Unstake_vote") [key1]
+    & expectFailedWith voterDoesNotExist
```
sras, line 601: AFAIU this test does not actually test if the "Unstaking" worked. Right?

src/types.mligo
sras: Oh. I didn't see that comment. Hope he can provide some context.

docs/specification.md
rinn7e: Nice catch, updated this.

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
rinn7e: I added `freeze_history` comparison now, this should test if unstaking works properly.

sras: Can we also add a test to check that the unstaking works when the proposal is dropped?

rinn7e: Make sense, I've added a case for when the proposal is dropped as well.

src/proposal.mligo
```diff
@@ -231,14 +272,16 @@ let unfreeze_proposer_and_voter_token
...
+  // Do the unstake for the proposer
   let store = unstake_tk(tokens, burn_amount, proposal.proposer, period, store) in
 
-  // unfreeze_voter_token
-  let do_unfreeze = fun
-        ( store, voter_with_vote : storage * ((address * vote_type) * nat))
-          -> unstake_tk(voter_with_vote.1, 0n, voter_with_vote.0.0, period, store) in
+  // Update `is_flushed_or_dropped`
+  let new_proposals = Map.add proposal_key
+        ({proposal with is_flushed_or_dropped = true}) store.proposals in
```
pasqu4le, line 281: Isn't it odd that this happens in a function called `unstake_proposer_token`?

I mean, by the name they don't seem related

src/types.mligo
pasqu4le: > Hope he can provide some context

I mentioned this elsewhere as well, but this was a requested feature: #151

sras: @pasqu4le the issue is that version in master branch stores the information in  a map, which in this PR it is changed to a list. So basically the difference is that earlier we indeed stored who voted up or down, but didn't store how many times they voted. That is if they voted two times with 5 upvotes, then we only stored that they voted with 10 upvotes, but not that it was 2 times, with 5 votes each. But this PR adds this additional information, because the votes are stored in a list, so each instance of a vote is stored, which is more information than we need.

This is the change that I am pointing out as being new.

rinn7e: Regarding this point, I'm good with both combining the votes (like the previous implementation) or keeping it as a list (this PR implementation). I use the list simply because it could be useful, but if it isn't and cost too much, I'll go with combining them instead.

src/types.mligo
```diff
@@ -94,10 +102,11 @@ type proposal =
   // ^ address of the proposer
   ; proposer_frozen_token : nat
   // ^ amount of frozen tokens used by the proposer, exluding the fixed fee
-  ; voters : voter_map
-  // ^ voter data
   ; quorum_threshold: quorum_threshold
   // ^ quorum threshold at the cycle in which proposal was raised
+  ; is_flushed_or_dropped: bool
```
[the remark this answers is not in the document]
rinn7e: At first, I didn't think there is a way to calculate it, but now that you mention it, I think it is possible by checking `proposal_key_list_sort_by_level`. I'll make the change on this.

src/base_DAO.mligo
```diff
@@ -25,6 +25,7 @@ let requiring_no_xtz (param, store, config : forbid_xtz_params * storage * confi
     | Freeze p               -> freeze(p, config, store)
     | Unfreeze p             -> unfreeze(p, config, store)
     | Update_delegate p      -> update_delegates(p, store)
+    | Unstake_vote p        -> unstake_vote(p, config, store)
```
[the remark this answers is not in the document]
rinn7e: Fixed.

src/types.mligo
```diff
@@ -94,10 +102,11 @@ type proposal =
...
+  // ^ If true, it indicates that the proposal is either flushed or dropped.
+  // Allow voters to call `Unstake_vote` on this proposal.
```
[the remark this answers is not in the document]
rinn7e: Updated the config section.

scripts/generate_error_code.hs
```diff
@@ -99,6 +98,12 @@ invalidInputErrors = errorsEnumerate 100
     :? ("Executing the proposal's decision lambda results in failure.", NoArg)
   , "entrypoint_not_found"
     :? ("The chosen custom entrypoint does not exist.", NoArg)
+  , "proposal_is_not_flushed"
+    :? ("The voter is trying to call `Unstake_vote` on proposals which are not flushed yet.", NoArg)
+  , "voter_does_not_exist"
+    :? ("The voter calling `Unstake_vote` does not vote on the proposal.", NoArg)
```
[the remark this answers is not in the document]
rinn7e: Updated.

scripts/generate_error_code.hs
```diff
@@ -99,6 +98,12 @@ invalidInputErrors = errorsEnumerate 100
...
+  , "voter_already_unstake"
+    :? ("The voter is trying to call `Unstake_vote` on proposals which his/her tokens have already been unstaked.", NoArg)
```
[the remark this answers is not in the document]
rinn7e: Updated.

scripts/generate_error_code.hs
```diff
@@ -99,6 +98,12 @@ invalidInputErrors = errorsEnumerate 100
     :? ("Executing the proposal's decision lambda results in failure.", NoArg)
   , "entrypoint_not_found"
     :? ("The chosen custom entrypoint does not exist.", NoArg)
+  , "proposal_is_not_flushed"
+    :? ("The voter is trying to call `Unstake_vote` on proposals which are not flushed yet.", NoArg)
```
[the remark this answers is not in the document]
rinn7e: I've renamed this to `unstake_invalid_proposal` to make it clearer. Let me know if this is good enough.

docs/specification.md
```diff
@@ -94,9 +94,6 @@ type config =
```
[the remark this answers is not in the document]
rinn7e: Updating the voting section part as well as the flush description. I'm not sure where else to update.

src/types.mligo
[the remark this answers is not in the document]
rinn7e: I've removed `is_flushed_or_dropped` now and calculate it by checking `proposal_key_list_sort_by_level` instead.

src/proposal.mligo
rinn7e: Since I've removed `is_flushed_or_dropped`, I also this part as well.

src/types.mligo
pasqu4le: Ah I see, I didn't understand what was the difference, sorry.

AFAIK we only _need_ to keep the total amount of upvotes/downvotes and not the historic list of votes submitted.

I agree that we shouldn't keep this as we have no use for it now, so it looks like indeed:
```
type vote_status =
 { upvotes : nat
 ; downvotes: nat
 ; is_unfreezed : bool
 }
```
would be more efficient and work out just fine for this.

docs/specification.md
```diff
@@ -681,6 +673,27 @@ Parameter (in Michelson):
...
+
+Unstake_vote of unstake_vote_param
+```
+
+Parameter (in Michelson):
+```
+(list %unstake_vote bytes)
+```
+
+- Unstake voter's tokens for proposals that are already flushed.
```
pasqu4le, line 691:
```suggestion
- Unstake voter's tokens for proposals that are already flushed or dropped.
```

docs/specification.md
```diff
@@ -681,6 +673,27 @@ Parameter (in Michelson):
...
+- Fails with `unstake_invalid_proposal` error code if one of the proposals are not yet flushed or dropped.
+- Fails with `voter_does_not_exist` error code if the sender does not vote on the proposal.
```
pasqu4le, line 693:
```suggestion
- Fails with `voter_does_not_exist` error code if the sender did not vote on the proposal.
```

docs/specification.md
```diff
@@ -681,6 +673,27 @@ Parameter (in Michelson):
...
+- Fails with `voter_unstaked` error code if the sender already called this entrypoint before.
```
pasqu4le, line 694:
```suggestion
- Fails with `voter_unstaked` error code if the sender already called this entrypoint before for one of the given proposals.
```

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
```diff
@@ -135,6 +131,9 @@ test_BaseDAO_Proposal =
 
       , testScenario "correctly track freeze history" $ scenario $
           checkFreezeHistoryTracking (originateLigoDaoWithConfigDesc dynRecUnsafe)
+
+      , testScenario "can unstake tokens after flush is called" $ scenario $
```
pasqu4le, line 135: nitpick, but I suggest to change this to something like:
```suggestion
      , testScenario "tokens are unstaked correctly, only when possible" $ scenario $
```
simply because the scenario tests for incorrect cases and dropped proposals as well

src/proposal.mligo
```diff
@@ -29,6 +29,13 @@ let check_if_proposal_exist (proposal_key, store : proposal_key * storage): prop
     then p
     else (failwith proposal_not_exist : proposal)
 
+[@inline]
+let check_if_proposal_flushed_or_dropped (proposal_key, store : proposal_key * storage): proposal =
+  let p = fetch_proposal (proposal_key, store) in
+  if Set.mem (p.start_level, proposal_key) store.proposal_key_list_sort_by_level
+    then (failwith unstake_invalid_proposal : proposal)
+    else p
```
pasqu4le, line 37: I would suggest to either:
- rename this function to `proposal_is_unstakeable`
- move its definition inside `unstake_vote_one` itself

for the reasons that this is only used by that function and that it throws `unstake_invalid_proposal` which is specific to this use.

src/proposal.mligo
```diff
@@ -88,6 +112,42 @@ let update_delegates (params, store : update_delegate_params * storage): return
   , { store with delegates = List.fold update_delegate params store.delegates }
   )
 
+
+// Unstake voter's tokens on a proposal that is already been flushed. Fail if the voter does not vote
+// on that proposal, or voter has already unfreezed, or the proposal is not flushed.
```
pasqu4le, line 117:
```suggestion
// Unstake voter's tokens on a proposal that has already been flushed or dropped.
// Fail if the voter did not vote on that proposal, or voter has already unfreezed, or
// the proposal was not yet flushed nor dropped.
```

docs/specification.md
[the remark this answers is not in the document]
pasqu4le: @rinn7e I checked the spec and I don't see any other change necessary :ok_hand:

docs/specification.md

docs/specification.md

docs/specification.md

haskell/test/Test/Ligo/BaseDAO/Proposal.hs
rinn7e: Make sense, I've updated this.

src/proposal.mligo
rinn7e: I've moved this to `unstake_vote_one` itself now.

src/proposal.mligo
rinn7e: Updated.

src/types.mligo
rinn7e: I've updated this. I think this is ready for re-review.

src/types.mligo
pasqu4le: Minor point, but wouldn't this field be better called `votes`, since it track all kinds of vote info (not just who voted)?

src/proposal.mligo
```diff
@@ -58,6 +58,23 @@ let ensure_proposal_is_unique (propose_params, store : propose_params * storage)
...
+        let new_freeze_history = Big_map.update addr (Some(fh)) store.freeze_history in
+        let new_total_supply =
+          match Michelson.is_nat (store.frozen_total_supply - burn_amount) with
+            Some new_total_supply -> new_total_supply
+          | None -> (failwith("BAD_STATE") : nat) in
+        { store with
+            freeze_history = new_freeze_history
+          ; frozen_total_supply = new_total_supply
+        }
+    | None -> (failwith("BAD_STATE") : storage)
```
pasqu4le, line 76: Shouldn't this be using the error code?
```suggestion
          | None -> (failwith bad_state : nat) in
        { store with
            freeze_history = new_freeze_history
          ; frozen_total_supply = new_total_supply
        }
    | None -> (failwith bad_state : storage)
```

src/proposal.mligo
```diff
@@ -88,6 +105,43 @@ let update_delegates (params, store : update_delegate_params * storage): return
...
+            then (failwith unstake_invalid_proposal : unit)
+            else unit in
+
+  // Check if voter exist.
+  let vote_status = match Big_map.find_opt (Tezos.sender, proposal_key) store.voters with
+      | Some v -> v
+      | None -> (failwith voter_does_not_exist : vote_status) in
+
+  // Ensure that the voter has not called `unstake_vote` before.
+  let _ =  if vote_status.is_unstaked = true
```
pasqu4le, line 126: Not sure if ligo optimizes it away, but this could be:
```suggestion
  let _ =  if vote_status.is_unstaked
```

pasqu4le (approved):
Minor comments but otherwise LGTM

Don't forget to clean up the commit history tho!

src/proposal.mligo
```diff
@@ -88,6 +105,43 @@ let update_delegates (params, store : update_delegate_params * storage): return
...
+      then (failwith voter_unstaked : unit)
+      else unit in
+
+  // Do the unstake
+  let tokens = vote_status.upvotes + vote_status.downvotes in
+  let store = unstake_tk(tokens, 0n, Tezos.sender, config.period, store) in
+
+  // Mark `is_unstaked` on voter's vote.
+  { store with
+      voters = Big_map.add (Tezos.sender, proposal_key) ({vote_status with is_unstaked = true}) store.voters
```
sras, line 136: I wonder if there is still a reason to maintain the voter in voters list after the voter has unstaked their token, @pasqu4le ?

docs/specification.md
```diff
@@ -304,10 +299,7 @@ Moreover the proposal to vote on must have been submitted in the proposing `stag
 immediately preceding and the voter must have frozen his tokens in one of the
 preceding `stage`s.
 
-Each vote stakes one frozen token. Staked tokens cannot be unfreezed till they
-are unstaked when the associated proposal is flushed. The number of staked tokens
-only depend on the number of votes, and does not depend on whether the vote is in
-favor or against a proposal.
+Each vote stakes one frozen token. Staked tokens cannot be unfreezed till the associated proposal is flushed and the tokens are unstaked by calling `unstake_vote`. The number of staked tokens only depend on the number of votes, and does not depend on whether the vote is in favor or against a proposal.
```
sras, line 302: > associated proposal is flushed..

or dropped, right?

docs/specification.md
```diff
@@ -610,6 +601,7 @@ Parameter (in Michelson):
   - If proposal got accepted:
     - The return amount for the proposer is equal to or less than the sum of the proposer frozen tokens and the fee paid for the proposal.
     - The return amount for each voters is equal to or less than the voter's frozen tokens.
+  - The token return to voters are not immediate. The voters should call `unstake_vote` to get their tokens after `flush` is called.
```
sras, line 604: We should also probably make it clear that voters should hold on to the proposal keys that they have voted on, if they want their staked tokens back.

src/proposal.mligo
pasqu4le: good point, I suppose that we could remove the entries here (and the `is_unstaked` field entirely), but I [asked here](https://serokell.slack.com/archives/C02KY15P5GE/p1642015165003300?thread_ts=1637159479.009800&cid=C02KY15P5GE) to be sure.

@rinn7e if that's the case I guess my [previous comment](https://github.com/tezos-commons/baseDAO/pull/313#discussion_r783180286) would be different and I would instead consider calling this `storage` field `staked_votes` instead

docs/specification.md
pasqu4le: They could also use an indexer/explorer for this tho.

For example, it would be sufficient to look at one's previous call to the `vote` entrypoint or traversing thie keys of the votes big_map.

In general I would avoid making suggestions in the specification when possible, given that this is supposed to be a document strictly describing what the contract does and how it behaves.

sras: Ok, I think we can at least say "should call `unstake_vote` with the proposal keys"

src/types.mligo
rinn7e: I've updated this to `staked_votes` now.

src/proposal.mligo
rinn7e: Updated.

src/proposal.mligo
rinn7e: Seems like the response is fine with not keeping track of the voters. I've removed `is_unstaked` and rename to `staked_votes`. This should be good now.

Since we no longer know if the voter has called the entrypoint previously, I removed the `voter_unstaked` and mention its description as part of `voter_does_not exist` instead.

docs/specification.md

docs/specification.md
rinn7e: I've added `with the proposal keys` part.

src/types.mligo
```diff
@@ -51,10 +51,9 @@ type nonce = nat
 // Represents whether a voter has voted against (false) or for (true) a given proposal.
 type vote_type = bool
 
-type vote_status =
+type staked_vote =
  { upvotes : nat
  ; downvotes: nat
```
sras, line 56: Do we need the separate `upvotes`, `downvotes` fields now?

rinn7e: I guess we don't need to separate them anymore, but it may still be useful for the test (Easier to do the check).

I combine them anyway, you can check it now.

sras: LGTM now. But maybe @pasqu4le also better signoff this change as well.

sras (approved):

src/types.mligo
pasqu4le: Yep, makes sense to me too.

This also was only used to keep historical data and is better off removed now :ok_hand:

src/proposal.mligo
```diff
@@ -88,6 +105,37 @@ let update_delegates (params, store : update_delegate_params * storage): return
...
+
+  // Check if voter exist.
+  let staked_vote_amount = match Big_map.find_opt (Tezos.sender, proposal_key) store.staked_votes with
+      | Some v -> v
+      | None -> (failwith voter_does_not_exist : staked_vote) in
+
+  // Do the unstake
+  let store = unstake_tk(staked_vote_amount, 0n, Tezos.sender, config.period, store) in
+
+  // Mark `is_unstaked` on voter's vote.
```
pasqu4le, line 128: nitpick:
```suggestion
  // Remove voter's vote from staked amounts
```

pasqu4le (approved):

PR #315, opened 2022-01-06
[#314] Improve `storage.start_level`

### LINKED ISSUE #314: Make network tests more resilient to origination time

### Background

Our network tests have a slight complication over most others: the scenarios are written with the expectation that, after origination, they'll be at the beginning of the first period.

This wouldn't normally be the case however, because origination time take a while, so if we were to use "now" as the start of the first period we would have issues (especially in local-chain, where blocks appear more frequently).

### Current solution

As a workaround to this problem we ended up using a delay in the "starting level" used at origination, so that when the next scenario "action" was executed it would be at the beginning of the first period.

This was done, in practice, by [adding an offset in the stored starting level for the delay](https://github.com/tezos-commons/baseDAO/blob/bbf0745cd80f07ea58b1bdf41b7311db62219b8a/haskell/test/Test/Ligo/BaseDAO/Common.hs#L174) and then [waiting to reach it](https://github.com/tezos-commons/baseDAO/blob/bbf0745cd80f07ea58b1bdf41b7311db62219b8a/haskell/test/Test/Ligo/BaseDAO/Common.hs#L194).

### Improvements needed

The current solution is somewhat brittle and more of a hack. The main problem being that the delay is somewhat arbitrary, because it needs to:
- be as short as possible, to avoid the unnecessary waiting times
- be as long, or longer, than the worst case scenario (aka the slowest origination).

These make the delay:
- hard to establish
- tuned to a specific protocol or tezos version (as we noticed with hangzhou's activation)
- never entirely safe, because origination can some times take longer due to external factors

So, we need to have a better system in place for this.

_note: some hints may be found in [this slack thread](https://serokell.slack.com/archives/C019PCGKTK2/p1639752186050600?thread_ts=1639659498.043400&cid=C019PCGKTK2)_

# Acceptance criteria

Network scenarios do not depend on contract origination time.

rinn7e:
## Description

Problem: The `storage.start_level` is set at the start of origination
instead of when the origination is completed. Because of this, in some
case when the origination takes longer than expected, the
`storage.start_level` is out of sync.

Solution: Provide a way to set `storage.start_level` after the
origination is completed (by calling `Transfer_ownership`). This won't
take effect if the `start_level` is already set previously.
Update the tests, and remove origination workaround.

## Related issue(s)

Resolves #314

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ $(OUT)/trivialDAO_storage.tz: src/**
             { address = (\"$(call require_defined,governance_token_address)\" : address) \
             ; token_id = ($(call require_defined,governance_token_id) : nat) \
             } \
-          ; start_level = {blocks = $(call require_defined,start_level)} \
+          ; start_level = Some {blocks = $(call require_defined,start_level)} \
           ; metadata_map = ($(call escape_double_quote,$(metadata_map)) : metadata_map) \
           ; freeze_history = ($(call escape_double_quote,$(freeze_history)) : freeze_history_list) \
           } \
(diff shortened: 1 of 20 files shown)


rinn7e:
@pasqu4le This PR is ready for review. I go with the approach that we have discussed, let me know if you think another approach is better.

pasqu4le:
I'm just commenting to say that this is put aside for the moment.

It's still not clear if this is the best solution and this does change the behavior quite significantly.

pasqu4le:
Closing this because ([aside from the above](https://github.com/tezos-commons/baseDAO/pull/315#issuecomment-1108462187)) in practice the origination issues were dealt when they appeared and timing issues in general [have recently improved a lot](https://github.com/tezos-commons/baseDAO/pull/344).

PR #317, opened 2022-01-20
[#250] fix documentation around rejected proposal slash value

pasqu4le:
## Description

Problem: the documentation wasn't fully updated to reflect the
behavior of the rejected proposal slash value lambda, the last
time it was changed.
As a result the docs contradict each other in some cases and
don't match the actual behavior.

Solution: fix the various issues in the docs related to this
lambda and its usage.

## Related issue(s)

Fixup for #250 (and probably older issues as well)

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e (approved):
Looks good for me 👍

PR #318, opened 2022-01-25
[#316] Remove proposal limit and use a linked list for sorting

### LINKED ISSUE #316: Remove proposal limit and use a linked list for sorting

There is currently a limit on the number of concurrent ongoing proposals, due to the fact that proposals need to be sorted (for `flush`) and that this is done using a `set`, which cannot grow indefinitely without causing gas cost problems.

This issue is meant to try to remove this limit by changing how ongoing proposals are tracked, and it's similar in intention to what #311 did for voters.

One idea is to remove `proposal_key_list_sort_by_level` and replace it with:
- two fields to track the `first` and `last` ongoing proposals
- a `big_map` where the key is a proposal's ID and the value is a tuple with the proposal ID that comes `next` and `prev`iously

in this way we would obtain and keep a doubly-linked-list of ongoing proposals that could be accessed and updated lazily.

With this different implementation it should be possible to remove the proposal limit while keeping the same interface and behavior of all the entrypoints.

Some details might change during implementation and the cost of this change will need to be checked as well, but it should be worthwhile.

Note: if accepted, this would replace #68 and make #172 unnecessary. 

# Acceptance criteria

`proposal_key_list_sort_by_level` is replaced by a lazy doubly-linked-list and the proposal limit is removed.

rinn7e:
## Description

Problem: There is currently a limit on the number of concurrent ongoing proposals, due to the fact that proposals need to be sorted (for `flush`) and that this is done using a `set`, which cannot grow indefinitely without causing gas cost problems.

Solution: Remove `proposal_key_list_sort_by_level` and use a doubly linked list instead to store the proposals in order.

## Related issue(s)

Resolves #316

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -92,8 +92,6 @@ type config =
   // It has access to the proposal, can modify `contractExtra` and perform arbitrary
   // operations.
 
-  ; max_proposals : nat
-  // ^ Determine the maximum number of ongoing proposals that are allowed in the contract.
   ; max_quorum_threshold : quorum_fraction
   // ^ Determine the maximum value of quorum threshold that is allowed.
   ; min_quorum_threshold : quorum_fraction
@@ -523,7 +521,6 @@ Parameter (in Michelson):
 - Fails with `NOT_PROPOSING_STAGE` if the current stage is not a proposing one.
 - Fails with `FAIL_PROPOSAL_CHECK` if the proposal is rejected by `proposal_check`
   from the configuration.
-- Fails with `MAX_PROPOSALS_REACHED` if the current amount of ongoing proposals is at max value set by the config.
 - Fails with `PROPOSAL_NOT_UNIQUE` if exactly the same proposal from the same author has been proposed.
 
 
--- a/haskell/test/SMT/Model/BaseDAO/Proposal.hs
+++ b/haskell/test/SMT/Model/BaseDAO/Proposal.hs
@@ -26,6 +26,7 @@ import Morley.Util.Named
 
 import Ligo.BaseDAO.Types
 import SMT.Model.BaseDAO.Permit
+import SMT.Model.BaseDAO.Proposal.Plist
 import SMT.Model.BaseDAO.Proposal.FreezeHistory
 import SMT.Model.BaseDAO.Proposal.QuorumThreshold
 import SMT.Model.BaseDAO.Token
@@ -39,7 +40,7 @@ checkIfProposalExist key = do
 
   let resultMaybe = do
         p <- Map.lookup key (store & sProposals & bmMap)
-        if Set.member (p & plStartLevel, key) (store & sProposalKeyListSortByDate) then
+        if plistMem key (store & sOngoingProposalsDlist) then
           Just p
         else Nothing
 
@@ -56,13 +57,6 @@ checkDelegate from author  = do
   if (author /= from) && not (Map.member key (store & sDelegates & bmMap)) then
     throwError NOT_DELEGATE
   else pure from
-
-checkProposalLimitReached :: ModelT ()
-checkProposalLimitReached = do
-  proposalKeyListSortByDate <- getStore <&> sProposalKeyListSortByDate
-  maxProposal <- getConfig <&> cMaxProposals
-  when (maxProposal <= (fromIntegral $ length proposalKeyListSortByDate)) $
-    throwError MAX_PROPOSALS_REACHED
 
 stakeTk :: Natural -> Address -> ModelT ()
 stakeTk tokenAmount addr = do
@@ -108,7 +102,7 @@ addProposal params = do
         }
   modifyStore $ \s -> pure $ s
     { sProposals = BigMap Nothing $ Map.insert proposalKey proposal (bmMap (s & sProposals))
-    , sProposalKeyListSortByDate = Set.insert (lvl, proposalKey) (s & sProposalKeyListSortByDate)
+    , sOngoingProposalsDlist = plistInsert proposalKey (s & sOngoingProposalsDlist)
     }
 
 unstakeTk :: Natural -> Natural -> Address -> ModelT ()
@@ -153,10 +147,6 @@ unstakeProposerToken isAccepted proposal = do
   unstakeTk tokens burnAmount (proposal & plProposer)
 
 
-removeFromProposalSortByLevel :: Natural -> ProposalKey -> ModelT ()
-removeFromProposalSortByLevel pLevel proposalKey = modifyStore $ \s ->
-  pure $ s { sProposalKeyListSortByDate = Set.delete (pLevel, proposalKey) (s & sProposalKeyListSortByDate)}
-
 applyPropose :: ModelSource -> ProposeParams -> ModelT ()
 applyPropose mso param@ProposeParams{..} = do
 
@@ -167,7 +157,6 @@ applyPropose mso param@ProposeParams{..} = do
 
   store <- getStore
   get <&> msProposalCheck >>= \f -> f (param, (store & sExtra))
-  checkProposalLimitReached
 
   let amountToFreeze = ppFrozenToken + fixedFee
   currentStage <- getCurrentStageNum
@@ -276,48 +265,56 @@ isLevelReached proposal target = do
   lvl <- get <&> msLevel
   pure (lvl >= (proposal & plStartLevel) + target)
 
-handleProposalIsOver :: (Natural, ProposalKey) -> ModelT Bool
-handleProposalIsOver (proposalLvl, proposalKey) = do
-  proposal <- checkIfProposalExist proposalKey
 
-  isExpired <- isLevelReached proposal =<< (getConfig <&> cProposalExpiredLevel)
-  canBeFlushed <- isLevelReached proposal =<< (getConfig <&> cProposalFlushLevel)
-  if isExpired then
-    throwError EXPIRED_PROPOSAL
-  else if canBeFlushed then do
-    cond <- getStore <&> \s ->
-              doTotalVoteMeetQuorumThreshold proposal s && ((proposal & plUpvotes) > (proposal & plDownvotes))
+handleProposalIsOver
+  :: Proposal
+  -> ModelT ()
+handleProposalIsOver proposal = do
+  cond <- getStore <&> \s ->
+            doTotalVoteMeetQuorumThreshold proposal s && ((proposal & plUpvotes) > (proposal & plDownvotes))
+  unstakeProposerToken cond proposal
+  -- Execute decision lambda if the proposal is passed.
+  when cond $ do
+    store <- getStore
+    (ops, newExtra, guardianMaybe) <-
+      get <&> msDecisionLambda >>= \f -> f (DecisionLambdaInput proposal (store & sExtra))
 
-    unstakeProposerToken cond proposal
-    when cond $ do
-      store <- getStore
-      (ops, newExtra, guardianMaybe) <-
-        get <&> msDecisionLambda >>= \f -> f (DecisionLambdaInput proposal (store & sExtra))
+    case guardianMaybe of
+      Just g -> modifyStore $ \s -> pure $ s { sGuardian = g }
+      Nothing -> pure ()
 
-      case guardianMaybe of
-        Just g -> modifyStore $ \s -> pure $ s { sGuardian = g }
-        Nothing -> pure ()
+    modifyStore $ \s -> pure $ s { sExtra = newExtra }
+    mapM_ execOperation ops
 
-      modifyStore $ \s -> pure $ s { sExtra = newExtra }
-      mapM_ execOperation ops
 
-    removeFromProposalSortByLevel proposalLvl proposalKey
-    pure True
-  else
-    pure False
+flushEach :: Integer -> ModelT Integer
+flushEach n = do
+  store <- getStore
+  let (plistHead, plistNew) = plistPop (store & sOngoingProposalsDlist)
+  case plistHead of
+    Just firstKey -> do
+      proposal <- checkIfProposalExist firstKey
+      isExpired <- isLevelReached proposal =<< (getConfig <&> cProposalExpiredLevel)
+      canBeFlushed <- isLevelReached proposal =<< (getConfig <&> cProposalFlushLevel)
+      when isExpired $
+        throwError EXPIRED_PROPOSAL
+
+      if (n > 0 && canBeFlushed) then do
+        handleProposalIsOver proposal
+        modifyStore $ \s -> pure $ s { sOngoingProposalsDlist = plistNew}
+        flushEach (n - 1)
+      else
+        pure n
+    Nothing -> pure n
+
 
 applyFlush :: ModelSource -> Natural -> ModelT ()
 applyFlush _ param = do
-  when (param == 0) $
-    throwError EMPTY_FLUSH
 
-  anyFlushedProposal <-
-    getStore
-      <&> (take (fromIntegral param) . Set.toAscList . sProposalKeyListSortByDate)
-      >>= mapM handleProposalIsOver
-      <&> or
+  newN <- flushEach (toInteger param)
 
-  unless anyFlushedProposal $ throwError EMPTY_FLUSH
+  if (newN == toInteger param) then throwError EMPTY_FLUSH
+  else pure ()
 
 applyDropProposal :: ModelSource -> ProposalKey -> ModelT ()
 applyDropProposal mso proposalKey = do
@@ -331,7 +328,10 @@ applyDropProposal mso proposalKey = do
     || proposalIsExpired
   then do
     unstakeProposerToken False proposal
-    removeFromProposalSortByLevel (proposal & plStartLevel) proposalKey
+    modifyStore $ \s ->
+      pure $ s
+        { sOngoingProposalsDlist = plistDelete proposalKey (s & sOngoingProposalsDlist)
+        }
   else
     throwError DROP_PROPOSAL_CONDITION_NOT_MET
 
@@ -364,12 +364,8 @@ applyUnstakeVoteOne mso key = do
   store <- getStore
 
   -- Ensure proposal is already flushed or dropped.
-  case Map.lookup key (store & sProposals & bmMap) of
-    Just p ->
-      if Set.member (p & plStartLevel, key) (store & sProposalKeyListSortByDate) then
-        throwError UNSTAKE_INVALID_PROPOSAL
-      else pure ()
-    Nothing -> throwError PROPOSAL_NOT_EXIST
+  when (plistMem key (store & sOngoingProposalsDlist)) $
+    throwError UNSTAKE_INVALID_PROPOSAL
 
   tokens <- case Map.lookup (mso & msoSender, key) (store & sStakedVotes & bmMap) of
           Just v -> pure v
@@ -380,8 +376,6 @@ applyUnstakeVoteOne mso key = do
   modifyStore $ \s -> pure $ s
     { sStakedVotes = BigMap Nothing $ Map.delete (mso & msoSender, key) (s & sStakedVotes & bmMap)
     }
-
-  pure ()
 
 applyUnstakeVote :: ModelSource -> UnstakeVoteParam -> ModelT ()
 applyUnstakeVote mso params =
--- a/haskell/test/Test/Ligo/BaseDAO/Plist.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Plist.hs
@@ -1,0 +1,90 @@
+-- SPDX-FileCopyrightText: 2021 TQ Tezos
+-- SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+module Test.Ligo.BaseDAO.Plist
+  ( test_Plist
+  ) where
+
+import Universum hiding (drop, swap, toList)
+
+import Data.List ((!!))
+import qualified Data.List as DL
+import Hedgehog
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+import Test.Tasty (TestTree)
+import Test.Tasty.Hedgehog (testProperty)
+
+import Hedgehog.Gen.Tezos.Crypto (genSecretKey)
+import Lorentz hiding (cast, concat, get, not)
+import Morley.Tezos.Address (mkKeyAddress)
+import Morley.Tezos.Crypto (toPublic)
+
+import Ligo.BaseDAO.Types
+import SMT.Model.BaseDAO.Proposal.Plist
+import Test.Ligo.BaseDAO.Common (makeProposalKey, metadataSize)
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+genProposalKeyList :: MonadGen m => Int -> m [ProposalKey]
+genProposalKeyList atLeast = do
+  i <- Gen.integral (Range.constant atLeast 10)
+  addrs <- genSecretKey
+    <&> (\secret -> mkKeyAddress . toPublic $ secret)
+    & Gen.list (Range.linear atLeast i)
+
+  pure $ (\addr ->
+      let proposalMeta = lPackValueRaw @Natural 1
+          metaSize = metadataSize proposalMeta
+          param = ProposeParams
+            { ppFrom = addr
+            , ppFrozenToken = metaSize
+            , ppProposalMetadata = proposalMeta
+            }
+      in makeProposalKey param
+    ) <$> addrs
+
+-- | Test functions that operate on `Maybe ProposalDoublyLinkedList`, and compare
+-- its result to similar function that operate on `[ProposalKey]`.
+test_Plist :: [TestTree]
+test_Plist =
+  [ testProperty "Conversion from and to `list`" $ withTests 50 $ property $ do
+      keyList <- forAll $ genProposalKeyList 0
+
+      keyList === (plistToList $ plistFromList keyList)
+
+  , testProperty "Compare `mem` function result between list and plist" $ withTests 50 $ property $ do
+      keyList <- forAll $ genProposalKeyList 1
+
+      (i :: Int) <- forAll $ Gen.integral (Range.constant 0 (length keyList - 1))
+      let k = keyList !! i
+
+      (k `elem` keyList) === (plistMem k $ plistFromList keyList)
+
+    , testProperty "Compare `insert` function result between list and plist" $ withTests 50 $ property $ do
+      keyList <- forAll $ genProposalKeyList 0
+      keyList2 <- forAll $ genProposalKeyList 1
+      (i :: Int) <- forAll $ Gen.integral (Range.constant 0 (length keyList2 - 1))
+      let k = keyList2 !! i
+
+      (keyList <> [k]) === (plistToList $ plistInsert k $ plistFromList keyList)
+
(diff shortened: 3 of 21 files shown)


src/types.mligo
```diff
@@ -139,6 +139,24 @@ type delegate =
...
+// Proposal Doubly Linked List
+
+type first_proposal_node = (proposal_key * (proposal_key option)) // The first key and its next key
+type last_proposal_node = (proposal_key * proposal_key) // The last key and its prev key
+
+type proposals_doubly_linked_list =
+  { first: first_proposal_node option
+  ; last: last_proposal_node option
+  ; map: (proposal_key, (proposal_key * proposal_key)) big_map
+  }
```
rinn7e, line 152: @pasqu4le I've finished the ligo implementation. The main type for storing the order of proposals is now:

```ocaml
type first_proposal_node = (proposal_key * (proposal_key option)) // The first key and its next key
type last_proposal_node = (proposal_key * proposal_key) // The last key and its prev key

type proposals_doubly_linked_list =
  { first: first_proposal_node option
  ; last: last_proposal_node option
  ; map: (proposal_key, (proposal_key * proposal_key)) big_map
  }
```

Basically, `first` store the first proposal key, also its next key (can be empty). `last` store the last key and its prev key.

In between `first` and `last`, all the keys are store in `map` field containing their prev and next key.

All the helper function involving this type is store in `common/plist.mligo`.

Hope to get your first impression on this, let me know if I should rename the types and function as well. If all is fine, I will update the STM tests, and ensure the existing tests passed.

rinn7e:
@pasqu4le Took me a while to finish fixing the tests. I also fix some bugs along the way. 

This PR is ready now, hope to get your review on this.

src/types.mligo
```diff
@@ -139,6 +139,25 @@ type delegate =
...
+
+type first_proposal_node = (proposal_key * (proposal_key option)) // The first key and its next key
+type last_proposal_node = (proposal_key * proposal_key) // The last key and its prev key
+
+type proposal_doubly_linked_list =
+  [@layout:comb]
+  { first: first_proposal_node option
+  ; last: last_proposal_node option
+  ; map: (proposal_key, (proposal_key * proposal_key)) big_map
+  }
```
pasqu4le, line 153: I've considered this quite a bit and I think that I came up with a different representation that would be more efficient in size as well as cost (of running functions on it).

So AFAIK this data structure needs to support a handful of operations:
- `mem` to find if a `proposal_key` is in the dlist
- `pop` to get the first item and remove it from the dlist
- `push` to put a new item at the end of the dlist
- `delete` to remove an item regardless of where it is in the dlist

My main considerations are this:
- `push` and `pop` only need to ever update a "link" in one direction (each)
- `remove` needs most of the times to update a "link" in both direction (except when there is a single element)
- `remove` is less commonly used than `push` and `pop`, so the last two should be favored
- `mem` however needs to be as cheap as possible
- when the list is empty, we don't need to keep track of any information (except that it's empty)
- it would be cheaper to have no `option`s in the types as much as possible

So, here's what came to my mind to store this instead:
1. make `proposal_doubly_linked_list` only track non-empty dlists, so in the `storage` the type of the field would not be `proposal_doubly_linked_list` but rather `proposal_doubly_linked_list option`. So when the list is empty this is `None`.
2. make `first` and `last` be only a `proposal_key`. If there is a single element their value is equal. 
3. in the `big_map` store "link" separately depending on direction.
   Let's use a type synonym `direction` for a `bool` and say that "next" is `true` and "previous" is false.
   What I propose is that the key for the `big_map` is `(proposal_key, direction)` and that the value is a `proposal_key`.
   - If a proposal is the `first` then there is no `(proposal_key, false)` in the map for it.
   - If a proposal is the `last` then there is no `(proposal_key, true)` in the map for it.
   - If there is a single proposal in the whole dlist, then there is no value for it in the list at all (combining the two previous points)
   - In all other cases there are two entries for each `proposal_key`, one with `true` and another with `false`.

This is perhaps a bit more complex, but as you can see:
- there are no `option` in the types (except a single one when the list is empty)
- I think that the cost of the operations above will be much smaller (see next comment)

pasqu4le: To summarize the operations and how they are cheaper (here I'm using `proposal_id` as the value of the input `proposal_key`)

for `mem`:
- if there is no dlist -> false
- if the `proposal_id` is equal to `first` -> true
- if there is a `(proposal_id, false)` in `big_map` -> true
- otherwise -> false

for `push`:
- if there is no dlist -> `first` and `last` are `proposal_id` and the `big_map` is empty
- otherwise:
   -  push `(last, false)` in the `big_map` with a value of `Some proposal_id`
   - set `last` to `proposal_id`

Similarly `pop` only needs to care about `first` and the `(first, true)` key of the `big_map`.

Only `delete` may need to update two values of the `big_map`, but it will still take less comparisons.


@rinn7e  what do you think? Does this sound like an improvement to you?

rinn7e: @pasqu4le Spent sometimes thinking about your suggestion, and definitely sounds good to me. I was thinking about how to store the proposal link separately as well initially, but didn't come up with a good solution. 

I'll start working on the changes, thanks a lot Pasquale

src/types.mligo
```diff
@@ -147,7 +166,7 @@ type storage =
   ; metadata : metadata_map
   ; extra : contract_extra
   ; proposals : (proposal_key, proposal) big_map
-  ; proposal_key_list_sort_by_level : (blocks * proposal_key) set
+  ; proposal_doubly_linked_list: proposal_doubly_linked_list
```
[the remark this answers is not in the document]
rinn7e: Updated this.

src/common/plist.mligo
```diff
@@ -0,0 +1,142 @@
...
+    | None -> Some
+        { first = key
+        ; last = key
+        ; map = (Big_map.empty : ((proposal_key * plist_direction), proposal_key) big_map)
+        }
+    | Some plist -> Some
+          { plist with
+            last = key
+          ; map =
+              Map.add (plist.last, true) key // Add next link of the old last key
```
sras, line 31: I think it would be better to alias `true` and `false` as something similar to `next` and `previous`, and use it in code.

haskell/test/SMT/Model/BaseDAO/Plist.hs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+
+plistDelete :: ProposalKey -> Maybe ProposalDoublyLinkedList -> Maybe ProposalDoublyLinkedList
+plistDelete key plistMb =
+  case plistMb of
+    Nothing ->
+      error "BAD_STATE: no proposal"
+    Just plist ->
+      if (plist & plFirst) == key then
```
sras, line 54: Is there any advantage in using `&` instead of simply writing `plFirst plist`?

Just want to share that I find it a bit harder than I would like to admit, to read this code even when I understand how `&` works.

src/types.mligo
```diff
@@ -139,6 +139,22 @@ type delegate =
...
+// Use to query the previous and next of a proposal.
+// false = previous
+// true = next
+type plist_direction = bool
+
+// Proposal Doubly Linked List
+type proposal_doubly_linked_list =
+  [@layout:comb]
+  { first: proposal_key
+  ; last: proposal_key
```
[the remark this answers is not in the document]
rinn7e: > is first the first one insterted, or the first one in the list?
Hmm, `first` here refer to the first one in the list, I'll add some comments about it.

haskell/test/SMT/Model/BaseDAO/Plist.hs
rinn7e: I choose to do since it resembles `plist.first` in ligo, which makes it easier to compare the ligo and haskell implementation.

src/common/plist.mligo
```diff
@@ -0,0 +1,142 @@
...
+                        Some { new_plist with first = key }
+
+                )
+            | None ->
+                ([%Michelson ({| { FAILWITH } |} : (nat * string) -> (proposal_doubly_linked_list option))]
+                    (bad_state, "Error updating head with a key that does not exist.") : (proposal_doubly_linked_list option))
+          )
+
+
+#endif // COMMON_PLIST_H included
```
sras, line 142: It seems that we have quite a bit of logic here.

Can we add hedgehog property tests to ensure the soundness of logic here (May be by comparing the behviour to data structures from `ordered-containers` package, since this seem to implement the same functionality?). 

I think we should also have separate SMT tests to target this section, as the subset of main SMT tests might not provide enough variance to properly test this section.

rinn7e: Make sense, I'll try to work on this and see if it could work out.

haskell/test/Test/Ligo/BaseDAO/Plist.hs
```diff
@@ -0,0 +1,102 @@
...
+          param = ProposeParams
+            { ppFrom = addr
+            , ppFrozenToken = metaSize
+            , ppProposalMetadata = proposalMeta
+            }
+      in makeProposalKey param
+    ) <$> addrs
+
+
+test_Plist :: [TestTree]
```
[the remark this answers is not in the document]
rinn7e: Add comments and description about each tests.

haskell/test/Test/Ligo/BaseDAO/Plist.hs
```diff
@@ -0,0 +1,102 @@
...
+      -- It's much easier to test updateHead f against a list.
+      (listUpdateHead k keyList) === (plistToList $ plistUpdateHead k plist)
+  ]
+    where
+      listUpdateHead k l =
+        case l of
+          (x:xs) ->
+            if (x == k) then l
+            else listUpdateHead k xs
+          _ -> []
```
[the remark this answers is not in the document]
rinn7e: Really not sure how to do this, also the main operations that modify the list are only `add` and `delete`, so I'm not sure if this is worth doing or not.

src/common/plist.mligo
```diff
@@ -0,0 +1,146 @@
...
+                  )
+                )
+              )
+          }
+        )
+
+
+
+[@inline]
+let plist_update_head (key, plist_o: proposal_key * proposal_doubly_linked_list option): proposal_doubly_linked_list option =
```
pasqu4le, line 103: AFAIU this sets the given key as head, but why do we need this function?

I thought that we only needed to `pop` the head of the list, not put it back. 

Could you explain why this is needed in `flush_one` please?

src/proposal.mligo
```diff
@@ -330,31 +327,93 @@ let handle_proposal_is_over
...
-      Set.fold flush_one store.proposal_key_list_sort_by_level (nil_op, store, counter)
+    let init = (nil_op, config, store, counter) in
+
+    let (ops, _, store, counter) =
+      fold_with_flush_one (store.ongoing_proposals_dlist, init)
     in
     // prevent empty flushes to avoid gas costs when unnecessary.
     if counter.current = 0n
     then (failwith empty_flush : return)
     else (ops, store)
```
pasqu4le, line 412: I see that this has gotten a lot more complex, couldn't it be handled with a single recursive function?

Such function would:
1. check if the number `n` was reached or the dlist is empty, return if so
2. `pop` the first element off of the dlist (updating the dlist in the storage)
3. `handle_proposal_is_over` on that first element (possibly failing)
4. repeat with the updated store and `n` decreased by one

Does that sound correct or am I missing something?

src/common/plist.mligo
rinn7e: Updated this.

src/proposal.mligo
```diff
@@ -106,14 +113,16 @@ let update_delegates (params, store : update_delegate_params * storage): return
...
+ * Unstake voter's tokens on a proposal that has already been flushed or dropped.
+ * Fail if the voter did not vote on that proposal, or voter has already unfreezed, or
+ * the proposal was not yet flushed nor dropped.
+ *)
 let unstake_vote_one (config: config) (store , proposal_key : storage * proposal_key): storage =
 
   // Ensure proposal is already flushed or dropped.
-  let p = fetch_proposal (proposal_key, store) in
-  let _ = if Set.mem (p.start_level, proposal_key) store.proposal_key_list_sort_by_level
+  let _ = fetch_proposal (proposal_key, store) in
```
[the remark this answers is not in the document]
rinn7e: Not sure if github is bugged, since the suggested changes remain the same. I've removed `fetch_proposal` anyway.

src/proposal.mligo
rinn7e: This definitely simplifies thing a lot. The previous implementation try to keep `handle_proposal_is_over` the same, but I think your suggestion is better.

I've updated this now but I slightly modify this by adding a check to ensure that `plist.first` is able to be flushed in the stage 1.

src/common/plist.mligo
rinn7e: Probably the naming is off a bit, but what the function does here is like `pop until reaching a proposal key`.

This is no longer necessary though, I've added a normal pop function, and update the flush implementation according to your suggestion below.

src/common/plist.mligo
```diff
@@ -0,0 +1,146 @@
...
+            map =
+              (Map.add (next_key, prev) prev_key // Update the next link
+                (Map.add (prev_key, next) next_key // Update the previous link
+                  (Map.remove (key, next) // Remove the next link of the key.
+                    (Map.remove (key, prev) plist.map) // Remove the prev link of the key.
+                  )
+                )
+              )
+          }
+        )
```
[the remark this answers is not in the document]
rinn7e: This is definitely a lot simpler, couldn't find any problem, I've updated this now.

src/proposal.mligo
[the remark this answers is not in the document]
pasqu4le: oh, it looks like I actually submitted the wrong suggestion, my bad.

src/proposal.mligo
```diff
@@ -277,84 +282,102 @@ let do_total_vote_meet_quorum_threshold (proposal, store: proposal * storage): b
...
+  in
+  let cons = fun (l, e : operation list * operation) -> e :: l in
+  let ops = List.fold cons ops new_ops in
+  (ops, store)
+
+
+(*
+ * Flush a proposal, starting from the first. Remove it when the flush is done, and move to the next proposal.
+ *)
+let rec flush_one (counter, config, store, ops : counter * config * storage * operation list): (operation list * storage * counter) =
```
pasqu4le, line 330: Minor nitpick: I think this should be renamed, given that it no longer flushes one proposal, but one proposal per iteration.

Maybe something like `flush_each` would be better.

src/proposal.mligo
```diff
@@ -277,84 +282,102 @@ let do_total_vote_meet_quorum_threshold (proposal, store: proposal * storage): b
...
 
-// Flush all proposals that passed their voting stage.
+(*
+ * Flush all proposals that passed their voting stage.
+ *)
 let flush(n, config, store : nat * config * storage): return =
   if n = 0n
   then (failwith empty_flush : return)
   else
     let counter : counter = { current = 0n; total = n } in
```
pasqu4le, line 369: :thinking: I think that perhaps this counter type worked better before these changes, but I may be wrong.

Have you tried instead to pass the `n` total to `flush_one` and reducing it by 1 at each iteration?
Stopping when it reaches `0` should have the same result.

src/common/plist.mligo
```diff
@@ -0,0 +1,98 @@
+// SPDX-FileCopyrightText: 2021 TQ Tezos
+// SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+#if !COMMON_PLIST_H
```
pasqu4le, line 4: I just realized that (based on how we did it until now) this module would be perhaps better located under the `src/proposal/` directory

src/common/plist.mligo
```diff
@@ -0,0 +1,98 @@
...
+            | Some next_key ->
+                Some
+                  { plist with
+                    first = next_key // Make the next as the first key
+                  ; map =
+                      Map.remove (next_key, prev) // Remove the previous link of the next key.
+                        (Map.remove (plist.first, next) plist.map) // Remove the first key link.
+                  }
+            | None -> (None : proposal_doubly_linked_list option) // Only the first key exist.
+        )
```
pasqu4le, line 33: This definitely looks more like what I had in mind and I don't have any comment on the logic (or its efficiency), however I think that it should be changed to return the first key itself (if any) as well as the updated list:
```suggestion
let plist_pop (plist_o : proposal_doubly_linked_list option)
    : (proposal_key option * proposal_doubly_linked_list option) =
  match plist_o with
  | None ->
    ( (None : proposal_key option)
    , (None : proposal_doubly_linked_list option)
    )
  | Some plist ->
    let new_plist = match Map.find_opt (plist.first, next) plist.map with
      | Some next_key ->
        Some
          { plist with
            first = next_key // Make the next as the first key
          ; map =
              Map.remove (next_key, prev) // Remove the previous link of the next key.
                (Map.remove (plist.first, next) plist.map) // Remove the first key link.
          }
      | None -> (None : proposal_doubly_linked_list option) // Only the first key exist.
    in ( Some plist.first, new_plist)
```
because it would save more work in its `proposal.mligo` use (see next comment).

src/proposal.mligo
```diff
@@ -277,84 +282,102 @@ let do_total_vote_meet_quorum_threshold (proposal, store: proposal * storage): b
...
+              // Continue to the next proposal
+              let counter = { counter with current = counter.current + 1n } in
+              flush_one (counter, config, store, ops)
+
+            else
+              (ops, store, counter)
+        else
+          (ops, store, counter)
+    | None ->
+        (ops, store, counter)
```
pasqu4le, line 360: If `plist_pop` returns the first element, then in a single call we can get the updated list and only need to care take care of the head key
```suggestion
let rec flush_one (counter, config, store, ops : counter * config * storage * operation list): (operation list * storage * counter) =
  let (plist_head, plist_new) = plist_pop store.ongoing_proposals_dlist in
  // Ensure plist is not empty.
  match plist_head with
    | Some first_key ->
        // Ensure counter is not reached yet.
        if (counter.current < counter.total) then
          let proposal = check_if_proposal_exist (first_key, store) in

          // Ensure the proposal is not expired.
          let _ = if is_proposal_age (proposal, config.proposal_expired_level)
              then (failwith expired_proposal : unit)
              else unit in
          // Ensure the the proposal is flushable.
          if (is_proposal_age (proposal, config.proposal_flush_level))
            then
              // Unstake tokens, and execute proposal's decision_lambda.
              let (ops, store) = handle_proposal_is_over (config, proposal, store, ops) in

              // Update the stored plist.
              let store = { store with ongoing_proposals_dlist = plist_new } in

              // Continue to the next proposal
              let counter = { counter with current = counter.current + 1n } in
              flush_one (counter, config, store, ops)
            else
              (ops, store, counter)
        else
          (ops, store, counter)
    | None ->
        (ops, store, counter)
```
which is a bit of an improvement.

BTW I think that rearranging the checks here might avoid us to doreturn `(ops, store, counter)` in 3 separate locations and improve this even more.

src/proposal.mligo
pasqu4le: I think that this is much better, I have added some suggestion to tweak it a little bit more, let me know what do you think about those.

src/proposal.mligo
rinn7e: Renamed this.

src/proposal.mligo
rinn7e: Make sense, I've removed counter, and just use `n` instead.

src/common/plist.mligo
rinn7e: Updated this.

src/common/plist.mligo
rinn7e: I was thinking that since we can access first element easily, we don't need to return it, but as you mentioned below, it definitely helps in our use case. I've updated this.

src/proposal.mligo
rinn7e: Updated this, also re-arranged the check a bit, it returns `(ops, store, counter)` in 2 places now.

src/proposal.mligo
rinn7e: @pasqu4le I've made all the changes according to your suggestions, also update the haskell SMT and tests. This should be ready for re-review.

src/proposal.mligo
```diff
@@ -277,84 +282,93 @@ let do_total_vote_meet_quorum_threshold (proposal, store: proposal * storage): b
...
+        (ops, store, n)
 
-// Flush all proposals that passed their voting stage.
+(*
+ * Flush all proposals that passed their voting stage.
+ *)
 let flush(n, config, store : nat * config * storage): return =
   if n = 0n
   then (failwith empty_flush : return)
   else
```
pasqu4le, line 360: Couldn't this be removed now?

IIUC `flush_each` will return right away if `n` is `0` and other check after that will fail with `empyt_flush` just the same, right?

docs/specification.md
```diff
@@ -523,7 +523,6 @@ Parameter (in Michelson):
 - Fails with `NOT_PROPOSING_STAGE` if the current stage is not a proposing one.
 - Fails with `FAIL_PROPOSAL_CHECK` if the proposal is rejected by `proposal_check`
   from the configuration.
-- Fails with `MAX_PROPOSALS_REACHED` if the current amount of ongoing proposals is at max value set by the config.
```
pasqu4le, line 526: There's also a mention of `max_proposals` in the "Configuration" section that should be removed (line 95 right now).

pasqu4le (commented):
I have a couple small comments, but otherwise this LGTM.

@sras I see that you haven't resolved your comments nor approved yet, can you take another look?

src/proposal.mligo
rinn7e: Make sense, I've updated this.

docs/specification.md
rinn7e: Removed this now.

pasqu4le (approved):

haskell/test/Test/Ligo/BaseDAO/Plist.hs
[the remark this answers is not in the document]
sras: >so I'm not sure if this is worth doing or not.

May be I am missing something, but in implementations like this, bug are often hidden in some specific combination of operations. For example, add 5 items, delete 5 items, and then add another item. Normally we don't have to write tests for them because we don't usually have to implement core data structures ourself.

But since we are doing that here, and we are doing that in a smart contract for something that is part of its core operation, I think we should spare no efforts in testing this implementation.  @pasqu4le please correct me if I am wrong here.

>Really not sure how to do this,

I was thinking something like this.  Write a small contract with storage type `plist`, and entrypoints to load it, then entrypoints to  add/delete from it. With this, we can implement the property tests by generating a list of operations, and implementing the tests the same way we implement SMT tests for the main contract.

src/common/plist.mligo
sras: Have this been update everywher, for example https://github.com/tezos-commons/baseDAO/pull/318/files/1a946282438110ec0b534b350413a353d8db1eca#diff-41e7ba34a244b5cdc2749fe8e91f9dfb4aede990d497e9fe2933ff48a22f049eR86 ?

sras: Sorry. Was looking at the wrong commit, it seems.

haskell/test/Test/Ligo/BaseDAO/Plist.hs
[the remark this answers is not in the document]
sras: If this makes sense, I am ok with adding the tests in another issue/PR.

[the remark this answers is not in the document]
rinn7e: If needed, I'm good with creating another issue for it as well.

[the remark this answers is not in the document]
rinn7e: @sras Created the issue: https://github.com/tezos-commons/baseDAO/issues/319, let me know if there are more changes you want to be made.

haskell/test/SMT/Model/BaseDAO/Proposal.hs
```diff
@@ -331,7 +328,12 @@ applyDropProposal mso proposalKey = do
     || proposalIsExpired
   then do
     unstakeProposerToken False proposal
-    removeFromProposalSortByLevel (proposal & plStartLevel) proposalKey
+    modifyStore $ \s ->
+      pure $ s
+        { sOngoingProposalsDlist = plistDelete proposalKey (s & sOngoingProposalsDlist)
+        }
+
+
```
sras, line 336: @rinn7e Redundant new lines.

Please clean up the changes a bit. I think there are redundant new lines in a number of places.

rinn7e: @sras Removed the unneeded new lines, this should be good to go.

src/proposal.mligo
```diff
@@ -277,84 +282,89 @@ let do_total_vote_meet_quorum_threshold (proposal, store: proposal * storage): b
...
-              })
-      else (nil_op, store)
-    in
-    let cons = fun (l, e : operation list * operation) -> e :: l in
-    let ops = List.fold cons ops new_ops in
-    let store = remove_from_proposal_sort_by_level (start_level, proposal_key, store) in
-    (ops, store, counter)
-  else (ops, store, counter)
+        // Ensure counter is not reached yet and the the proposal is flushable.
+        if (n > 0 && is_proposal_age (proposal, config.proposal_flush_level)) then
```
sras, line 342: @rinn7e Why is `is_proposal_age` called here a second time? AFAIU we are calling it here https://github.com/tezos-commons/baseDAO/pull/318/files?file-filters%5B%5D=.mligo&show-viewed-files=false#diff-9ca2a645567904e8f5b557f933b05a7a497e46c5852951c9320c693a6a20590cR337

rinn7e: One is checked with `config.proposal_expired_level`, another is checked `config.proposal_flush_level`. Let me know if this makes sense to you.

sras: Ah, yes.

sras (approved):
Minor nitpick. It seems that there is still a bunch of redundant newlines, like in this file https://github.com/tezos-commons/baseDAO/pull/318/files?file-filters%5B%5D=.hs&show-viewed-files=false#diff-0ec2c4866849a227f5d2c49ff069bd0bfb0bbf512b9b3bf1d7f560cbb8e1b2f6R33

haskell/test/Test/Ligo/BaseDAO/Plist.hs
```diff
@@ -0,0 +1,90 @@
...
+          param = ProposeParams
+            { ppFrom = addr
+            , ppFrozenToken = metaSize
+            , ppProposalMetadata = proposalMeta
+            }
+      in makeProposalKey param
+    ) <$> addrs
+
+-- | Test functions that operate on `Maybe ProposalDoublyLinkedList`, and compare
+-- its result to similar function that operate on `OSet ProposalKey`.
```
sras, line 48: Oh this comment probably needs an update

haskell/test/Test/Ligo/BaseDAO/Plist.hs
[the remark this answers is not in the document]
pasqu4le: I agree, normally the current tests would be enough IMO, but given that is based on a new (fairly complex) data structure we should test this more carefully.

I think it was a good idea to merge this PR now, but let's not put these tests off for much longer
@rinn7e I assigned the new issue to you

PR #320, opened 2022-02-28
[#319] Add `plist` ligo contract, add property tests for it

### LINKED ISSUE #319: Add more tests for `plist` related functions

Currently, we only tests plist utility functions one at a time. It may be helpful to test combinations of them at the same time.

More explanation can be found here: https://github.com/tezos-commons/baseDAO/pull/318#discussion_r810755672

# Acceptance criteria

Tests are added according to explanation above.

rinn7e:
## Description

Problem: Currently, we only tests plist utility functions one at
a time. It may be helpful to test combinations of them at the same time.

Solution: Add `plist` ligo contract, and add property tests for it
by comparing it with haskell list implementation.

## Related issue(s)

Resolves #319

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Test/Plist/Property.hs
+++ b/haskell/test/Test/Plist/Property.hs
@@ -1,0 +1,134 @@
+-- SPDX-FileCopyrightText: 2021 Tezos Commons
+-- SPDX-License-Identifier: LicenseRef-MIT-TC
+
+module Test.Plist.Property
+  ( hprop_Plist
+  ) where
+
+import Universum hiding (drop, swap)
+
+import qualified Data.List as DL
+import Data.List ((!!))
+import Fmt (build, unlinesF)
+import Hedgehog hiding (assert)
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+import Lorentz hiding (abs, assert, and, div, not, now, (>>))
+import Test.Cleveland
+
+import Ligo.BaseDAO.Types
+import SMT.Model.BaseDAO.Proposal.Plist
+import Test.Ligo.BaseDAO.Plist (genProposalKeyList)
+import Test.Plist.Type
+import Test.Plist.Contract
+
+
+genPlistParameter :: MonadGen m => [ProposalKey] -> [ProposalKey] -> m PlistParameter
+genPlistParameter keyList keyList2 = do
+  requireExistingKeyEps <-
+    if (not $ null keyList) then do
+      (i :: Int) <- Gen.integral (Range.constant 0 (length keyList - 1))
+      let k = keyList !! i
+      pure [ Mem k
+           , Delete k
+           ]
+    else pure []
+
+  notRequireExistingKeyEps <-
+    if (not $ null keyList2) then do
+      (i2 :: Int) <- Gen.integral (Range.constant 0 (length keyList2 - 1))
+      let k2 = keyList2 !! i2
+      pure [ Insert k2
+           , Mem k2
+           , Delete k2
+           ]
+    else pure []
+
+  Gen.choice $ pure <$> (requireExistingKeyEps <> notRequireExistingKeyEps <> [Pop ()])
+
+genPlistParameters :: MonadGen m => m ([ProposalKey], [PlistParameter])
+genPlistParameters = do
+  initSize <- Gen.integral (Range.constant 0 30)
+  initSize2 <- Gen.integral (Range.constant 0 30)
+  -- Init List to be used as storage
+  keyList <- genProposalKeyList initSize
+  -- A list to pick new key from
+  keyList2 <- genProposalKeyList initSize2
+  params <- Gen.list (Range.linear 10 20) $ genPlistParameter keyList keyList2
+  pure (keyList, params)
+
+
+hprop_Plist :: Property
+hprop_Plist =
+  withTests 100 $ property $ do
+    plistTests
+
+plistTests :: PropertyT IO ()
+plistTests = do
+  (keyList, params) <- forAll genPlistParameters
+  testScenarioProps $
+    scenarioEmulated $ do
+      let defaultStore = PlistStorage (plistFromList keyList) False Nothing
+      let haskellStore = (keyList, False, Nothing)
+      -- Originate plist contract
+      plistContract <- originateTypedSimple @PlistParameter "PlistContract" defaultStore plistContractLigo
+      callContractLoop params (TAddress $ chAddress plistContract) haskellStore
+
+callContractLoop :: MonadEmulated caps base m => [PlistParameter] -> TAddress PlistParameter -> ([ProposalKey], Bool, Maybe ProposalKey) -> m ()
+callContractLoop param addr haskellStore =
+  case param of
+    p:ps -> do
+      let newHaskellStore = callHaskell haskellStore p
+      newLigoStore <- callLigo addr p
+      assert (newHaskellStore == newLigoStore) $
+        unlinesF
+          [ "━━ Error: Haskell and Ligo storage are different ━━"
+          , build p
+          , "━━ Haskell storage ━━"
+          , build newHaskellStore
+          , "━━ Ligo storage ━━"
+          , build newLigoStore
+          ]
+      callContractLoop ps addr newHaskellStore
+    [] -> pure ()
+
+-- | Call haskell implementation which uses normal list.
+callHaskell :: ([ProposalKey], Bool, Maybe ProposalKey) -> PlistParameter -> ([ProposalKey], Bool, Maybe ProposalKey)
+callHaskell (keyList, memResult, popResult) param =
+  case param of
+    Insert k ->
+      let newKeyList = if not (k `elem` keyList)
+            then keyList <> [k]
+            else keyList
+      in (newKeyList, memResult, popResult)
+    Delete k -> (DL.delete k keyList, memResult, popResult)
+    Mem k -> (keyList, k `elem` keyList, popResult)
+    Pop _ ->
+      let (list, newPopResult) = case DL.uncons keyList of
+            Just (h, r) -> (r, Just h)
+            Nothing -> ([], Nothing)
+      in (list, memResult, newPopResult)
+
+-- | Call ligo plist contract.
+callLigo :: MonadEmulated caps base m => TAddress PlistParameter -> PlistParameter -> m ([ProposalKey], Bool, Maybe ProposalKey)
+callLigo addr param = do
+  nettestResult <- attempt @TransferFailure $ case param of
+    Insert k -> do
+      PlistStorage{..} <- getFullStorage @PlistStorage (unTAddress addr)
+
+      -- Ensure that the key being inserted does not exist in the list. In the main contract, this is checked by
+      -- `check_if_proposal_exist`.
+      if (k `elem` plistToList psPlist) then
+        pure ()
+      else
+        call addr (Call @"Insert") k
+    Delete k -> call addr (Call @"Delete") k
+    Mem k -> call addr (Call @"Mem") k
+    Pop _ -> call addr (Call @"Pop") ()
+
+  case nettestResult of
+    Right _ -> do
+      PlistStorage{..} <- getFullStorage @PlistStorage (unTAddress addr)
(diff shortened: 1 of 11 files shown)


rinn7e:
@sras This PR is ready for review, let me know if this solution is what you have in mind.

haskell/test/Test/Plist/Property.hs
```diff
@@ -0,0 +1,128 @@
...
+    , pure $ Mem k2
+    , pure $ Mem k
+    , pure $ Delete k
+    , pure $ Pop ()
+    ]
+
+
+genPlistParameters :: MonadGen m => m ([ProposalKey], [PlistParameter])
+genPlistParameters = do
+  initSize <- Gen.integral (Range.constant 1 10)
```
sras, line 44: Shouldn't range start from `0`?  Shouldn't we also include empty lists in the generated possibilites?

haskell/test/Test/Plist/Property.hs
```diff
@@ -0,0 +1,128 @@
...
+      callContractLoop params (TAddress $ chAddress plistContract) defaultStorage
+
+callContractLoop :: MonadEmulated caps base m => [PlistParameter] -> TAddress PlistParameter -> PlistStorage -> m ()
+callContractLoop param addr haskellStore =
+  case param of
+    p:ps -> do
+      let newHaskellStore = callHaskell haskellStore p
+      newLigoStore <- callLigo addr p
+
+      assert (newHaskellStore == newLigoStore) $
```
sras, line 81: Shouldn't we also check the result of `mem` operation is the same for both?

sras, line 81: Here it seems that we are using the `Eq` instance of `ProposalDoublyLinkedList` for comparision. I think it is safer to convert the `ProposalDoublyLinkedList` to a list and compare it to the result list from Haskell version. So I think both callLigo and callHaskell should return the list along with memresult.

rinn7e: Make sense, I've udpated both `callLigo` and `callHaskell` to return `([ProposalKey], Bool)` instead now.

sras:
@rinn7e When I run these tests locally, it hangs sometimes. Can you please check on your side?

src/proposal/plist.mligo
```diff
@@ -57,9 +57,8 @@ let plist_insert (key, plist_o : proposal_key * (proposal_doubly_linked_list opt
 let plist_delete (key, plist_o : proposal_key * proposal_doubly_linked_list option): proposal_doubly_linked_list option =
   match plist_o with
     | None ->
-        // Expect to have at least a proposal to delete it.
-        ([%Michelson ({| { FAILWITH } |} : (nat * string) -> proposal_doubly_linked_list option)]
-            (bad_state, "no proposal") : proposal_doubly_linked_list option)
+        // Do nothing in case of no proposal key.
+        plist_o
```
sras, line 61: Also, why was this required? I mean, Is this meant to be reverted during merge?

src/proposal/plist.mligo
```diff
@@ -39,27 +39,28 @@ let plist_pop (plist_o : proposal_doubly_linked_list option)
...
+          ; map = (Big_map.empty : ((proposal_key * plist_direction), proposal_key) big_map)
           }
+      | Some plist -> Some
+            { plist with
+              last = key
+            ; map =
+                Map.add (plist.last, next) key // Add next link of the old last key
+                  (Map.add (key, prev) plist.last plist.map) // Add previous link of the new last key
+            }
+  else plist_o // Shouldn't happen, key not existed check should be done before calling this function.
```
rinn7e, line 56: > When I run these tests locally, it hangs sometimes.

@sras Took me a while to figure this out. This is because the generator sometimes generate the same key that already exists, and inserting that key for some reason cause the hang. Not sure why this happens, I add a check to ensure the inserted key does not exist yet.

This wouldn't happen normally in baseDao, since we have a check to ensure the key is not exist yet. But in our plist contract, we don't have it.

src/proposal/plist.mligo
rinn7e: Most delete func (Data.List.delete, Map.delete) does not throw error when the key does not exist. So I want to keep the same functionality here (It is an impossible state anyway, since we use plist_delete on key that we're sure it exists).

src/proposal/plist.mligo
rinn7e: Also fixed some other errors, this should be good for re-review.

sras: > inserting that key for some reason cause the hang

I think we should find the reson for this happening, and after that may be we can ignore it.

haskell/test/Test/Plist/Property.hs
```diff
@@ -0,0 +1,140 @@
...
+      callContractLoop params (TAddress $ chAddress plistContract) haskellStore
+
+callContractLoop :: MonadEmulated caps base m => [PlistParameter] -> TAddress PlistParameter -> ([ProposalKey], Bool) -> m ()
+callContractLoop param addr haskellStore =
+  case param of
+    p:ps -> do
+      let newHaskellStore = callHaskell haskellStore p
+      newLigoStore <- callLigo addr p
+
+      assert (newHaskellStore == newLigoStore) $
```
sras, line 92: Oh, are we also comparing the result of `pop` calls somehow?

src/proposal/plist.mligo
sras: I think when we test something, we need to isolate the entity that is being tested, and define and isolate the behavior that is being tested, and write tests for that behavior. And I think we should not consider how some other code will be using the code under test, when writing tests, becaue that external code might use it in a different way in future.

So I think in these tests, we should document the expected behavior of plist. If it is close to the behavior of data structure X, we can say that plist behave mostly like X, except for so and so cases, and that is the behavior that we are testing. 

This is also regarding this thread https://github.com/tezos-commons/baseDAO/pull/320/files#r819505802 regarding the insertion behavior.

haskell/test/Test/Plist/Property.hs
```diff
@@ -0,0 +1,140 @@
...
+            then keyList <> [k]
+            else keyList
+      in (newKeyList, memResult)
+    Delete k -> (DL.delete k keyList, memResult)
+    Mem k -> (keyList, k `elem` keyList)
+    Pop _ ->
+      let list = (\case
+                      Just (_, r) -> r
+                      Nothing -> []
+                  ) $ DL.uncons $ keyList
```
sras, line 122: Is this just 

```
case DL.uncons keyList of
   Just  (_, r) -> r
   _  -> []
``` 

??

src/proposal/plist.mligo
rinn7e: The closest data structure to plist is a unique element list (or a set that is ordered by its insertion time). Also if the same element is inserted it shouldn't overwrite the existing element. This is the expected behavior, I'll add this as a comments.

haskell/test/Test/Plist/Property.hs
rinn7e: I only compared the updated list, I've added the head result for the comparison as well.

haskell/test/Test/Plist/Property.hs
rinn7e: Updated.

sras:
@rinn7e When I completely delete the `plist_delete` and  `plist_insert` function (in original LIGO code) and replace it with a nop code, and then run these tests, they still appear to pass. The main SMT tests pass as well.

But regular integrational tests fail.

Can you please try this and see if you are getting the same result?

rinn7e:
> @rinn7e When I completely delete the `plist_delete` and `plist_insert` function (in original LIGO code) and replace it with a nop code, and then run these tests, they still appear to pass. The main SMT tests pass as well.
> 
> But regular integrational tests fail.
> 
> Can you please try this and see if you are getting the same result?

@sras Looks like the BaseDAO smt was passing because the proposal min token is too high and no proposal get through (RegistryDAO and TreasuryDAO still fail as expected). I've fixed this now by setting the min token lower.

The property tests for plist functions still fail as expected:
image

Not sure why it does not fail for you, did you compile the plist contract? (by using `make all` or `make out/plist_contract.tz`).

src/proposal/plist.mligo
sras: @rinn7e Since we are modifying the plist insert by adding a membership check, I think we should make sure that loop is indeed in the ligo code. When I tested, it seems that if I comment the assert statement in the test, the loop disappears. Same happen if I only compare the second element of the tuple. Any chance that loop is in the Haskell counterpart?

If that is the case, we would be adding the membership check for no good reason.

rinn7e: That might be because of the laziness. I think you can try to force strict ```!ligoStore``` to force evaluate it which will cause the hang.

rinn7e: What I mean is that, it does not hang, because it does not evaluate the ligoStore. So there is no call to the ligo code at all. If there is no call to ligo code, it obviously won't hang since the problem is in the ligo code. Let me know if you don't get what I mean.

sras: AFAIU this line https://github.com/tezos-commons/baseDAO/pull/320/files#diff-08120fcaa0dee5bc8d0750d807e29874ae66df02d426d6039ab48a5220126c54R42 

with the membership check was added to fix the hanging test, right?

rinn7e: Yeah, this is mainly meant to fix the hanging in the tests.

sras: Yes, so if the test hangs because of some flaw in Haskell code, I think the proper way to fix it is by fixing the Haskell code, and not by adding runtime checks on the Ligo code.

rinn7e: I couldn't figure the flaw in haskell code. If that is the case, hope to get your suggestion on how to fix it without updating the ligo code.

sras: Ok. The issue seems to be in the `plistToList` Haskell function which ends up in a loop when there are duplicate insertions. So the solution would be to detect the loop and cancel the current test. Or better, modify the generator to generate operation lists without duplicate insertions.

rinn7e: @sras So I've removed the check in ligo code and add it in the insert call in `callLigo` instead. This should ensure no duplicate key in the list and avoid updating the ligo code.

src/types.mligo
```diff
@@ -140,7 +140,12 @@ type plist_direction = bool
 let prev = false
 let next = true
 
-// Proposal Doubly Linked List
+(*
+ * Proposal Doubly Linked List
+ *
+ * Behave like a unique element list (or a set ordered by its insertion time.)
+ * When trying to insert an element that already exists, the list remains unchanged.
```
sras, line 147: >Behave like a unique element list (or a set ordered by its insertion time.)

I think it is better to just say an [OrderdSet](https://hackage.haskell.org/package/ordered-containers-0.2.2/docs/Data-Set-Ordered.html)?

>  * When trying to insert an element that already exists, the list remains unchanged.

But this is wrong. Right? AFAIU If you try to insert an existing element, then the data structure becomes corrupt.

rinn7e: Fixed the description.

src/types.mligo
```diff
@@ -140,7 +140,13 @@ type plist_direction = bool
 let prev = false
 let next = true
 
-// Proposal Doubly Linked List
+(*
+ * Proposal Doubly Linked List
+ *
+ * Behave like `OrderedSet`.
+ * When inserting a new key, it must be sure that it does not exist in the list or else
```
sras, line 147: > it must be sure that

"It should be ensured that" sounds a bit better..

rinn7e: Updated this.

sras:
@rinn7e Can you please rebase

rinn7e:
> @rinn7e Can you please rebase

@sras I've rebased with master now, have to squash the commits to fix come conflicts.

sras (approved):
Approving. Please clean up all the redundant newlines before merging.

sras (approved):
Approving. But please clean up some redundant newlines that appear to be present in the changes before merging.

PR #321, opened 2022-03-09
[Chore] Replace hangzhou steps with ithaca in CI

rvem:
## Description
Problem: Tezos has a new protocol and we should check if this project is
compatible with it.

Solution: Replace hangzhou local-chain and scheduled testnet jobs with ithaca
in the CI configuration.

## Related issue(s)

https://issues.serokell.io/issue/TM-628

- Tests
        silently reappearing again.

- Documentation
    - Haddock


.buildkite/pipeline.yml
```diff
@@ -4,7 +4,7 @@
 env:
   TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER: "Y"
   # this key is defined in local-chain bootstrap accounts list in
-  # https://github.com/serokell/serokell-profiles/blob/master/profiles/servers/jupiter/default.nix
+  # https://github.com/serokell/aquarius-infra/blob/master/servers/albali/chain.nix#L23
```
pasqu4le, line 7: Links to the wrong like and I would remove it entirely:
```suggestion
  # https://github.com/serokell/aquarius-infra/blob/master/servers/albali/chain.nix
```
because this links to `master` and that line could change.

Alternatively linking to a specific commit would work as well, with the line fixed.

.buildkite/pipeline.yml
```diff
@@ -106,7 +106,7 @@ steps:
     # Note that 'refill-balance' below is the initial 'TASTY_CLEVELAND_MONEYBAG_SECRET_KEY' balance
     # which may need to be adjusted in case of insufficient balance errors
     - $(nix-build ci.nix -A run-chain-tests
-        --argstr refill-balance 200
+        --argstr refill-balance 5000
```
pasqu4le, line 109: :money_with_wings: just a reminder: if this is enough please remember to re-set it to a value close-ish to what was actually used.

pasqu4le (approved):
Approving already because this LGTM aside from two minor comments:

rvem:
I forgot to undraft it :sweat_smile:

Issue #322, opened 2022-03-11
Consider switching to big_maps to store registry and registry_affected in registryDAO


sras:
# Clarification and motivation

Followup issue for this discussion regarding switching to big_maps to store registry and registry_affected values at
 https://github.com/tezos-commons/baseDAO/pull/305#discussion_r824049134

# Acceptance criteria

Decision is made wether to make this switch and the change is implemented.

PR #327, opened 2022-04-28
[#322] Switch to bigmaps to store registry in registryDAO

sras:
Problem: We think it might be better for registryDAO to store registry
and registry affected maps in a BigMap instead of a Map

Solution: Amend the record to change these two fields to a BigMap and
make corresponding changes to the LIGO and Haskell tests.

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/variants/registry/types.mligo
+++ b/src/variants/registry/types.mligo
@@ -8,8 +8,8 @@
 type registry_key = string
 type registry_value = string
 
-type registry = (registry_key, registry_value) map
-type registry_affected = (registry_key, proposal_key) map
+type registry = (registry_key, registry_value) big_map
+type registry_affected = (registry_key, proposal_key) big_map
 type registry_diff = (registry_key * registry_value option) list
 type proposal_receivers = address set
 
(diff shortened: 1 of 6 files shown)


src/variants/registry/types.mligo
```diff
@@ -8,8 +8,8 @@
 type registry_key = string
 type registry_value = string
 
-type registry = (registry_key, registry_value) map
-type registry_affected = (registry_key, proposal_key) map
+type registry = (registry_key, registry_value) big_map
+type registry_affected = (registry_key, proposal_key) big_map
```
pasqu4le, line 12: It seems that this should be updated in the `docs/registry.md` file

pasqu4le (commented):
Only one small comment, however see the one about the `morley` issue too.

I think it would be nice to handle this right away in this PR, but I'm ok with making a separate one too.

src/variants/registry/types.mligo
sras: Have changed.

pasqu4le (approved):

PR #328, opened 2022-05-24
[#263] Cleanup TODOs w.r.t use of deriveRPC

### LINKED ISSUE #263: Stop using lambdas for instantiation and make compile-time choices instead

sras:
## Description

Problem: We have resorted to deriving RPC instances for some type due to
a bug in morley. Now the bug is fixed, and we can switch back to
auto derivation.

Solution: Update morley version to the one with the fix and switch to
auto derivation in the concerned types.

## Related issue(s)

Resolves #263 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


rinn7e (approved):
Looks good to me 👍

PR #330, opened 2022-06-16
[#329] Implement LambdaDAO

### LINKED ISSUE #329: Implement LambdaDAO and its registry/treasury variants along with tests.

Implement LambdaDAO, its registry/ treasury variants along with the tests.

# Acceptance criteria

1. LambdaDAO is implemented as per the spec
2. Tests for spec behaviour is implemented.
3. Registry/Treasury logic is implemented using LambdaDAO.

sras:
## Description

* Implemented core LambdaDAO functionality as a variant. 
* Made changes to the Makefile to build LambdaDAO and its storage.
* Added tests to check the core LambdaDAO behavior.
* Implemented pre-loading of Handlers
* Implement registry/treasury via LambdaDAO variant and use existing registry/treasury tests for testing their behavior.

### Deviations from the spec

* In the spec, the `handler_check` procedure accept `proposal_params` and `handler_storage` as input
but here in the implementation I have only used `bytes` that represent the actual argument to the target handler, instead of the whole `proposal_params`. This is so that we can additional unpacking and pattern matches in the procedure.

* Proposal handler input aka `ph_input` needed to include `proposal_info` field of type `propose_params` so that the handler can
compute proposal_key. This was required in the registryDAO implementation using LambdaDAO to properly implement the
'registry_affected' procedure to compute the proposal_key of the proposal that affected a certain key.

* Proposal handler output aka `ph_output` needed to include a guardian field of type `address option` to allow the updation of contract guardian from within the handlers. 

## Related issue(s)
Resolves #329

- Tests
        silently reappearing again.

- Documentation
    - Haddock


sras:
@pasqu4le I think this is now ready for initial reviews.

sras:
I can't really figure how the items found by `weeder` are actually un-used in this pipeline https://buildkite.com/serokell/basedao/builds/2887

rinn7e:
@sras Which ligo version are you using currently? I'm currently on the latest version (`0.46.1`), and couldn't really compile the contract. Would be nice if we update the ligo version as well.

src/variants/lambda/implementation.mligo
```diff
@@ -0,0 +1,101 @@
...
+          | Some hinfo ->
+              let
+                hc_input = (ehp.packed_argument, ce.handler_storage)
+              in hinfo.handler_check hc_input
+          | None -> (failwith proposal_handler_not_found: unit))
+
+let add_handler(ce, ahp : contract_extra * add_handler_param) : contract_extra =
+  if handler_exists(ahp.name, ce)
+    then
+     (failwith proposal_handler_exists: contract_extra)
```
rinn7e, line 44: Is this not possible to ensure that the handler name is unique during `proposal_check` so that we don't have to check it again in `flush`?

I imagine that if a user submit a proposal with the same handler name to another proposal (whether or not a proposal is ongoing or already completed), we should fail right away and not having to wait until `flush`).

I think we can keep track of `ongoing` proposal's handler name in contract extra, so that we can check it right away in `proposal_check`. Though, this solution require changing `proposal_check` to be able to modify `contract_extra` (adding new handler name if `Add_handler` succeed), not sure if this is worth the effort.

@sras What do you think about this?

rinn7e (changes requested):

src/variants/lambdaregistry/implementation.mligo
```diff
@@ -0,0 +1,160 @@
...
+  in ((operation :: nil_op), store)
+
+// -----------------------------------------------------------------
+// Custom entrypoint handler for this variant
+//
+// -----------------------------------------------------------------
+let custom_ep (custom_ep_param, storage : custom_ep_param * storage): operation list * storage
+  = match custom_ep_param with
+  | Lookup_registry p -> lookup_registry(p, storage)
+  | _ -> (([]: operation list), storage)
```
rinn7e, line 157: From what I understand, we need to have different variant to support the custom entrypoints. This is probably not intended since if the user want custom entrypoints, he/she should use compile version of BaseDAO instead.

I think either we don't support custom entrypoints in LambdaDAO for now or find another way to support custom entrypoints in a dynamic way (user may be able to create/remove) later on.

sras: Yes, LambdaDAO itself does not have a way to accomodate custom entrypoints. But LambdaDAO, being a variant itself, can be modified to have any custom entrypoint that is required.

src/variants/lambda/implementation.mligo
sras: > think we can keep track of ongoing proposal's handler name in contract extra, so that we can check it right away in proposal_check.

But if I understand your suggestion correctly, what should happen if this ongoing proposal fail? Then our rejection of the new proposal would turn out to be not-required, right?

Either way, I think it is not worth the effort and additional code and associated costs.

rinn7e: > But if I understand your suggestion correctly, what should happen if this ongoing proposal fail? Then our rejection of the new proposal would turn out to be not-required, right?

If the ongoing proposal fail, then that would be fine. The reject of the new proposal would make the proposer rename their handler, and proceed to propose again anyway.

If we wait till the proposal is already passed and fail the proposal just because of the name, that would waste the proposer and voters times.

@pasqu4le could give his view on this when he got back.

rinn7e (changes requested):

src/variants/lambdaregistry/implementation.mligo
rinn7e: > But LambdaDAO, being a variant itself

I suggest to not have a variant of LambdaDAO and simply not support custom entrypoints for now. 

The main reason is that, when using LambdaDAO, to have treasury or registry dao, one simply just update the handler without having to deploy another contract. 

If we have variant of LambdaDAO for treasury or registry, it seems to defeat the purpose, because then one have to deploy a separate contract.

@pasqu4le may have a different idea.

src/variants/lambda/implementation.mligo
```diff
@@ -0,0 +1,101 @@
...
+
+let unpack_proposal_metadata (pm: proposal_metadata) : lambda_dao_proposal_metadata =
+  match ((Bytes.unpack pm) : (lambda_dao_proposal_metadata option)) with
+  | Some (v) -> v
+  | None -> (failwith unpacking_proposal_metadata_failed : lambda_dao_proposal_metadata)
+
+let proposal_check (propose_params, ce : propose_params * contract_extra) : unit =
+  match unpack_proposal_metadata(propose_params.proposal_metadata) with
+    | Add_handler ahp -> if handler_exists(ahp.name, ce) then (failwith proposal_handler_exists : unit) else unit
+    | Remove_handler rhp -> if handler_exists(rhp, ce) then unit else (failwith proposal_handler_not_found : unit)
```
rinn7e, line 33: The `remove_handler` proposal should only succeed if the handler is active, so we should ensure if the handler is active as well.

src/variants/lambda/implementation.mligo
```diff
@@ -0,0 +1,101 @@
...
+
+let add_handler(ce, ahp : contract_extra * add_handler_param) : contract_extra =
+  if handler_exists(ahp.name, ce)
+    then
+     (failwith proposal_handler_exists: contract_extra)
+    else let
+      handler_info = { code = ahp.code ; handler_check = ahp.handler_check; is_active = true }
+      in {ce with lambdas = Big_map.add ahp.name handler_info ce.lambdas }
+
+let remove_handler(ce, rhp : contract_extra * remove_handler_param) : contract_extra = {ce with lambdas = Big_map.remove rhp ce.lambdas }
```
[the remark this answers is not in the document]
sras: Yea, I think so. Good catch here as well..

haskell/test/Test/Ligo/LambdaDAO.hs
```diff
@@ -0,0 +1,255 @@
...
+  toField #piHandler_storage #
+  L.nil @Operation #
+  L.swap #
+  L.none #
+  constructStack @PhOutput
+
+-- | We test non-token entrypoints of the BaseDAO contract here
+test_LambdaDAO :: [TestTree]
+test_LambdaDAO =
+  [ testGroup "LambdaDAO Tests"
```
rinn7e, line 103: I think we should also add test cases for when:
- `remove_handler` call succeed
- After `remove_handler` call, the proposal that has not been flush should be still allow to execute the handler
- After `remove_handler` call, a new `execute_handler` proposal shouldn't be allowed to use that handler.

rinn7e (changes requested):

src/variants/lambdaregistry/implementation.mligo
sras: Do you mean that we should not include the "lookup_registry" custom entrypoint for registry variant of LambdaDAO? If yes, do you have a suggestion on how to include it?

src/variants/lambdaregistry/implementation.mligo
```diff
@@ -0,0 +1,160 @@
...
+// operations.
+// -----------------------------------------------------------------
+
+let decision_callback (input : decision_callback_input)
+    : decision_callback_output = let
+  (ops, new_ce, new_guardian) = match unpack_proposal_metadata(input.proposal.metadata) with
+    | Add_handler ahp -> (([] : operation list), add_handler(input.extras, ahp), None)
+    | Remove_handler rhp -> (([] : operation list), remove_handler(input.extras, rhp), None)
+    | Execute_handler ehp -> execute_handler(input.proposal.proposer, input.proposal.proposer_frozen_token, input.extras, input.proposal.metadata, ehp)
+  in { operations = ops; extras = new_ce; guardian = new_guardian }
```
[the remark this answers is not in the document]
sras: Yea, there is some duplication. I think I made some efforts to extract them, but couldn't due to circular imports, which LIGO does not do a very good job of communicating (it just fails with an error that amounts to "Symbol not found"). 

Will give another try.

src/variants/lambdaregistry/implementation.mligo
rinn7e: Yep, I think we simply state we don't support custom entrypoint for now. 

If this is needed as it should be, we should figure out another way to create custom entrypoints later on, perhaps allowing the ability to create/remove custom entrypoints similarly to how the handler work. 

I guess, we may have to wait for Pasquale feedback.

sras: >Yep, I think we simply state we don't support custom entrypoint for now.

But we don't, and the 'custom' entrypoint for registry variant is not really dynamic. It is included because it is required to complete the registryDAO functionality.

src/variants/lambda/implementation.mligo
sras: Basically I think users should not forced to select a different name because there is another proposal that proposes to add a handler with same name. Becuase it is just a proposal, and so is the one that is being added. 

I think it makes sense to allow multiple proposals for the same thing, and let only one of them pass.

src/variants/lambda/implementation.mligo
sras: Oh, about this, I think as per spec, we planned `remove_handler` to do nothing if the handler is already de-activated, so I think this validation is not required.

rinn7e: Hmm I think we do nothing when in `decision_lambda`, but in `proposal_check` it mentions that it check for `active` handler:

image

So I think the validation is still required in this case.

sras: You are right. I got confused there for a minute.

src/variants/lambda/implementation.mligo
```diff
@@ -0,0 +1,101 @@
...
+
+let proposal_check (propose_params, ce : propose_params * contract_extra) : unit =
+  match unpack_proposal_metadata(propose_params.proposal_metadata) with
+    | Add_handler ahp -> if handler_exists(ahp.name, ce) then (failwith proposal_handler_exists : unit) else unit
+    | Remove_handler rhp -> if handler_exists(rhp, ce) then unit else (failwith proposal_handler_not_found : unit)
+    | Execute_handler ehp -> (match Big_map.find_opt ehp.handler_name ce.lambdas with
+          | Some hinfo ->
+              let
+                hc_input = (ehp.packed_argument, ce.handler_storage)
+              in hinfo.handler_check hc_input
```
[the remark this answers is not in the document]
sras: Have added this check and added a test for this behavior.

src/variants/lambda/implementation.mligo
sras: Have added this and a test.

src/variants/lambdaregistry/implementation.mligo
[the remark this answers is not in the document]
sras: Have extracted them out into a `variants/lambda/common.mligo` module.

rinn7e:
Looks good to me, beside the 2 points that need some more discussion, probably will have to wait for Pasquale's feedback after he gets back.

haskell/test/Test/Ligo/Common.hs
```diff
@@ -0,0 +1,73 @@
+-- SPDX-FileCopyrightText: 2021 Tezos Commons
+-- SPDX-License-Identifier: LicenseRef-MIT-TC
+--
+module Test.Ligo.Common
+  ( IsProposalArgument(..)
+  , VariantExtraHasField(..)
+  , TestableVariant(..)
```
rinn7e, line 7:
```suggestion
  , TestableVariant(toProposalMetadata, getVariantStorageRPC)
```
This probably fixes the weeder, can you try it.

rinn7e (changes requested):

haskell/test/Test/Ligo/Common.hs
rinn7e: On the side note, is it just me or the test compilation is noticeably slower? Especially during `RegistryDAO.Tests.**` tests

sras: It is not you. The compilation is now a lot slower due to the constraints added to the registry/treasury tests to use them for testing Lambda versions of the same.

sras: >This probably fixes the weeder, can you try it?

It does not appear possible to only export those methods, because the other ones are used in other modules.

sras: >This probably fixes the weeder, can you try it?

It does not appear possible to only export those methods, because the other ones are used in other modules.

rinn7e: > It does not appear possible to only export those methods, because the other ones are used in other modules.

Right, I guess we can update weeder.yaml to ignore it, similarly to the one in `morley`: https://gitlab.com/morley-framework/morley/-/blob/master/.weeder.yaml#L23

> It is not you. The compilation is now a lot slower due to the constraints added to the registry/treasury tests to use them for testing Lambda versions of the same.

I think this be a huge problem, we should try to find way to reduce the compilation time.

sras: > I guess we can update weeder.yaml to ignore it,

Sure, but I would like to figure why is those considered as weeds first..

I had tried to address it by splitting the registryDAO tests into many different modules, each with a single test. But it only helps with incremental compilation, and not build-from-scratch that CI does.

Another alternative I can think of is to duplicate the entire test modules for registry and treasury. Maybe Haskell has got parameteric modules, we could have made good use of that here...

src/variants/lambda/implementation.mligo
pasqu4le: > I think users should not forced to select a different name because there is another proposal that proposes to add a handler with same name. Becuase it is just a proposal, and so is the one that is being added.

I agree, it's only at `flush` that we know if a proposal has been accepted and so that another `handler` with the same name is no longer allowed.

> If we wait till the proposal is already passed and fail the proposal just because of the name, that would waste the proposer and voters times.

That's a valid concern, but a lot of other inefficiencies are saved this way. For example I might want to propose two competing handlers, in which case I could submit both at (roughly) the same time and not have to wait another voting cycle. Likewise this is true even if multiple parties want to propose competing handlers, they don't have to wait until a proposal has been `flush`ed to be able to propose themselves.

src/variants/lambdaregistry/implementation.mligo
pasqu4le: > I suggest to not have a variant of LambdaDAO and simply not support custom entrypoints for now.

I think that's not possible, one requirement for this variant is the ability to define the other existing ones (`registry` and `treasury`) within it, and since they have some custom entrypoints they are needed here too.

> If we have variant of LambdaDAO for treasury or registry, it seems to defeat the purpose, because then one have to deploy a separate contract.

I think that the point is to be able to tweak the behavior of a DAO by changing the available handlers, but not change it drastically, so choosing the (fixed) custom entrypoint once should be fine.

> one simply just update the handler without having to deploy another contract.

those are two different things however, custom entrypoints can be called directly, whereas the available handlers can only be used for proposals (following the same cycle of voting and `flush`ing as every other proposal).

src/variants/lambda/common.mligo
```diff
@@ -0,0 +1,67 @@
...
+#include "types.mligo"
+#include "../../types.mligo"
+
+let handler_exists(handler_name, ce : string * contract_extra) : bool =
+  Big_map.mem handler_name ce.lambdas
+
+let handler_exists_and_active(handler_name, ce : string * contract_extra) : bool =
+  match Big_map.find_opt handler_name ce.lambdas with
+    | Some (v) -> v.is_active
+    | None -> (failwith proposal_handler_not_found: bool)
```
pasqu4le, line 18: Why not return `false` in this case?

AFAICT this function is only used in `common_proposal_check` where a `false` would end up producing exactly the same result (a `failwith proposal_handler_not_found`).

Moreover, I think that it would be nice to add a documentation comment on this function, as well as the other helpers in this file, to specify what they are supposed to do.

src/variants/lambda/common.mligo
```diff
@@ -0,0 +1,67 @@
...
+    then
+     (failwith proposal_handler_exists: contract_extra)
+    else let
+      handler_info = { code = ahp.code ; handler_check = ahp.handler_check; is_active = true }
+      in {ce with lambdas = Big_map.add ahp.name handler_info ce.lambdas }
+
+let disable_handler(ce, rhp : contract_extra * remove_handler_param) : contract_extra =
+  match Big_map.find_opt rhp ce.lambdas with
+    | Some hinfo -> {ce with lambdas = Big_map.update rhp (Some {hinfo with is_active = false}) ce.lambdas }
+    | None -> ce
```
pasqu4le, line 44: I need to mention again that the documentation is quite lacking.

Here for example one might think that this could be a mistake. It isn't, because we decided for this logic, but it would help to have a comment here for the future as well.

More importantly, there is no document for `lambdaDAO` in the `docs` folder, nor existing docs have been updated to mention it. The only existing documentation right now is the draft spec used to implement this, but the usual docs should be added to this repo as well.

Please update the comments and documentation for this PR.

pasqu4le (commented):
@rinn7e I see that there's a lot of comments from you that are still not marked as resolved, could you please get back to them and see if they still haven't been addressed?

@sras could you please clean up the commit history as well (perhaps after @rinn7e has checked his comments)?

src/variants/template/implementation.mligo
```diff
@@ -47,7 +47,7 @@ type custom_ep_param = unit
 // Custom entrypoint handler for this variant
 //
 // -----------------------------------------------------------------
-let custom_ep (_, storage, _ : custom_ep_param * storage * config): operation list * storage
+let custom_ep (_, storage : custom_ep_param * storage): operation list * storage
```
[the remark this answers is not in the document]
sras: AFAIU we have bundled `config` with `storage` in this PR https://github.com/tezos-commons/baseDAO/pull/326, so I thought this was a leftover.

src/variants/lambda/common.mligo
sras: I have added the doc for lambdaDAO and added some comments.

sras:
I have cleaned up the commit history a bit.

src/variants/lambda/common.mligo
sras: >Why not return false in this case?

Seems right. Have changed. Also added some comments.

src/common.mligo
```diff
@@ -9,7 +9,7 @@
 
 [@inline]
 let authorize_admin (store : storage): storage =
-  if sender = store.admin then
+  if (Tezos.get_sender unit) = store.admin then
```
pasqu4le, line 12: Nitpick, but parenthesis should not be needed here:
```suggestion
  if Tezos.get_sender unit = store.admin then
```

nix/sources.json
```diff
@@ -17,17 +17,17 @@
...
-        "rev": "a4e105dcdb6d2b9d0627b8f0b893cec80485c14e",
-        "sha256": "144cfklhmqrq1kmvnmvcfmn819ijy4hr20whfazypxa72lcs3lz0",
+        "rev": "c427aeda2bc82fe4f8ff9354e70bdb4ecb5121b2",
+        "sha256": "1l2n46n5ym42ksyk7mvsh1a89ccdf4kc6k3vn4nvf22rvwm5v0w5",
         "type": "tarball",
-        "url": "https://github.com/serokell/haskell.nix/archive/a4e105dcdb6d2b9d0627b8f0b893cec80485c14e.tar.gz",
+        "url": "https://github.com/serokell/haskell.nix/archive/c427aeda2bc82fe4f8ff9354e70bdb4ecb5121b2.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "ligo": {
```
pasqu4le, line 26: AFAICT you updated to version `0.46.0` however the `docs/building.md` hasn't been updated to mention that this is the latest version used.

docs/lambda.md
```diff
@@ -0,0 +1,121 @@
...
+# Lambda DAO
+
+LambdaDAO allows dynamic configuration of proposals and proposal handlers,
+via proposals. This allows amending the logic of handling different kinds of
+proposals, or even adding or removing proposal types.
+
+LambdaDAO contains three concrete proposal types, which are "add handler", "remove handler"
+and "execute handler". The "add handler" proposal type can be used to defined a new
+proposal handler and "execute handler" proposal type can be used to invoke an existing
+proposal handler.
```
pasqu4le, line 15: A bit odd that there is no explanation of "remove handler", however TBH I would remove this sentence entirely and instead say that they are described below.

src/variants/lambdatreasury/storage.mligo
```diff
@@ -0,0 +1,185 @@
...
+#include "error_codes.mligo"
+
+// Xtz transfer amount cannot be 0
+let zero_mutez_err_msg = "ZERO_MUTEZ"
+
+// Xtz transfer amount cannot be smaller than `min_xtz_amount`
+let too_small_xtz_err_msg = "LOW_XTZ"
+
+// Xtz transfer amount cannot be bigger than `max_xtz_amount`
+let too_large_xtz_err_msg = "HIGH_XTZ"
```
pasqu4le, line 17: Why not import these from `src/common/errors.mligo`?

Likewise in `lambdaregistry`

src/variants/lambdatreasury/storage.mligo
```diff
@@ -0,0 +1,185 @@
...
+
+let fail_proposal_check_ (msg : string) : unit =
+  ([%Michelson ({| { FAILWITH } |} : (nat * string) -> unit)]
+    (fail_proposal_check, msg) : unit)
```
pasqu4le, line 21: This is also duplicated in `lambdaregistry`, can it not be extracted in a common module? :thinking:

src/variants/lambdatreasury/storage.mligo
```diff
@@ -0,0 +1,185 @@
...
+type token_transfer =
+  [@layout:comb]
+  { contract_address : address
+  ; transfer_list : transfer_item list
+  }
+
+type transfer_type =
+  [@layout:comb]
+  | Xtz_transfer_type of xtz_transfer
+  | Token_transfer_type of token_transfer
```
pasqu4le, line 60: These types are also the same between the two lambda variants, can they be moved to avoid duplication?

src/variants/lambdatreasury/storage.mligo
```diff
@@ -0,0 +1,185 @@
...
+          (Tezos.transaction tt.transfer_list 0mutez contract) :: ops
+      | None -> (failwith fail_decision_callback : operation list)
+    end
+  | Xtz_transfer_type xt ->
+    begin
+      match (Tezos.get_contract_opt xt.recipient : unit contract option) with
+      | Some contract ->
+          (Tezos.transaction unit xt.amount contract) :: ops
+      | None -> (failwith fail_decision_callback : operation list)
+    end
```
pasqu4le, line 92: nitpick, but AFAIK when all that's contained in a `match` is another `match` then `begin`/`end` are not necessary:
```suggestion    match (Tezos.get_entrypoint_opt "%transfer" tt.contract_address
        : transfer_params contract option) with
    | Some contract ->
        (Tezos.transaction tt.transfer_list 0mutez contract) :: ops
    | None -> (failwith fail_decision_callback : operation list)
  | Xtz_transfer_type xt ->
    match (Tezos.get_contract_opt xt.recipient : unit contract option) with
    | Some contract ->
        (Tezos.transaction unit xt.amount contract) :: ops
    | None -> (failwith fail_decision_callback : operation list)
```

I think it looks a bit more clean, but since it's a matter of taste I'll let you decide.

P.s. if you go with it there are also other places where this can be applied.

src/variants/lambdatreasury/implementation.mligo
```diff
@@ -0,0 +1,78 @@
...
+#include "helper/unpack.mligo"
+#include "variants/lambda/types.mligo"
+#include "variants/lambda/common.mligo"
+#include "types.mligo"
+
+// Incorrect token amounts locked
+let wrong_token_amount_err_msg = "WRONG_TOKEN_AMOUNT"
+
+// Proposal size is bigger than `max_proposal_size`
+let large_proposal_err_msg = "LARGE_PROPOSAL"
```
pasqu4le, line 17: Likewise, both here and in `lambdaregistry`, why not import these from `src/common/errors.mligo`?

src/variants/lambda/common.mligo
```diff
@@ -0,0 +1,83 @@
...
+let handler_exists_and_active(handler_name, ce : string * contract_extra) : bool =
+  match Big_map.find_opt handler_name ce.lambdas with
+    | Some (v) -> v.is_active
+    | None -> false
+
+// The proposal check for lambdaDAO.
+// 1. Add_handler proposal gets checked to ensure that no proposal handler with the input name exists already.
+// 2. Remove handler proposal gets checked to ensure that an active proposal handler with the given name exist.
+// 3. Execute_handler proposals gets checked to ensure that an active proposal handler with the given name exists, and
+//    then calls that handlers "handler_check" procedure.
```
pasqu4le, line 26:
```suggestion
//    then calls that handler's "handler_check" procedure.
```

src/variants/lambda/common.mligo
```diff
@@ -0,0 +1,83 @@
...
+    | Some hinfo -> {ce with lambdas = Big_map.update rhp (Some {hinfo with is_active = false}) ce.lambdas }
+    // If it does not, do nothing.
+    | None -> ce
+
+let execute_handler(proposer, frozen, ce, md, ehp : address * nat * contract_extra * proposal_metadata * execute_handler_param) : ((operation list) * contract_extra * (address option)) =
+  match Big_map.find_opt ehp.handler_name ce.lambdas with
+    | Some hinfo ->
+        // If handler is present, execute it. Note that we don't check for active status of the
+        // handler since we want inactive handlers to be executable as well, so that existing proposals
+        // that use the handler can proceed to completion. After execution, store the updated ja
```
pasqu4le, line 63:
```suggestion
        // that use the handler can proceed to completion. After execution, store the updated
```
it seems left over by accident, not sure if you meant something else.

src/variants/lambda/common.mligo
```diff
@@ -0,0 +1,83 @@
...
+// 1. Add_handler proposal gets checked to ensure that no proposal handler with the input name exists already.
+// 2. Remove handler proposal gets checked to ensure that an active proposal handler with the given name exist.
+// 3. Execute_handler proposals gets checked to ensure that an active proposal handler with the given name exists, and
+//    then calls that handlers "handler_check" procedure.
+let common_proposal_check (propose_params, ce : propose_params * contract_extra) : unit =
+  match unpack_proposal_metadata(propose_params.proposal_metadata) with
+    | Add_handler ahp -> if handler_exists(ahp.name, ce) then (failwith proposal_handler_exists : unit) else unit
+    | Remove_handler rhp -> if handler_exists_and_active(rhp, ce) then unit else (failwith proposal_handler_not_found : unit)
+    | Execute_handler ehp -> (match Big_map.find_opt ehp.handler_name ce.lambdas with
+          // Checks if the handler is active, and if it is, fetch its handler-check lambda
```
pasqu4le, line 32:
```suggestion
          // Check if the handler is active, and if it is, fetch its handler-check lambda
```

src/variants/lambda/common.mligo
```diff
@@ -0,0 +1,83 @@
...
+        // handler_storage, and return the emitted operations.
+        let
+          ph_input =
+            { packed_argument = ehp.packed_argument
+            ; handler_storage = ce.handler_storage
+            ; proposal_info = { from = proposer; frozen_token = frozen; proposal_metadata = md }
+            } in
+        let ph_out = hinfo.code ph_input
+        in (ph_out.operations, { ce with handler_storage = ph_out.handler_storage }, ph_out.guardian)
+    | None -> (failwith proposal_handler_not_found: ((operation list) * contract_extra * (address option)))
```
pasqu4le, line 73: If this happens doesn't that mean that there is a bug in the contract?

If so, I think it shouldn't be using the `proposal_handler_not_found` error IMO, as that's used for when a given parameter is incorrect.

docs/error-codes.md
```diff
@@ -39,6 +39,8 @@ Here is a summary of all the error codes thrown by the contract.
 | 121 | `fail_decision_callback` | Executing the proposal's decision callback results in failure. |
 | 123 | `unstake_invalid_proposal` | Cannot call `unstake_vote` on the proposal that is not flushed or dropped. |
 | 124 | `voter_does_not_exist` | The sender did not vote on the proposal or already unstaked tokens from the proposal. |
+| 125 | `proposal_handler_exists` | A proposal handler with the same name exists already |
+| 126 | `proposal_handler_not_found` | A proposal handler with the given name was not found |
```
pasqu4le, line 43: :thinking: should these be here?

AFAIK these are specific to the lambdaDAO variant (and the others based on it) whereas the error codes here are instead of baseDAO itself.

IOW shouldn't these be treated like `zero_mutez_err_msg` which is used by some variants but not all `baseDAO`-based contracts?

docs/lambda.md
```diff
@@ -0,0 +1,121 @@
...
+address must be an implicit account or refer to an entrypoint of the `unit`
+type.
+
+In addition to this, there are separate proposal check logic for "add handler",
+"remove_handler" and "execute_proposal" proposals, which are described below
+
+#### add handler
+
+The proposal check for "add handler" proposal checks there are no proposal with the same
+name as the one being added.
```
pasqu4le, line 70:
```suggestion
The proposal check for the "add handler" proposal checks there are no proposal
handlers with the same name as the one being added.
```

docs/lambda.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### remove handler
+
+The proposal check for "remove handler" proposal checks there is an active proposal with input
+name.
```
pasqu4le, line 75:
```suggestion
The proposal check for the "remove handler" proposal checks there is an active
proposal handler with the given name.
```

docs/lambda.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### execute handler
+
+The proposal check for "execute handler" proposal checks there is an active proposal with input
+name. It also fetches the "handler check" procedure for the handler, and runs it with the input
+from this proposal.
```
pasqu4le, line 81:
```suggestion
The proposal check for the "execute handler" proposal checks there is an active
proposal handler with the given name.
It also fetches the "handler check" procedure for the handler, and runs it with
the input from this proposal.
```

haskell/baseDAO-ligo-meta.cabal
```diff
@@ -32,6 +32,7 @@ library
   exposed-modules:
```
pasqu4le, line 32: It looks like this file as well as `package.yaml` are missing some updates in the `extra-source-files`.

haskell/test/Test/Ligo/LambdaDAO.hs
```diff
@@ -0,0 +1,386 @@
...
+             withSender admin $
+               call baseDao (Call @"Flush") (1 :: Natural)
+
+             -- Raise a proposal to add the duplicate handler
+             advanceToLevel (startLevel + 4 * toPeriod fs)
+             let proposeParams' = ProposeParams wallet1 proposalSize proposalMeta
+             withSender wallet1 $ call baseDao (Call @"Propose") proposeParams'
+               & expectFailedWith proposalHandlerExists
+
+      , testScenario "RemoveHandler proposal check checks for existing active handler" $ scenario $
```
pasqu4le, line 200: I think that these should be two different tests, one that tests the failure when the handler does not exist and the other that tests it when it does exist but is no longer active.

Same for the equivalent "execute handler" test

src/variants/template/implementation.mligo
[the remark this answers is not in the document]
pasqu4le: Ah, right, makes sense then

src/variants/lambdatreasury/storage.mligo
sras: This appear to cause circular imports because `common/errors.mligo` import `types.mligo` and `types.mligo` has to import `storage.mligo` for the variant being built.

src/variants/lambdatreasury/storage.mligo
sras: >when all that's contained in a match is another match then begin/end are not necessary:

This does not appear to be true, and I get a parse error without `begin..end` wrapping.

haskell/test/Test/Ligo/LambdaDAO.hs
sras: I thought using storage preloaded with inactive handlers cover both cases. So the error won't be thrown if  contract starts accepting non-existing handlers, or starts to accept non-active handlers, making the test fail.

But for remove handler proposal, it seems that I used storage with no handlers at all. So that was indeed a mistake.

But I have now separated the tests. May be it ll be useful in some manner.

docs/lambda.md
sras: Have removed the section.

src/variants/lambdatreasury/storage.mligo
sras: Importing the existing common module seems to create circular imports, since the `types.mligo` module imports this module.

src/variants/lambdatreasury/storage.mligo
sras: Same as above, unable to import `common.mligo` module here because of circular imports.

src/variants/lambdatreasury/implementation.mligo
sras: Done

src/variants/lambda/common.mligo
sras: Have switched it to use `fail_decision_callback` error

pasqu4le (approved):
There is [one comment left](https://github.com/tezos-commons/baseDAO/pull/330#discussion_r943567588) from the previous review, but other than that everything LGTM.

Please handle that and squash the commits, I'm approving already.

docs/error-codes.md
sras: Have removed these errors from the global list and made them to throw with `fail_proposal_check`.

pasqu4le (approved):
LGTM now

rinn7e (approved):

PR #335, opened 2022-08-22
[#329] Add storage building instructions for LambdaDAO

sras:
## Description

Problem: We missed to add storage building instructions for LambdaDAO
and variants to 'building.md'.

Solution: Add the instructions to 'building.md'.

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,14 +240,14 @@ $(OUT)/lambdaDAO_storage.tz: src/**
 	$(BUILD_STORAGE) --output-file $(OUT)/lambdaDAO_storage.tz \
       src/base_DAO.mligo -e base_DAO_contract "default_storage( \
         { admin = (\"$(call require_defined,admin_address)\" : address) \
-				; guardian = (\"$(call require_defined,guardian_address)\" : address) \
-				; governance_token = \
-					{ address = (\"$(call require_defined,governance_token_address)\" : address) \
-					; token_id = ($(call require_defined,governance_token_id) : nat) \
-					} \
-				; start_level = {blocks = $(call require_defined,start_level)} \
-				; metadata_map = ($(call escape_double_quote,$(metadata_map)) : metadata_map) \
-				; freeze_history = ($(call escape_double_quote,$(freeze_history)) : freeze_history_list) \
+        ; guardian = (\"$(call require_defined,guardian_address)\" : address) \
+        ; governance_token = \
+          { address = (\"$(call require_defined,governance_token_address)\" : address) \
+          ; token_id = ($(call require_defined,governance_token_id) : nat) \
+          } \
+        ; start_level = {blocks = $(call require_defined,start_level)} \
+        ; metadata_map = ($(call escape_double_quote,$(metadata_map)) : metadata_map) \
+        ; freeze_history = ($(call escape_double_quote,$(freeze_history)) : freeze_history_list) \
         ; config_data = \
           { max_quorum = { numerator = (($(max_quorum) : nat) * quorum_denominator)/100n } \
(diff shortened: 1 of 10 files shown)


docs/building.md
```diff
@@ -233,6 +233,105 @@ The `admin_address`, `guardian_address`, `governance_token_address`,
 `start_level` and `governance_token_id` are required values. The rest of the
```
pasqu4le, line 233: I just noticed that there is a mistake in the `make` call example for `treasuryDAO`: the `freeze_history` option is specified twice, please fix that as well.

docs/building.md
```diff
@@ -233,6 +233,105 @@ The `admin_address`, `guardian_address`, `governance_token_address`,
...
+  guardian_address=KT1QbdJ7M7uAQZwLpvzerUyk7LYkJWDL7eDh \
+  governance_token_address=KT1RdwP8XJPjFyGoUsXFQnQo1yNm6gUqVdp5 \
+  governance_token_id=0n \
+  frozen_scale_value=1n \
+  frozen_extra_value=0n \
+  max_proposal_size=100n \
+  slash_scale_value=1n \
+  slash_division_value=1n \
+  min_xtz_amount=0mutez \
+  max_xtz_amount=100mutez \
```
pasqu4le, line 253:
```suggestion
```
these seem unnecessary

docs/building.md
```diff
@@ -233,6 +233,105 @@ The `admin_address`, `guardian_address`, `governance_token_address`,
...
```
pasqu4le, line 286: I believe that the should be used, but looking at the `Makefile` they don't seem to be? Is this a mistake?

docs/building.md
```diff
@@ -233,6 +233,105 @@ The `admin_address`, `guardian_address`, `governance_token_address`,
...
```
pasqu4le, line 319: Same question as for `LambdaRegistryDAO`

pasqu4le (changes requested):

docs/building.md
sras: I think it is a mistake. And it seems that tests need to be expanded to include this check for these variants.

sras: The configuration values passed from makefile did not appeared to be used even for original registry/treasury. Have fixed it now.

docs/building.md
sras: Fixed.

pasqu4le (approved):
One more nitpick, but the rest LGTM

Makefile
```diff
@@ -282,33 +289,48 @@ $(OUT)/lambdaregistryDAO_storage.tz: src/**
...
-          ; proposal_expired_level = { blocks = ($(proposal_expired_level) : nat) }\
-          ; fixed_proposal_fee_in_token = ($(fixed_proposal_fee_in_token) : nat) \
-          ; quorum_threshold = { numerator = (($(quorum_threshold) : nat) * quorum_denominator)/100n } \
-          ; quorum_change = { numerator = (($(quorum_change) : nat) * quorum_denominator)/100n } \
-          ; max_quorum_change = { numerator = (($(max_quorum_change) : nat) * quorum_denominator)/100n } \
-          ; governance_total_supply = ($(governance_total_supply) : nat) \
-          } \
+      src/base_DAO.mligo -e base_DAO_contract "default_lambda_registry_DAO_storage( \
+        { base_data = \
+					{ admin = (\"$(call require_defined,admin_address)\" : address) \
```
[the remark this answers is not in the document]
sras: Have used spaces instead of tabs for indenting new lines.

PR #341, opened 2022-10-08
Update morley dependencies and update CI to use kathmandu chains

sras:
## Description

Update deps of Morley and friends and include Kathmandu network in CI tests

## Related issue(s)

Resolves #

- Tests
        silently reappearing again.

- Documentation
    - Haddock


haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
```diff
@@ -44,24 +34,24 @@ test_FA2 :: TestTree
...
-    ]
-  ]
+mkFA2Tests _ _ = testGroup "FA2 off-chain views" []
+-- TODO: Uncomment when some way to run views is available.
+--   [ testGroup "governance_token" $
+--     [ testCase "Get the address and token_id of the associated FA2 contract" $
+--         runView @GovernanceToken (governanceTokenView mc) storage NoParam
+--         @?= (Right $ GovernanceToken (MkAddress genesisAddress) FA2.theTokenId)
+--     ]
+--   ]
```
lierdakil, line 57: Correct me if I'm wrong, but there's certainly a way to run this test without `interpretView`? Just convert it to a Cleveland test, originate the contract with this view and use `callOffChainView`?

sras: I gave this a try, and found myself having a hard time getting `MetadataMap` from `Metadata` (provided by `knownBaseDAOMetadata`)  to populate the views in `offchainViewStorage`.

lierdakil: Uh. So, here's the diff that _should_ work (notice modules are imported wherever):

```diff
diff --git a/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs b/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
index f273899..d0679eb 100644
--- a/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
@@ -1,11 +1,14 @@
 -- SPDX-FileCopyrightText: 2021 Tezos Commons
 -- SPDX-License-Identifier: LicenseRef-MIT-TC
 
+{-# LANGUAGE OverloadedLists #-}
+
 module Test.Ligo.BaseDAO.OffChainViews
   ( test_FA2
   ) where
 
 import Universum hiding (view)
+import Lorentz (Contract, defaultContract, cdr, nil, pair, (#), NiceStorage, HasAnnotation)
 
 import Morley.Michelson.Runtime.GState (genesisAddress)
 import Morley.Tezos.Address
@@ -14,13 +17,23 @@ import Test.Tasty (TestTree, testGroup)
 
 import Ligo.BaseDAO.Types
 import Ligo.BaseDAO.TZIP16Metadata
+import Lorentz.Contracts.Spec.FA2Interface qualified as FA2
+import Lorentz.Contracts.Spec.TZIP16Interface
+import Test.Morley.Metadata
+import Test.Cleveland
+import Data.Aeson (encode)
+import Data.ByteString.Lazy qualified as BS
+
+metadataBigMap :: MetadataConfig -> MetadataMap
+metadataBigMap conf = metadataURI (tezosStorageUri selfHost "metadata")
+  <> [("metadata", BS.toStrict $ encode $ knownBaseDAOMetadata $ mkMetadataSettings conf)]
 
 offChainViewStorage :: Storage
 offChainViewStorage =
   (mkStorage
   ! #admin addr
   ! #extra ()
-  ! #metadata mempty
+  ! #metadata (metadataBigMap defaultMetadataConfig)
   ! #level 100
   ! #tokenAddress (MkAddress genesisAddress)
   ! #quorumThreshold (mkQuorumThreshold 1 100)
@@ -31,27 +44,10 @@ offChainViewStorage =
     addr = [ta|tz1M6dcor9QNTFr9Ri68cBYvpxrogZaMttuE|]
 
 test_FA2 :: TestTree
-test_FA2 =
-  mkFA2Tests offChainViewStorage (mkMetadataSettings defaultMetadataConfig)
-
--- runView
---   :: forall ret mp. (HasCallStack, IsoValue ret)
---   => View $ ToT Storage
---   -> Storage
---   -> ViewParam mp
---   -> Either MichelsonFailed ret
--- runView view storage param =
---   case interpretView dummyContractEnv view param storage of
---     Right x -> Right x
---     Left (VIEMichelson _ (MSVIEMichelsonFailed (MichelsonFailureWithStack e _))) -> Left e
---     Left err -> error (pretty err)
-
-mkFA2Tests :: Storage -> MetadataSettings -> TestTree
-mkFA2Tests _ _ = testGroup "FA2 off-chain views" []
--- TODO: Uncomment when some way to run views is available.
---   [ testGroup "governance_token" $
---     [ testCase "Get the address and token_id of the associated FA2 contract" $
---         runView @GovernanceToken (governanceTokenView mc) storage NoParam
---         @?= (Right $ GovernanceToken (MkAddress genesisAddress) FA2.theTokenId)
---     ]
---   ]
+test_FA2 = testScenarioOnEmulator "FA2 off-chain views" $ scenario $ do
+  contract <- originate "test_FA2" offChainViewStorage (idContract @Storage)
+  callOffChainView @GovernanceToken contract "governance_token" NoParam
+    @@== (GovernanceToken (MkAddress genesisAddress) FA2.theTokenId)
+
+idContract :: (NiceStorage st, HasAnnotation st) => Contract () st ()
+idContract = defaultContract $ cdr # nil # pair

```

However, I've spent the last 45 minutes or so staring at GHC struggling to compile baseDAO tests, so I don't know if this works. Was it always this slow? I mean, the last build took a full hour on CI.

lierdakil:
Also, please update morley-infra, poor CI has been running since yesterday in a loop stuck with "unknown protocol".

haskell/test/Test/Ligo/BaseDAO/OffChainViews.hs
lierdakil:
```
*** Desugar [Test.Ligo.BaseDAO.Management.TransferOwnership]:
Result size of Desugar (before optimization)
  = {terms: 19,984,
     types: 168,417,
     coercions: 2,119,458,
     joins: 0/5,488}
Result size of Desugar (after optimization)
  = {terms: 16,309,
     types: 152,046,
     coercions: 2,137,815,
     joins: 10/3,905}
!!! Desugar [Test.Ligo.BaseDAO.Management.TransferOwnership]: finished in 35340.72 milliseconds, allocated 95255.164 megabytes
```

> coercions: 2,119,458,
> allocated 95255.164 megabytes

WTH

sras: My current implimentation is almost the same as your diff, but returns the following error for some reason. Trying your diff now..

```
governanceToken <- callOffChainView @GovernanceToken baseDao "governance_token" NoParam
                     ┃                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     ┃                              | Metadata bigmap does not contain the key: ''
```

sras: > Was it always this slow? I mean, the last build took a full hour on CI.

Yes, AFAIU it became this slow when we had to share tests for `registryDAO` between its native version and the `LambdaDAO` version. So these tests were changed to be polymorphic over storage and parameter. If you can think of a better way to solve this (that is reuse the tests between two contracts with slightly different parameter/storage), please let us know. Slow compilation has been such a pain for a while now..

sras: Worked. Thanks!

lierdakil: So... It's still slow as heck, but _apparently_ GHC really dislikes all those partial matches everywhere. I've got rid of those using `SizedList`, and now it's still slow, but it got from "I'll die sooner than this finishes" to "I hate it, but I can live with it" slow. To put some specific numbers to that, with `--ghc-options=-j`:

```
real    5m31,311s
user    19m36,553s
sys     3m49,918s
```

I'll push the changes I made to some branch here if I have commit rights.

sras: So the slowdown was not due to the large number of constraints on the polymorphic tests? That is comforting, but please, when you have a minute, can you share how you figured it?

>I'll push the changes I made to some branch here if I have commit rights.

For the timebeing, can you please send me a patch from `git` that I can import?

lierdakil: > can you share how you figured it?

Eh. How I _could_ have figured it out is by looking at the core output. How I _did_... Well, a lucky educated guess, frankly. Since running the compiler with `-v2` shows it gets very stuck on the desugarer phase, and reports an ungodly amount of memory allocations, this suggests it generates _huge_ core output. I vaguely remembered that partial pattern matches are supposed to turn into explicit errors in the desugarer, so I suspected there might be some issues with that. Getting rid of partial pattern matches was the most obvious thing to try at that point, really. TBH, it's mostly guesswork until I tried and seen some improvement.

Apparently I do have commit rights, so I've pushed `lierdakil/try-use-sized-list`

lierdakil: Fair warning: the code I pushed is messy, it needs a bit of clean-up.

sras: The compile time has dropped from 30 minutes to 2 minutes 40 secs for a recompiling to account for change in a single module. I cannot figure how this slowdown could be caused by partial pattern matches alone. Is it possible that the constraints somehow play a role in this? becuase IIRC introducing them triggered the slow down, when the partial patterns were there from the begining ..

Anyway this change takes away some massive pain, so a big thanks for figuring this out.

lierdakil:
Certainly it's not partial pattern matches alone, but I don't know enough
compiler internals to really say what else is at play. I vaguely suspect
polymorphic code might have exponentially increased the amount of work the
desugarer has to do wrt partial pattern matches or vice versa, but I might
be totally wrong, and I couldn't tell you the exact details even if I'm
right. I'm pretty convinced this is an unfortunate interaction of multiple
subsystems. Staring at the core output for a few hours might offer more
insight, but that's not something I'm very eager to do, core is very
verbose, and finding relevant tidbits takes a lot of effort.

пн, 17 окт. 2022 г., 02:29 Sandeep.C.R ***@***.***>:

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -100,18 +101,18 @@ makeProposalKey params = toHashHs $ lPackValue params
 
 addDataToSign
   :: (MonadCleveland caps m)
-  => TAddress param vd
+  => ContractHandle p s ()
   -> Nonce
   -> d
   -> m (DataToSign d, d)
-addDataToSign (toAddress -> dsContract) dsNonce dsData = do
+addDataToSign ((toAddress . chAddress) -> dsContract) dsNonce dsData = do
```
lierdakil, line 108:
```suggestion
addDataToSign (toAddress -> dsContract) dsNonce dsData = do
```

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -122,32 +123,32 @@ permitProtect author (toSign, a) = do
...
 sendXtz addr = withFrozenCallStack $ do
   let transferData = TransferData
-        { tdTo = toAddress addr
+        { tdTo = addr
         , tdAmount = [tz|0.5|] -- 0.5 xtz
         , tdEntrypoint = DefEpName
         , tdParameter = ()
         }
-  transfer transferData
+  void $ runTransfer transferData
```
lierdakil, line 134: Why not just

```haskell
transfer addr [tz|0.5|]
```
?

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -122,32 +123,32 @@ permitProtect author (toSign, a) = do
...
 sendXtzWithAmount amt addr = withFrozenCallStack $ do
   let transferData = TransferData
-        { tdTo = toAddress addr
+        { tdTo = addr
         , tdAmount = amt
         , tdEntrypoint = DefEpName
         , tdParameter = ()
         }
-  transfer transferData
+  void $ runTransfer transferData
```
lierdakil, line 146: Same here, `transfer addr amt`

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -156,30 +157,30 @@ originateLigoDaoWithConfig
...
+  operator2 <- refillable $ newAddress "operator2"
 
-  admin :: Address <- refillable $ newAddress "admin"
+  admin <- refillable $ newAddress "admin"
 
   owner1Balance <- getBalance owner1
 
   when (owner1Balance < 2_e6) $ do
-    let fundAddress x = transfer $ TransferData x [tz|5|] DefEpName ()
+    let fundAddress x = runTransfer $ TransferData x [tz|5|] DefEpName ()
```
lierdakil, line 170:
```suggestion
    let fundAddress x = transfer x [tz|5|]
```

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -156,30 +157,30 @@ originateLigoDaoWithConfig
...
-  operator1 :: Address <- refillable $ newAddress "operator1"
-  owner2 :: Address <- refillable $ newAddress "owner2"
-  operator2 :: Address <- refillable $ newAddress "operator2"
+  owner1 <- refillable $ newAddress "owner1"
+  operator1 <- refillable $ newAddress "operator1"
+  owner2 <- refillable $ newAddress "owner2"
+  operator2 <- refillable $ newAddress "operator2"
 
-  admin :: Address <- refillable $ newAddress "admin"
+  admin <- refillable $ newAddress "admin"
```
lierdakil, line 165: Might as well create these in bulk:

```haskell
owner1 ::< operator1 ::< owner2 ::< operator2 ::< admin ::< Nil' <- refillables $
  newAddresses $ "owner1" ::< "operator1" ::< "owner2" ::< "operator2" ::< "admin" ::< Nil'
```

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -156,30 +157,30 @@ originateLigoDaoWithConfig
...
+  (guardianContract :: ContractHandle (Address, ProposalKey) () ()) <- originate "guardian" () dummyGuardianContract
 
   let originationOffset = 12
   let storage =
             ( mkStorage
               ! #extra extra
               ! #admin admin
               ! #metadata mempty
-              ! #tokenAddress tokenContract
+              ! #tokenAddress (toAddress $ toContractAddress tokenContract)
```
lierdakil, line 183:
```suggestion
              ! #tokenAddress (toAddress tokenContract)
```

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -156,30 +157,30 @@ originateLigoDaoWithConfig
...
   when (owner1Balance < 2_e6) $ do
-    let fundAddress x = transfer $ TransferData x [tz|5|] DefEpName ()
+    let fundAddress x = runTransfer $ TransferData x [tz|5|] DefEpName ()
     mapM_ fundAddress [owner1, operator1, owner2, operator2, admin]
 
   currentLevel <- getLevel
-  tokenContract <- chAddress <$> originateSimple "TokenContract" [] dummyFA2Contract
-  guardianContract <- chAddress <$> originateSimple "guardian" () dummyGuardianContract
+  (tokenContract :: ContractHandle FA2.Parameter [FA2.TransferParams] ()) <- originate "TokenContract" [] dummyFA2Contract
+  (guardianContract :: ContractHandle (Address, ProposalKey) () ()) <- originate "guardian" () dummyGuardianContract
```
lierdakil, line 175: I don't think these signatures are required?

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -188,24 +189,14 @@ originateLigoDaoWithConfig extra config qt = do
               ! #quorumThreshold qt
               ! #config config
             )
-            { sGuardian = guardianContract
+            { sGuardian = (toAddress $ toContractAddress guardianContract)
```
lierdakil, line 192:
```suggestion
            { sGuardian = toAddress guardianContract
```

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -252,10 +243,10 @@ createSampleProposals (counter1, counter2) dodOwner1 dao = do
...
 
 checkStorage
-  :: forall st caps m.
+  :: forall st caps m p.
       ( AsRPC st ~ st, MonadCleveland caps m
       , HasCallStack, Eq st, NiceStorage st)
-  => Address -> st -> m ()
+  => ContractHandle p st () -> st -> m ()
 checkStorage addr expected = do
   realSt <- getStorage @st addr
```
lierdakil, line 251:
```suggestion
  realSt <- getStorage addr
```

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -23,95 +23,96 @@ module Test.Ligo.BaseDAO.Common.StorageHelper
...
 import Universum
 
+import Morley.Tezos.Address
 import Test.Cleveland
 
 import Ligo.BaseDAO.Types
 
-getStorageRPC' :: forall cep p vd caps m. (HasCallStack, CEConstraints cep, MonadCleveland caps m) => TAddress p vd ->  m (StorageSkeletonRPC (VariantToExtra cep))
-getStorageRPC' addr = getStorage @(StorageSkeleton (VariantToExtra cep)) (unTAddress addr)
+getStorageRPC' :: forall cep caps m p s. (HasCallStack, CEConstraints cep, MonadCleveland caps m, NiceStorage (VariantToExtra cep)) => ContractHandle p s () ->  m (StorageSkeletonRPC (VariantToExtra cep))
```
lierdakil, line 31:
```suggestion
getStorageRPC' 
  :: forall cep caps m p s.
     ( HasCallStack, CEConstraints cep, MonadCleveland caps m
     , NiceStorage (VariantToExtra cep)
     ) 
  => ContractHandle p s () -> m (StorageSkeletonRPC (VariantToExtra cep))
```

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -23,95 +23,96 @@ module Test.Ligo.BaseDAO.Common.StorageHelper
...
-getFrozenTotalSupply' :: forall cep p vd caps m. (CEConstraints cep, MonadCleveland caps m) => TAddress p vd -> m Natural
+getFrozenTotalSupply' :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m) => ContractHandle p s () -> m Natural
 getFrozenTotalSupply' addr = sFrozenTotalSupplyRPC <$> (getStorageRPC' @cep addr)
 
-getFrozenTotalSupply :: forall p vd caps m. MonadCleveland caps m => TAddress p vd -> m Natural
+getFrozenTotalSupply :: forall p s caps m. MonadCleveland caps m => ContractHandle p s () -> m Natural
 getFrozenTotalSupply addr = getFrozenTotalSupply' @'Base addr
 
-getFreezeHistory' :: forall cep p vd caps m. (CEConstraints cep, MonadCleveland caps m) => TAddress p vd -> Address -> m (Maybe AddressFreezeHistory)
+getFreezeHistory' :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m) => ContractHandle p s () -> ImplicitAddress -> m (Maybe AddressFreezeHistory)
```
lierdakil, line 43:
```suggestion
getFreezeHistory'
  :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m)
  => ContractHandle p s () -> ImplicitAddress -> m (Maybe AddressFreezeHistory)
```

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -23,95 +23,96 @@ module Test.Ligo.BaseDAO.Common.StorageHelper
...
 getFreezeHistory' addr owner = do
   freezeHistoryBmId <- sFreezeHistoryRPC <$> (getStorageRPC' @cep addr)
-  getBigMapValueMaybe freezeHistoryBmId owner
+  getBigMapValueMaybe freezeHistoryBmId (toAddress owner)
 
-getFreezeHistory :: forall p vd caps m. (MonadCleveland caps m) => TAddress p vd -> Address -> m (Maybe AddressFreezeHistory)
+getFreezeHistory :: forall p s caps m. (MonadCleveland caps m) => ContractHandle p s () -> ImplicitAddress -> m (Maybe AddressFreezeHistory)
```
lierdakil, line 48:
```suggestion
getFreezeHistory
  :: forall p s caps m. (MonadCleveland caps m)
  => ContractHandle p s () -> ImplicitAddress -> m (Maybe AddressFreezeHistory)
```

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -23,95 +23,96 @@ module Test.Ligo.BaseDAO.Common.StorageHelper
...
 getFreezeHistory addr owner = getFreezeHistory' @'Base addr owner
 
-getQtAtCycle' :: forall cep p vd caps m. (CEConstraints cep, MonadCleveland caps m) => TAddress p vd -> m QuorumThresholdAtCycle
+getQtAtCycle' :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m) => ContractHandle p s () -> m QuorumThresholdAtCycle
```
lierdakil, line 51:
```suggestion
getQtAtCycle'
  :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m)
  => ContractHandle p s () -> m QuorumThresholdAtCycle
```

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -23,95 +23,96 @@ module Test.Ligo.BaseDAO.Common.StorageHelper
...
-  -> (Address, ProposalKey)
+  :: forall p s caps m. MonadCleveland caps m
+  => ContractHandle p s ()
+  -> (ImplicitAddress, ProposalKey)
   -> m (Maybe StakedVote)
-getVoter addr key = do
+getVoter addr (a, pk) = do
   bId <- sStakedVotesRPC <$> getStorageRPC addr
-  getBigMapValueMaybe bId key
+  getBigMapValueMaybe bId (toAddress a, pk)
```
lierdakil, line 81: Why not
```haskell
getVoter addr key = do
  bId <- sStakedVotesRPC <$> getStorageRPC addr
  getBigMapValueMaybe bId $ first toAddress key
```
?

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -137,45 +138,45 @@ checkIfAProposalExist' proposalKey dodDao expected = do
...
 
 checkIfAProposalExist
-  :: forall p vd caps m. MonadCleveland caps m
-  => ProposalKey -> TAddress p vd -> Bool -> m ()
+  :: forall p s caps m. MonadCleveland caps m
+  => ProposalKey -> ContractHandle p s () -> Bool -> m ()
 checkIfAProposalExist = checkIfAProposalExist' @'Base
 
-checkGuardian' :: forall cep p vd caps m. (CEConstraints cep, MonadCleveland caps m) => TAddress p vd -> Address -> m ()
+checkGuardian' :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m) => ContractHandle p s () -> ImplicitAddress -> m ()
```
lierdakil, line 145:
```suggestion
checkGuardian'
  :: forall cep p s caps m. (CEConstraints cep, MonadCleveland caps m)
  => ContractHandle p s () -> ImplicitAddress -> m ()
```

haskell/test/Test/Ligo/BaseDAO/Common/StorageHelper.hs
```diff
@@ -137,45 +138,45 @@ checkIfAProposalExist' proposalKey dodDao expected = do
...
 checkGuardian' addr guardianToChk = do
   actual <- sGuardianRPC <$> (getStorageRPC' @cep addr)
-  actual @== guardianToChk
+  actual @== (toAddress guardianToChk)
```
lierdakil, line 148:
```suggestion
  actual @== toAddress guardianToChk
```

haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
```diff
@@ -23,136 +20,134 @@ import Universum
...
-  => WithOriginateFn m -> WithStorage -> m ()
+  => WithOriginateFn 2 m -> WithStorage -> m ()
 transferOwnershipSetsPendingOwner withOriginatedFn initialStorage =
-  withOriginatedFn 2 (\(owner:_) -> initialStorage owner) $ \[owner, wallet1] baseDao -> do
-    withSender owner $ call baseDao (Call @"Transfer_ownership") (#newOwner :! wallet1)
+  withOriginatedFn (\(owner::<_) -> initialStorage owner) $ \(owner ::< wallet1 ::< Nil') baseDao -> do
+    withSender owner $ (transfer baseDao $ calling (ep @"Transfer_ownership") (#newOwner :! (toAddress wallet1)))
     mNewPendingOwner <- sPendingOwnerRPC <$> getStorageRPC baseDao
-    assert (mNewPendingOwner == wallet1) "Pending owner was not set as expected"
+    assert (mNewPendingOwner == (toAddress wallet1)) "Pending owner was not set as expected"
```
lierdakil, line 52:
```suggestion
    assert (mNewPendingOwner == toAddress wallet1) "Pending owner was not set as expected"
```

haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
```diff
@@ -23,136 +20,134 @@ import Universum
...
-      withSender owner $ call baseDao (Call @"Transfer_ownership")
-        (#newOwner :! wallet1)
-      withSender wallet1 $ call baseDao (Call @"Accept_ownership") ()
+  withOriginatedFn (\(owner::<_) -> initialStorage owner) $
+    \(owner ::< wallet1 ::< Nil') baseDao -> do
+      withSender owner $ (transfer baseDao $ calling (ep @"Transfer_ownership") (#newOwner :! (toAddress wallet1)))
+      withSender wallet1 $ (transfer baseDao $ calling (ep @"Accept_ownership") ())
       administrator <- sAdminRPC <$> (getStorageRPC baseDao)
-      assert (administrator == wallet1) "Administrator was not set from pending owner"
+      assert (administrator == (toAddress wallet1)) "Administrator was not set from pending owner"
```
lierdakil, line 73:
```suggestion
      assert (administrator == toAddress wallet1) "Administrator was not set from pending owner"
```

haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
```diff
@@ -23,136 +20,134 @@ import Universum
...
-        call baseDao (Call @"Transfer_ownership")
-          (#newOwner :! wallet2)
+        (transfer baseDao $ calling (ep @"Transfer_ownership")
+          (#newOwner :! (toAddress wallet1)))
+        (transfer baseDao $ calling (ep @"Transfer_ownership")
+          (#newOwner :! (toAddress wallet2)))
         pure ()
       pendingOwner <- sPendingOwnerRPC <$> (getStorageRPC baseDao)
-      assert (pendingOwner == wallet2) "Pending owner from earlier call was not re-written"
+      assert (pendingOwner == (toAddress wallet2)) "Pending owner from earlier call was not re-written"
```
lierdakil, line 123:
```suggestion
      assert (pendingOwner == toAddress wallet2) "Pending owner from earlier call was not re-written"
```

You get the idea, there are a couple more here.

haskell/test/Test/Ligo/BaseDAO/Management/TransferOwnership.hs
```diff
@@ -23,136 +20,134 @@ import Universum
...
-    \[owner] baseDao -> do
+  withOriginatedFn (\(owner::<_) -> initialStorage owner) $
+    \(owner ::< Nil') baseDao -> do
       withSender owner $ do
-        call baseDao (Call @"Transfer_ownership")
-          (#newOwner :! (unTAddress baseDao))
-      currentAdmin <- sAdminRPC <$> getStorage @Storage (unTAddress baseDao)
-      assert (currentAdmin == (unTAddress baseDao)) "Admin address was not set"
+        (transfer baseDao $ calling (ep @"Transfer_ownership")
+          (#newOwner :! (toAddress $ chAddress baseDao)))
```
lierdakil, line 148:
```suggestion
          (#newOwner :! toAddress baseDao))
```
Anyway, it's probably easier to do search and replace on `toAddress $ chAddress`, so I won't mention it below.

haskell/test/Test/Ligo/BaseDAO/Proposal/Flush.hs
```diff
@@ -282,16 +281,17 @@ flushFailOnExpiredProposal originateFn = withFrozenCallStack $ do
   DaoOriginateData{..} <-
     originateFn
      (testConfig
-       >>- (ConfigDesc configConsts{ cmProposalFlushTime = Just 40 })
-       >>- (ConfigDesc configConsts{ cmProposalExpiredTime = Just 50 })
+       >>- (ConfigDesc configConsts{ cmProposalFlushTime = Just 80 })
+       >>- (ConfigDesc configConsts{ cmProposalExpiredTime = Just 130 })
+       >>- (ConfigDesc configConsts{ cmPeriod = Just 40 })
```
lierdakil, line 286: Why the change here?

haskell/test/Test/Ligo/BaseDAO/Proposal/Propose.hs
```diff
@@ -57,22 +57,22 @@ validProposal originateFn = do
...
   assert (storage ==
-    ([[FA2.TransferItem { tiFrom = dodOwner1, tiTxs = [FA2.TransferDestination { tdTo = unTAddress dodDao, tdTokenId = FA2.theTokenId, tdAmount = 10 }] }]]))
+    ([[FA2.TransferItem { tiFrom = toAddress dodOwner1, tiTxs = [FA2.TransferDestination { tdTo = toAddress $ chAddress dodDao, tdTokenId = FA2.theTokenId, tdAmount = 10 }] }]]))
     "Unexpected Transfers"
 
   -- Advance one voting period to a proposing stage.
   advanceToLevel (startLevel + dodPeriod)
 
-  withSender dodOwner1 $ call dodDao (Call @"Propose") params
+  withSender dodOwner1 $ transfer dodDao$ calling (ep @"Propose") params
```
lierdakil, line 75:
```suggestion
  withSender dodOwner1 $ transfer dodDao $ calling (ep @"Propose") params
```

lierdakil (commented):
Mostly code style and formatting nitpicks, but I really don't understand why the change to test config in `haskell/test/Test/Ligo/BaseDAO/Proposal/Flush.hs`

sras:
> I really don't understand why the change to test config in haskell/test/Test/Ligo/BaseDAO/Proposal/Flush.hs

That was to make the real network tests to pass in CI. So in this contract, there is some amount of time/level based logic, so sometimes when real network tests fail in CI, we have to tweak the configuration, like here we have increased the duration of the period, and proposal expiry time, so that certain procedures in the tests fall in the expected time/level frame. This is done so that couple of `freeze` calls at the start of the test will fall in the initial period, and extended the proposal expiry time so that the entire tests runs before the proposal is expired.

lierdakil:
> That was to make the real network tests to pass in CI

Okay, makes sense. However, I think that should've been a separate commit, with the description more or less explaining what this comment explained (maybe in less detail). Otherwise, the change looks very much out of place amongst purely mechanical API updates. 

> an interface to the localchain to pause

While possible with local-chain, impossible with testnet, and overall doesn't sound like a great solution (as local-chain is shared amongst all CI jobs).

One thing that _might_ help a bit with timing is running nettests with tasty `-j1` option: nettests can't run in parallel anyway, and trying to run those in parallel can lead to arbitrary pauses in random places. On the testnet still might end up breaking though, but time between blocks is significantly larger on the testnet compared to local-chain, so at least level-sensitive tests should work there if they work for local-chain.

haskell/test/SMT/Common/Run.hs
```diff
@@ -243,52 +233,52 @@ handleCallViaLigo (dao, gov, viewC) mc = do
...
+    XtzAllowed (ConcreteEp (Accept_ownership p)) -> transfer dao $ calling (ep @"Accept_ownership") p
+    XtzAllowed (ConcreteEp (Default _)) -> transfer dao
+
+    XtzForbidden (Vote p) -> transfer dao $ calling (ep @"Vote") p
+    XtzForbidden (Flush p) -> transfer dao $ calling (ep @"Flush") p
+    XtzForbidden (Freeze p) -> transfer dao $ calling (ep @"Freeze") p
+    XtzForbidden (Unfreeze p) -> transfer dao $ calling (ep @"Unfreeze") p
+    XtzForbidden (Update_delegate p) -> transfer dao $ calling (ep @"Update_delegate") p
+    XtzForbidden (Drop_proposal p) -> transfer dao $ calling (ep @"Drop_proposal") p
+    XtzForbidden (Unstake_vote p) -> transfer dao $ calling (ep @"Unstake_vote") p
```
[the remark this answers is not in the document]
sras: Gave this a try, but couldn't apply because of the [last branch](https://github.com/tezos-commons/baseDAO/blob/master/haskell/test/SMT/Common/Run.hs#L281) of the case statement.

haskell/test/Test/Ligo/BaseDAO/Common.hs
sras: Seems to throw this error...

```
• Couldn't match type ‘first-class-families-0.8.0.1:Fcf.Core.Eval
||                              (first-class-families-0.8.0.1:Fcf.Data.Common.FromMaybe
||                                 cp
||                                 (first-class-families-0.8.0.1:Fcf.Core.Eval
||                                    (EpdLookupEntrypoint
||                                       (Data.Type.Bool.If
||                                          (Lorentz.Entrypoints.Helpers.CanHaveEntrypoints cp)
||                                          (ParameterEntrypointsDerivation cp)
||                                          EpdNone)
||                                       cp
||                                       Lorentz.Entrypoints.Core.DefaultEpName)))’
||                      with ‘()’
||         arising from a use of ‘transfer’
```

lierdakil: Ah, indeed, that would be the case here, it needs `HasDefEntrypointArg cp ep ()` constraint. The error message is very arcane though, I need to remember to file an issue on morley.

This happens because `ContractHandle` is checked by default. The alternative is to throw away typing information by using `transfer (chAddress addr) [tz|0.5|]` for instance.

haskell/test/SMT/Common/Run.hs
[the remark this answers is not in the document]
lierdakil: Ah, indeed, I missed that last case. Still possible, but the amount of fiddling with types in `CallCustomEp` is just not worth it at that point.

haskell/test/Test/Ligo/BaseDAO/Common.hs
sras: Ok, I ll leave it as it is for now.

lierdakil: Just use `transfer (chAddress addr) [tz|0.5|]`.

haskell/test/Test/Ligo/BaseDAO/Common.hs
sras: For some reason the compilare figures this to be a failable pattern and refuses to accept for want of `MonadFail` instance for the `ClevelandBaseMonad`.

haskell/test/Test/Ligo/BaseDAO/Common.hs
lierdakil: Actually, `transfer (chAddress addr) amt` as we figured out above.

haskell/test/Test/Ligo/BaseDAO/Common.hs
lierdakil: Ugh. Yeah, GHC isn't smart enough to figure out this pattern isn't actually fallible in this context. We do provide `MonadFail` in scenarios, so adding `MonadFail m` will work, but apparently you'd need to add it in a lot of places elsewhere that use this function, which is quite annoying. You could get away with using a pattern-match in a lambda, although it's somewhat ugly:

```haskell
  refillables (newAddresses $ "owner1" :< "operator1" :< "owner2" :< "operator2" :< "admin" :< Nil)
  >>= \(owner1 :< operator1 :< owner2 :< operator2 :< admin :< Nil) -> do
    owner1Balance <- getBalance owner1
-- ...
```

lierdakil (approved):
Okay, I think I need to stop nitpicking, so approving. However, please clean up the commit history before merge, in particular, my point about the change in testing parameters would be better in a separate commit still stands.

sras:
> impossible with testnet

I might be missing something but AFAIU we run on testnets considerably less frequently than we do on localchain. Localchains being shared btw CI jobs is something that we can control, and might even be possible to queue access to it, because the tests can run faster with this hack since we don't have to target delays/periods for the worstcase network performance (hence can use smaller wait times, so for example in BaseDAO, in the place of using `15` levels for one period, we might be able to use `4`).

PR #344, opened 2023-01-17
[Chore] Bump morley &c

lierdakil:
## Related issue(s)

Part of TM-667

- Tests
        silently reappearing again.

- Documentation
    - Haddock


lierdakil:
Note (to self and reviewers): there's a `TODO [this MR]` left in the code, I did figure out the reason is the very same morley#922, and there's a fix MR open on morley, but I forgot to amend the TODOs. I'll just let CI run for now and fix those together with anything else.

lierdakil:
Hmm. Some network tests failed, and I don't immediately see why that would be. Let's try rerunning for now, maybe it's a timing issue.

lierdakil:
So... there are still failures, but specific failing tests are different :confused: seems like a timing issue indeed. I'll try to relax timing a bit and run network tests in a single thread to avoid unfortunate interleaving.

lierdakil:
Single-threading apparently helped, but I'm not sure if I entirely understand why upon reflection. In theory, the network lock is acquired once the tasty test starts running, so only one scenario should run at any single time.

lierdakil:
This may be just down to spurious thread wake-ups. Since network tests are essentially single-threaded anyway, timings will be more consistent if we don't try to run several tests in parallel anyway. Let's try to rerun the test one more time in case it just lucked out, and also make the testnet test single-threaded while at it (network is less timing-sensitive as block period is several times longer, but it won't hurt).

lierdakil:
Okay, at this point, I'm questioning whether relaxing timings more is something we want to do. Consider that the looser we make those, the longer the testsuite takes, and it already takes arguably long enough.

I did come up with an alternative: https://hackage.haskell.org/package/tasty-rerun. Now, it could be argued it's a giant hack. However, I don't know if there's a better alternative, frankly. I've played around with the idea of making a more smart `waitForStage` instead of `advanceLevelTo (start + period*n)` we use now, but timings will still get out of whack unless `period` is rather large.

Thoughts? /cc @sras since IIUC you're more familiar with the issue.

lierdakil:
Well, the combination of relaxing choice timings, reshuffling some initialization and targeted retries seemingly works well enough. Testnet should be generally less susceptible to these issues than local chain due to larger minimal block time.

I'm calling this "good enough" for now, at least for the purposes of review.

pasqu4le:
> there's a TODO [this MR] left in the code

or 3 :slightly_smiling_face: 
> I forgot to amend the TODOs. I'll just let CI run for now and fix those together with anything else.

reminder about that.


> oops, it already is, the command is reused.

could you clarify this [here](https://github.com/tezos-commons/baseDAO/pull/344#issuecomment-1398332290), please? I think I'm missing the point.
Also a reminder that 60cc913038160ffd5a7adafa615b09cd4da11985 is tagged as `[tmp]`, but AFAICT the latest CI runs all keep it, including the one that worked.

> Well, the combination of relaxing choice timings, reshuffling some initialization and targeted retries seemingly works well enough.

It certainly looks like it :clap: 
It's unfortunate that they are needed at all, but IDK what I can suggest about these changes.

The `tasty-rerun` stuff is quite interesting and after some thought it seems a good pick.
However, I'm wondering about the number of re-runs.
IMO  `5` might be too many and potentially hide an issue that appears up to 80% of the times and  should instead be dealt with in a better way.

lierdakil:
> IMO 5 [re-runs] might be too many

Well, technically 4, i.e. one run and 4 re-runs. I guess I could reduce this number to 3 total runs, which seems reasonable.

> should instead be dealt with in a better way.

Eh, as long as we don't have a lot of failures (like, hundreds), and failures are non-deterministic, as seems to be the case here, it could be argued retries are better than relaxing timings across the board. And I don't know of other, "more better", reliable options of dealing with this. I see your point, but myself I'm a bit on the fence about this.

The main downside here is that if some MR breaks a lot of tests, this will try to rerun all of them. However, local-chain tests depend on emulator tests already, so we're pretty confident that if local-chain failed, it's most likely due to the timings.

> reminder about that.

Right, thanks, I got distracted by CI failures.

> could you clarify this here, please? I think I'm missing the point.

The same test script is reused for both local-chain and scheduled testnet tests in `pipeline.yml` via YAML anchors. The strikedthrogh text I wrote under the incorrect assumption it's not.

lierdakil:
I did a force push to fix the description of the `[tmp]` commit, all changes to the code base are in the fixups.

pasqu4le (approved):

lierdakil:
Been a while since I touched this repo the last time, apparently we don't have a changelog here.

Okay, then this should be ready to merge. As a heads-up, I'm enabling auto-merge. If you disagree, you have ~3 hours to disable it :sweat_smile:

PR #345, opened 2023-01-20
[#343] Add ability to call FA1.2 transfers in proposals

### LINKED ISSUE #343: Add support for FA1.2 in transfer proposals

Currently transfer proposals support (in all of our variants) both XTZ transfers as well as FA2 token transfers.

We've been asked to add support for FA1.2 token transfers as well, which have a different type signature from FA2.

# Acceptance criteria

Transfer proposals can perform FA1.2 token transfers as well as the current options.

sras:
## Description

Problem: We need to support FA1.2 transfers in the transfer proposals.

Solution: Add a new branch to the transfer type to represent FA1.2 transfers. Implement handlers for this branch. Update tests to test FA1.2 transfer proposals.

## Related issue(s)

Resolves #343

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -34,11 +34,18 @@ Proposals can be of different type and include (`proposal_metadata`) different d
 2. Proposal to update the set of successful-proposal receiver contract addresses.
 This proposal takes a parameter that has two constructors. The proposal can add to,
 or remove from the set of addresses.
-3. Transfer proposals, that can transfer XTZ or tokens as well as update the
+3. Transfer proposals, that can transfer XTZ or tokens (FA2 as well as FA1.2) as well as update the
 registry. This includes:
    - a list of `update` items parameterized by `k` and `v` types. Each `update` item contains a key of the type `k` and a new value of the type `option v`.
-   - a list of items where each item contains:
-   `or %transfers (pair (mutez %amount) (address %recipient)) (pair (address %fa2) (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))))` specifies what transfer to make. The left part is used for XTZ transfers, the right part is used for FA2 transfers.
+   - a list of transfer instructions. Each item in this list can be either of an Xtz transfer, an FA2 token transfer or an FA1.2 token transfer. These alternatives are represented by the type,
+      ` (or (pair %xtz_transfer_type (mutez %amount) (address %recipient))
+                                   (or (pair %token_transfer_type
+                                          (address %contract_address)
+                                          (list %transfer_list
+                                             (pair (address %from_) (list %txs (pair (address %to_) (nat %token_id) (nat %amount))))))
+                                       (pair %legacy_token_transfer_type
+                                          (address %contract_address)
+                                          (pair %transfer (address %from) (pair %target (address %to) (nat %value))))))`
    - `nat %agoraPostID` is used to refer to an Agora post explaining the proposed transfer and/or changes and motivation for them.
 4. Proposal to update the guardian address in the BaseDAO contract:
    - This proposal takes an address parameter and use it to update the guardian address in the storage.
--- a/haskell/test/Test/Ligo/BaseDAO/Common.hs
+++ b/haskell/test/Test/Ligo/BaseDAO/Common.hs
@@ -10,6 +10,7 @@ module Test.Ligo.BaseDAO.Common
   , OriginateFn
 
   , dummyFA2Contract
+  , dummyFA12Contract
   , dummyGuardianContract
   , makeProposalKey
   , addDataToSign
@@ -37,6 +38,7 @@ import Universum hiding (drop, swap)
 
 import Data.ByteString qualified as BS
 import Lorentz hiding (assert, now, (>>))
+import Lorentz.Contracts.Spec.AbstractLedgerInterface qualified as FA12
 import Lorentz.Contracts.Spec.FA2Interface qualified as FA2
 import Morley.Michelson.Typed.Scope (HasNoOp)
 import Morley.Util.SizedList (SizedList'(..), pattern (::<))
@@ -83,6 +85,18 @@ dummyFA2Contract = defaultContract $
       , #cTransfer /-> cons
       , #cUpdate_operators /-> Lorentz.drop
       )) # nil # pair
+
+-- | A dummy contract with FA1.2 parameter that remembers the
+-- transfer calls.
+dummyFA12Contract :: Contract FA12.Parameter [FA12.TransferParams] ()
+dummyFA12Contract = defaultContract $
+  unpair #
+    (entryCase @FA12.Parameter (Proxy @PlainEntrypointsKind)
+      ( #cTransfer /-> cons
+      , #cGetTotalSupply /-> Lorentz.drop
+      , #cGetBalance /-> Lorentz.drop
+      )) # nil # pair
+
 
 -- | A dummy contract that act as guardian contract for BaseDAO
 dummyGuardianContract :: Contract (Address, ProposalKey) () ()
--- a/haskell/test/Test/Ligo/Common.hs
+++ b/haskell/test/Test/Ligo/Common.hs
@@ -7,6 +7,7 @@ module Test.Ligo.Common
   , TestableVariant(..)
   , withOriginated
   , withOriginatedSetup
+  , withOriginatedFA12
   , toPeriod
   , SizedList'(..)
   , refillables
@@ -15,6 +16,7 @@ module Test.Ligo.Common
 import Prelude
 
 import Lorentz as L hiding (assert, div)
+import Lorentz.Contracts.Spec.AbstractLedgerInterface qualified as FA12
 import Lorentz.Contracts.Spec.FA2Interface qualified as FA2
 import Morley.Michelson.Typed hiding (S, TAddress)
 import Morley.Tezos.Address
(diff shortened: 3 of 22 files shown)


src/variants/lambdaregistry/storage.mligo
```diff
@@ -58,10 +58,29 @@ type token_transfer =
...
+  { from : address
+  ; target : legacy_transfer_target
+  }
+
+type legacy_token_transfer =
+  [@layout:comb]
+  { contract_address : address
+  ; transfer : legacy_transfer
+  }
+
```
lierdakil, line 78: I'm not very familiar with LIGO, why is this duplicated from `common/types.mligo`? Or am I missing something?

src/variants/lambdatreasury/storage.mligo
```diff
@@ -53,10 +53,29 @@ type token_transfer =
...
+  { from : address
+  ; target : legacy_transfer_target
+  }
+
+type legacy_token_transfer =
+  [@layout:comb]
+  { contract_address : address
+  ; transfer : legacy_transfer
+  }
+
```
lierdakil, line 73: Same as above.

haskell/test/Test/Ligo/BaseDAO/Common.hs
```diff
@@ -84,6 +86,18 @@ dummyFA2Contract = defaultContract $
       , #cUpdate_operators /-> Lorentz.drop
       )) # nil # pair
 
+-- | A dummy contract with FA2 parameter that remembers the
```
lierdakil, line 89:
```suggestion
-- | A dummy contract with FA1.2 parameter that remembers the
```

haskell/test/Test/Ligo/Common.hs
```diff
@@ -45,10 +47,29 @@ class VariantConstraints variant =>
...
+  -> (SizedList num ImplicitAddress
+        -> StorageSkeleton (VariantToExtra variant)
+        -> ContractHandle (Parameter' (VariantToParam variant)) (StorageSkeleton (VariantToExtra variant)) ()
+        -> ContractHandle FA2.Parameter [FA2.TransferParams] ()
+        -> ContractHandle FA12.Parameter [FA12.TransferParams] ()
+        -> m a)
+  -> m a
+withOriginatedFA12 storageFn tests = do
+  withOriginated @variant storageFn (\addrs s tc fa2 -> do
+    dodFA12TokenContract <- originate "token_contract" [] dummyFA12Contract
```
lierdakil, line 62: Tests take a long time, and timings are annoying as is, let's not exacerbate the issue.

Please refactor `withOriginated` to accept a parameter which type of token it expects instead of this horrible hack. I'm objecting to this on the grounds that FA1.2 contract is originated during the initial voting period, which will throw off timings, and those are super annoying to deal with as it is.

lierdakil (commented):
I'm not super familiar with the code base here, so it may be confusion on my part, but I'm under the impression the intention is that FA1.2 could be used entirely instead of FA2. Does that imply that the entire test-suite could run either with FA2 or FA1.2 backing? If that is the case, I don't think a couple tests for FA1.2 cover everything. That being said, essentially running the test-suite twice is probably not feasible, but I'd like to hear (well, read) your opinion on the matter.

src/variants/lambdaregistry/storage.mligo
pasqu4le: I may be wrong, but IIRC this is forced by the combination of:
1. the layout of the variants
2. the fact that LIGO is compiled top to bottom and `include` really just put the included code there

I'm going off my memory tho, so I'd like @sras to confirm and add more details

docs/registry.md
```diff
@@ -34,11 +34,12 @@ Proposals can be of different type and include (`proposal_metadata`) different d
 2. Proposal to update the set of successful-proposal receiver contract addresses.
 This proposal takes a parameter that has two constructors. The proposal can add to,
 or remove from the set of addresses.
-3. Transfer proposals, that can transfer XTZ or tokens as well as update the
+3. Transfer proposals, that can transfer XTZ or tokens (FA2 as well as FA1.2) as well as update the
```
pasqu4le, line 37: A few things here seem to be incorrect

For once, one can transfer a list of XTZ, FA2 and FA1.2 tokens in a single proposal AFAIK. It seems that currently this is phrased as one having to pick between:
- an XTZ transfer
- a list of FA2 transfers
- a single FA1.2 transfer

docs/registry.md
```diff
@@ -34,11 +34,12 @@ Proposals can be of different type and include (`proposal_metadata`) different d
...
 registry. This includes:
    - a list of `update` items parameterized by `k` and `v` types. Each `update` item contains a key of the type `k` and a new value of the type `option v`.
-   - a list of items where each item contains:
+   - for FA2 transfers a list of items where each item contains:
    `or %transfers (pair (mutez %amount) (address %recipient)) (pair (address %fa2) (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))))` specifies what transfer to make. The left part is used for XTZ transfers, the right part is used for FA2 transfers.
```
pasqu4le, line 41: (continuing from the previous comment)

this line seems to be outdated too and IDK why FA1.2 wouldn't be considered as an option here as well

docs/registry.md
```diff
@@ -34,11 +34,12 @@ Proposals can be of different type and include (`proposal_metadata`) different d
 2. Proposal to update the set of successful-proposal receiver contract addresses.
 This proposal takes a parameter that has two constructors. The proposal can add to,
 or remove from the set of addresses.
-3. Transfer proposals, that can transfer XTZ or tokens as well as update the
```
pasqu4le, line 37: There is an explanation of the `proposal_metadata` including the transfer proposals in `docs/treasury.md` that needs to be updated too.

pasqu4le (commented):
> I'm under the impression the intention is that FA1.2 could be used entirely instead of FA2. 

Not quite, the intention is only to be able to have the option of calling arbitrary `FA1.2` contract's `transfer` as a result of a successful proposal (as there is the option to do the same for `FA2` contracts already).

This means, among other things, that there is no need/intention to allow for FA1.2 contracts to be used as the governance token (which is where FA2 calls get the most use in baseDAO).

> Does that imply that the entire test-suite could run either with FA2 or FA1.2 backing?

Nope, indeed only transfer proposals should be changed by this PR and IMO those have had an okay coverage in here.

haskell/test/Test/Ligo/Common.hs
sras: > I'm objecting to this on the grounds that FA1.2 contract is originated during the initial voting period, which will throw off timings

Good catch. I have tried a fix in https://github.com/tezos-commons/baseDAO/pull/345/commits/52b8dfcaf174b63bfdc19afc7e9170608886eb4f

docs/registry.md
pasqu4le: Maybe the original text could have been improved, but I meant that the changes introduced seem incorrect.

The original text looked something like this:
```
Transfer proposals [...] This includes:
- a list of `update` items [...]
- a list of items where each item contains: [... long_type_here ...] The left part is used for XTZ transfers, the right part is used for FA2 transfers.
- `nat %agoraPostID`  [...]
```
Reading this I'd think that each transfer proposal contains 3 fields, one of which is a list of transfers, of which each item can be of two different types (which was correct AFAIK).

With the edit here, the text looks something like this:
```
Transfer proposals [...] This includes:
- a list of `update` items [...]
- for FA2 transfers a list of items where each item contains:  [... long_type_here ...] The left part is used for XTZ transfers, the right part is used for FA2 transfers.
  for FA1.2 transfers a single transfer item represented [...]
- `nat %agoraPostID`  [...]
```
here it's not longer clear:
- if FA1.2 transfers will be in the same list
- why the long type that contains two types of transfers doesn't contain all 3 and why it is introduced as "for FA2 transfers" if it also used for XTZ ones?
- can there be only one FA1.2 transfer? Are only FA2 ones in a list?

sras: Ok, makes sense, I have revised that section now.

pasqu4le (commented):
Overall LGTM now, but it looks like there are still failing tests.

It may be wise to rebase this PR on top of #344 to solve that

pasqu4le (approved):
LGTM now

Please squash the two commits together however

haskell/test/Test/Ligo/Common.hs
```diff
@@ -76,7 +90,18 @@ withOriginatedSetup setup storageFn tests = do
...
+withOriginated
+  :: forall variant num num' caps m a num0. (IsoNatPeano num num', SingI num', TestableVariant variant, MonadCleveland caps m, num' ~ 'S num0, MonadFail m)
+  => (SizedList num ImplicitAddress -> StorageSkeleton (VariantToExtra variant) -> StorageSkeleton (VariantToExtra variant))
+  -> (SizedList num ImplicitAddress
+        -> StorageSkeleton (VariantToExtra variant)
+        -> ContractHandle (Parameter' (VariantToParam variant)) (StorageSkeleton (VariantToExtra variant)) ()
+        -> ContractHandle FA2.Parameter [FA2.TransferParams] ()
+        -> m a)
+  -> m a
+withOriginated storageFn tests = withOriginatedSetup @variant (\_ _ -> pass) storageFn (\addr stor cont token _ -> tests addr stor cont token)
```
lierdakil, line 104: Writing out arguments is kinda sad. Why not use `...`?

```suggestion
withOriginated storageFn tests = withOriginatedSetup @variant (\_ _ -> pass) storageFn (const ... tests)
```

lierdakil (approved):
Aside from a minor stylistic nitpick, LGTM.

PR #348, opened 2023-05-08
[Chore] Fix config proposal's frozen_extra_value updates in lambda registry

pasqu4le:
## Description

We were notified of a mistake in the lambda registry's handler for config proposal, which prevents `frozen_extra_value` from being changed.

This makes the required change and adds a regression test, see individual commits messages for details.

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/variants/registry/implementation.mligo
+++ b/src/variants/registry/implementation.mligo
@@ -178,22 +178,22 @@ let decision_callback (input : decision_callback_input)
 
       let new_ce = match cp.frozen_extra_value with
         | Some (frozen_extra_value) ->
-            { extras with frozen_extra_value = frozen_extra_value }
+            { new_ce with frozen_extra_value = frozen_extra_value }
         | None -> new_ce in
 
       let new_ce = match cp.max_proposal_size with
         | Some (max_proposal_size) ->
-            { extras with max_proposal_size = max_proposal_size }
+            { new_ce with max_proposal_size = max_proposal_size }
         | None -> new_ce in
 
       let new_ce = match cp.slash_scale_value with
         | Some (slash_scale_value) ->
-            { extras with slash_scale_value = slash_scale_value }
+            { new_ce with slash_scale_value = slash_scale_value }
         | None -> new_ce in
 
       let new_ce = match cp.slash_division_value with
         | Some (slash_division_value) ->
-            { extras with slash_division_value = slash_division_value }
+            { new_ce with slash_division_value = slash_division_value }
         | None -> new_ce
       in { operations = ops; extras = new_ce; guardian = (None : (address option)) }
   | Transfer_proposal tp ->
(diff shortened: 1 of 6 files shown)


src/variants/registry/implementation.mligo
```diff
@@ -178,22 +178,22 @@ let decision_callback (input : decision_callback_input)
...
       let new_ce = match cp.frozen_extra_value with
         | Some (frozen_extra_value) ->
-            { extras with frozen_extra_value = frozen_extra_value }
+            { new_ce with frozen_extra_value = frozen_extra_value }
         | None -> new_ce in
 
       let new_ce = match cp.max_proposal_size with
         | Some (max_proposal_size) ->
-            { extras with max_proposal_size = max_proposal_size }
+            { new_ce with max_proposal_size = max_proposal_size }
```
sras, line 186: Can you please refresh my memory and tell me why we started this practice of reusing variable names?

pasqu4le: CameLIGO is quite a bit imperative-oriented in its syntax IMO, so reusing variables allows us to update a value multiple times in the context of a single block of code.

I don't think there was any point where we started doing this BTW, it was common from the beginning as the code would be much more verbose to write otherwise.

IOW I think that the reason is that this syntax works well with it and not so much otherwise (compared to other languages) :shrug: 

Here we re-use `new_ce` because we need to check several conditions, each of which can change the resulting `extras` and (with this fix) all these changes should be able to happen at the same time.
Not sure if you are asking about what does this change means here, but just in case: here `new_ce` is re-used because otherwise the updates made in the previous `let new_ce =` will be deleted by using `extras`.

sras: > so reusing variables allows us to update a value multiple times in the context of a single block of code.

Sure, But AFAICS this makes the "unused variable name" warnings pretty unhelpful to detect bugs like what we have here (Which AIU was casued by using the `extra` variable as a source value for the update). So I am asking if there was some reason good enough to make this sacrifice? Maybe we came to it during that time when LIGO didn't have those warnings?

pasqu4le: > Which AIU was casued by using the extra variable as a source value for the update

yep, that's exactly right

> AFAICS this makes the "unused variable name" warnings pretty unhelpful to detect bugs like what we have here

Those warnings unfortunately wouldn't have helped here, most likely.

The code (without the fix) currently looks like a series of these:
```ocaml
      let new_ce = match cp.some_value with
        | Some (some_value) ->
            { extras with some_value = some_value }
        | None -> new_ce in
```
with different variable names, it could have looked like a series of these:
```ocaml
      let new_ce_4 = match cp.some_value with
        | Some (some_value) ->
            { extras with some_value = some_value }
        | None -> new_ce_3 in
```
and there would still be no warning of an unused `new_ce_<x>` as it's in the definition of `new_ce_<x+1>` (or in the case of the last one, it's used in the update of the stored `extras`).

There are certainly cases in which these warnings would help, but this is not one of them and this is instead an actual logical mistake.

> I am asking if there was some reason good enough to make this sacrifice?

With the above in mind, even in the cases where those warnings would help there is still an argument for the ease of coding/maintenance in reusing the variable (e.g. editing this series of almost-identical blocks would be more involved and potentially introduce more bugs as well).

And yeah, AFAIR it's also true that
> we came to it during that time when LIGO didn't have those warnings?

We can reconsider this if you want, but let's do it somewhere else please, as it's not really the in the scope of this PR.

sras (approved):

PR #351, opened 2023-05-23
[TM-679] Fix dependabot warnings

sras:
## Description

There are some dependabot warnings that we would like to fix. So Update package versions in the yarn.lock file until the warnings go away in the output of `yarn audit`.

### DIFF
--- a/typescript/baseDAO/yarn.lock
+++ b/typescript/baseDAO/yarn.lock
@@ -227,10 +227,10 @@
   dependencies:
     "@types/yargs-parser" "*"
 
-ansi-regex@^5.0.0:
-  version "5.0.0"
-  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-5.0.0.tgz#388539f55179bf39339c81af30a654d69f87cb75"
-  integrity sha512-bY6fj56OUQ0hU1KjFNDQuJFezqKdrAyFdIevADiqrWHwSlbmBNMHp5ak2f40Pm8JTFyM2mqxkG6ngkHO11f/lg==
+ansi-regex@>=5.0.1, ansi-regex@^5.0.0:
+  version "6.0.1"
+  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-6.0.1.tgz#3183e38fae9a65d7cb5e53945cd5897d0260a06a"
+  integrity sha512-n5M855fKb2SsfMIiFFoVrABHJC8QtHwVx+mHWP3QcEqBHYienj5dHSgjbxtC0WEZXYt4wcD6zrQElDPhFuZgfA==
 
 ansi-styles@^3.2.1:
   version "3.2.1"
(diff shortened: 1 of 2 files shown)


typescript/baseDAO/yarn.lock
```diff
@@ -227,10 +231,20 @@
...
+    json-schema-traverse "^0.4.1"
+    uri-js "^4.2.2"
+
 ansi-regex@^5.0.0:
-  version "5.0.0"
-  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-5.0.0.tgz#388539f55179bf39339c81af30a654d69f87cb75"
-  integrity sha512-bY6fj56OUQ0hU1KjFNDQuJFezqKdrAyFdIevADiqrWHwSlbmBNMHp5ak2f40Pm8JTFyM2mqxkG6ngkHO11f/lg==
+  version "5.0.1"
+  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-5.0.1.tgz#082cb2c89c9fe8659a311a53bd6a4dc5301db304"
+  integrity sha512-quJQXlTSUGL2LH9SUXo8VwsY4soanhgo6LNSm84E1LBcE8s3O0wpdiRzyR9z/ZZJMlMWv37qOOb9pdJlMUEKFQ==
```
pasqu4le, line 247: AFAICT this still mentions `5.0.0` (which is reported as something to fix) and also `ansi-regex "^5.0.0"` is a dependency of `pretty-format`.

In this case, won't dependabot still report this as something to fix?

pasqu4le (commented):
> until the warnings go away in the output of yarn audit.

Is that how dependabot looks for packages that need to be updated too? :thinking: 

Also, did you update the packages' version with `yarn upgrade`?

sras:
> Is that how dependabot looks for packages that need to be updated too? thinking

Not sure. Do you have a suggestion on an alternate/better way to  test it?

pasqu4le:
> Just assumed it would be the same, because why would it have some different logic

Well, I suppose that's a good question but I was hoping you'd know, given that looking into the validity of these warnings is the first part of TM-679.

A quick look [in their docs](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts#detection-of-insecure-dependencies) reveals that it's based on a completely separate GitHub Advisory Database. If you are interested you can also find more info to answer the "why" there.

> do you have a suggestion on an alternate/better way to test it?

No, `yarn audit` seems the standard and the warnings there seem to be pushed into the database I just mentioned.

So, I suppose that this is a good enough way to check.

> No. Just edited the versions in the yarn.lock file.

This seems to be [advised against by Yarn](https://classic.yarnpkg.com/lang/en/docs/yarn-lock/#toc-managed-by-yarn) and may get different results, so please do it again by using `yarn` instead.

sras:
> This seems to be [advised against by Yarn](https://classic.yarnpkg.com/lang/en/docs/yarn-lock/#toc-managed-by-yarn) and may get different results, so please do it again by using `yarn` instead.

Done. Had to add "resolutions" field to package.json though.

pasqu4le (approved):
I seem to have different results with `yarn` version `1.22.19` :thinking: 

Maybe I'm doing something wrong tho, because I've run:
```bash
yarn upgrade "ansi-regex@>=5.0.1"
yarn upgrade "cookiejar@>=2.1.4"
yarn upgrade "lodash@>=4.17.21"
yarn upgrade "path-parse@>=1.0.7"
yarn upgrade "validator@>=13.7.0"
```
and `yarn audit` still complains, have you used the same commands?
_BTW, you can check the results in [pasqu4le/tm679-tmp](https://github.com/tezos-commons/baseDAO/tree/pasqu4le/tm679-tmp) if you want_

However, other than that, all LGTM and everything seems to work, so I'm approving already.

sras:
> have you used the same commands?

Maybe your commands only updates the direct dependency versions. But AFAIU we want to update the transitive dependencies, so I had explicitly specified the version of these dependencies using the "resolution" field in package.json, and just did a `yarn install`.

